PR #287, opened 2015-01-02
Keep active plugin window open

### LINKED ISSUE #273: Save and Share state to include open plugin
When we add the Launchpad feature which can open specific plugins, utilize that code path to reopen the active plugin when a perma linke is activated.  The framework will need to capture that state.

mmcfarland:
When initializing a saved state, keep the appropriate plugin window
open.

Fixes #273

### DIFF
--- a/src/GeositeFramework/js/Pane.js
+++ b/src/GeositeFramework/js/Pane.js
@@ -64,12 +64,31 @@
         var mapModel = model.get('mapModel'),
             regionData = model.get('regionData');
 
-        model.get('plugins').each(function (pluginModel) {
-            var stateOfPlugin = model.get('stateOfPlugins')[pluginModel.name()];
+        model.get('plugins').each(function(pluginModel) {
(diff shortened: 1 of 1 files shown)


mmcfarland:
@kdeloach Can you post steps to reproduce?  I can't get that to happen.

kdeloach:
1. Goto http://hanoi-ubuntu-vm:84/
2. Click Map Layers > Coastal Resilience > Social and Economic > Families in Poverty
3. Click Save & Share
4. Navigate to permalink

Here's what I see at this point:
image

Trying to open other plugins at this point doesn't work.

mmcfarland:
Seems to be an issue with the Layer Selector plugin when activating close to setting the state.  Looking into it now.  Other plugins don't seem to be affected.

mmcfarland:
Updated.  @kdeloach Discovered #288 while fixing the issue you reported, but doesn't seem to be related to this PR.

kdeloach:
Works great now +1

PR #289, opened 2015-01-08
Show subregions

mmcfarland:
Stubs out communication between subregions and plugins


kdeloach:
I can verify that subregions are visible and clickable. They prevent mouse clicks from propagating to the map and to active plugins. Not sure what they do beyond that.

kdeloach:
+1

mmcfarland:
That's all they're supposed to do for now, thanks.

Issue #291, opened 2015-01-30
Expose raw permalink has to admin users


mmcfarland:
On the Save 'n Share dialog, a user should be able to expose the encoded URL (not shortened) so that they can copy it when creating "Launchpad Issues"

caseypt:
@mmcfarland Can you clarify what you mean by admin users? Do we want to allow anyone to access the URL?

mmcfarland:
Yeah, since we have no user system, yeah - anyone should be able to.  Sorry for the confusing language.  I'm picturing some expandable section on that panel that has the URL or maybe even just the fragment we'll need.

caseypt:
Thanks! Great. That's what I implemented.

caseypt:
Implemented in #296

PR #296, opened 2015-02-09
Add long permalink to share dialog.

caseypt:
* Allow a user to access the unshortened version of the permalink
for the purposes of configuring the launchpad.

Refs #291

### DIFF
--- a/src/GeositeFramework/js/Permalink.js
+++ b/src/GeositeFramework/js/Permalink.js
@@ -51,7 +51,8 @@
 
         events: {
             'change .embed-size.width': function () { this.updateSize('width'); },
-            'change .embed-size.height': function () { this.updateSize('height'); }
+            'change .embed-size.height': function () { this.updateSize('height'); },
+            'click .show-long-permalink': 'toggleLongPermalink'
         },
         
         render: function () {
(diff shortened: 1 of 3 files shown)


caseypt:
I'm not sure if there is more to this than what I did. This addresses the description of #296, but not necessarily the title, which references "admin users".

caseypt:
selection_032

selection_033

src/GeositeFramework/js/Permalink.js
```diff
@@ -108,6 +109,16 @@
 
         updateSize: function (key) {
             this.model.set(key, this.$('.' + key).val());
+        },
+
+        toggleLongPermalink: function() {
+            var $textbox = this.$el.find('#long-permalink-textbox');
+
+            if ($textbox.is(':hidden')) {
```
steventlamb: does `$textbox.toggle()` work here?

steventlamb:
LGTM. I don't have an environment set up to exercise this branch at the moment though.

src/GeositeFramework/js/Permalink.js
caseypt: Right. Forgot about that. Thanks

caseypt:
The launchpad config [only needs](https://github.com/CoastalResilienceNetwork/GeositeFramework/commit/9b4a705bf82727c424a23b43cbdd949c0f8f22f1#diff-3a581756cbee3b507ef56092b7661596R102) the hash from the URL. Any opinions on whether the new input should include the URL prefix + the hash, or just the hash.

kdeloach:
I think it's fine as-is.

kdeloach:
+1

PR #304, opened 2015-02-17
Add support for launchpad button and modal.

caseypt:
* Add launchpad menu config values to schema, class,
and config.
* Create basic modal with launchpad config for each
launchpad button.
* Modify the config to allow for multiple launchpads.

### DIFF
--- a/src/GeositeFramework/App_Data/regionSchema.json
+++ b/src/GeositeFramework/App_Data/regionSchema.json
@@ -31,6 +31,7 @@
                     "text": {"type": "string", "required": true},
                     "url": {"type": "string", "required": true},
                     "popup": {"type": "boolean", "required": false},
+                    "launchpadId": {"type": "string", "required": false},
                     "elementId": {"type": "string", "required": false},
                     "items": {"type": "array", "items": {
                             "type": "object",
@@ -136,27 +137,31 @@
             "type": "array",
             "items": { "type": "string" }
         },
-        "launchpad": {
-            "type": "object",
+        "launchpads": {
+            "type": "array",
             "required": false,
-            "properties": {
-                "showSubRegions": { "type": "boolean", "required": false },
-                "categories": {
-                    "type": "array",
-                    "items": {
-                        "type": "object",
-                        "properties": {
-                            "name": { "type": "string" },
-                            "description": { "type": "string" },
-                            "issues": { 
-                                "type": "array", 
-                                "items": {
-                                    "type": "object",
-                                    "properties": {
-                                        "name": { "type": "string" },
-                                        "description": { "type": "string" },
-                                        "saveCode": { "type": "string" },
-                                        "faIcon": { "type": "string" }
+            "items": {
+                "type": "object",
+                "properties": {
+                    "id": { "type": "string", "required": true },
+                    "showSubRegions": { "type": "boolean", "required": false },
+                    "categories": {
+                        "type": "array",
+                        "items": {
+                            "type": "object",
+                            "properties": {
+                                "name": { "type": "string" },
+                                "description": { "type": "string" },
+                                "issues": {
+                                    "type": "array",
+                                    "items": {
+                                        "type": "object",
+                                        "properties": {
+                                            "name": { "type": "string" },
+                                            "description": { "type": "string" },
+                                            "saveCode": { "type": "string" },
+                                            "faIcon": { "type": "string" }
+                                        }
                                     }
                                 }
                             }
@@ -164,7 +169,7 @@
                     }
                 }
             }
-		}
+        }
     },
     "additionalProperties": false
 }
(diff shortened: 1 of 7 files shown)


caseypt:
Tested with multiple launchpads, but only kept one in the checked-in config.

src/GeositeFramework/Models/Geosite.cs
```diff
@@ -30,6 +30,16 @@ public class Link
             public bool Popup;
 
             /// <summary>
+            /// Should this link open a launchpad
+            /// </summary>
+            public bool Launchpad;
```
mmcfarland: For simplicity, I think we could just get away with the int? `LaunchpadId`.  Since it's nullable, that could be our check.  Reduces the number of attributes the region.json link config would need.

src/GeositeFramework/js/App.js
```diff
@@ -100,6 +104,32 @@
...
         });
     };
+
+    function registerLaunchpadHandlers(launchpadConfigs) {
+        $('a.launchpad-trigger').click(function() {
+            var launchpadId = $(this).data('launchpad-id'),
+                launchpadConfig = getLaunchpadConfig(launchpadConfigs, launchpadId);
+
+            if (!launchpadConfig) {
+                throw new Error("No configuration found for launchpad id " + launchpadId);
```
mmcfarland: There should be an `app.log` method which will log client side errors like this to the server log.

src/GeositeFramework/js/App.js
```diff
@@ -49,6 +49,10 @@
             N.app.syncedMapManager = new N.SyncedMapManager(N.app.models.screen);
 
             registerPopupHandlers();
+
+            if (regionData.launchpads && $('a.launchpad-trigger').length !== 0) {
```
mmcfarland: I like `_.any` for clarity/readability, but just a consideration.

src/GeositeFramework/region.json
```diff
@@ -8,7 +8,8 @@
         "url": "http://www.azavea.com/"
     },
     "headerLinks": [
-        { "text": "Menu", "url":"#", "items": [
+       { "text": "Getting started", "launchpad": true, "launchpadId": 1, "url": "javascript:;" },
```
mmcfarland: Looks like the cs template for creating the launchpad link inherently puts in an href of `javascript:;`, can this be omitted here?

src/GeositeFramework/App_Data/regionSchema.json
```diff
@@ -31,6 +31,8 @@
                     "text": {"type": "string", "required": true},
                     "url": {"type": "string", "required": true},
                     "popup": {"type": "boolean", "required": false},
+                    "launchpad": {"type": "boolean", "required": false},
+                    "launchpadId": {"type": "integer", "required": false},
```
mmcfarland: Consider using a string for id, it might be a little easier for the configur-er to match the launchmap id as a name to the link in the config, rather than an arbitrary number.

mmcfarland:
This isn't obvious (we should migrate to a readme), but the wiki for the region.json settings should get updated to include launchpad settings

https://github.com/CoastalResilienceNetwork/GeositeFramework/wiki/Region.json-Settings

src/GeositeFramework/js/App.js
```diff
@@ -100,6 +104,32 @@
             });
         });
     };
+
+    function registerLaunchpadHandlers(launchpadConfigs) {
```
mmcfarland: Consider creating and moving this code to a new Launcpad module/controller (similar to SubRegion) since the feature will likely add a bit of code.  Would be nice not to have it directly embedded in App.js.

mmcfarland:
Works great, good implementation.

src/GeositeFramework/Models/Geosite.cs
caseypt: Good call.

src/GeositeFramework/js/App.js
caseypt: Thanks.

src/GeositeFramework/js/App.js
caseypt: Switched.

src/GeositeFramework/region.json
caseypt: `url` is a required parameter. I could make it not required, but I'm not totally sure that we want to do that.

src/GeositeFramework/App_Data/regionSchema.json
caseypt: Done.

src/GeositeFramework/js/App.js
caseypt: Good call. Done.

src/GeositeFramework/region.json
mmcfarland: What about just blank in the config then?  Does it still work?

caseypt:
@mmcfarland re: blank `url`. Yeah that works. Updated to be that.

mmcfarland:
+1

kdeloach:
+1

PR #305, opened 2015-02-17
Remove invalid plugins before app loads

### LINKED ISSUE #302: Region polygon toggle plugin errors when no subregions are defined.
If regions don't specify a subregion section in their region.json config, this line throws an undefined error:  https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/master/src/GeositeFramework/plugins/subregion_toggle/main.js#L14

I think the better check would be to prevent that plugin loading if there are no subregions in the config, though this will be a little extra work since it's a generic plugin and not explicitly tied into the framework subregion code.

mmcfarland:
Perform custom validation checks on plugins before loading them.
Currently, the only invalide plugin would be the toggle-subregion plugin
when there are no subregions defined, but more could be added.

Fixes #302

### DIFF
--- a/src/GeositeFramework/js/Pane.js
+++ b/src/GeositeFramework/js/Pane.js
@@ -20,13 +20,20 @@
             });
         model.set('mapModel', mapModel);
 
+        initializeSubregionDisplays(mapModel, model);
+
+    }
+
+    function initializeSubregionDisplays(mapModel, pane)
+    {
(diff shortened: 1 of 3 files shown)


mmcfarland:
@caseypt update per our conversation.

caseypt:
+1. Looks good. Tested locally.

PR #307, opened 2015-02-19
Subregions can define available plugins

mmcfarland:
If there are `availablePlugins` specified, they become whitelisted to be
show in a subregion mode.  If none are specified for a subregion, all
plugins are shown.  Leaving a subregion to full extent will also show
all plugins.


caseypt:
+1

PR #308, opened 2015-02-19
"Pick a Geography" and subregion activation support

caseypt:
Subregions can be activated by either clicking on the
subregion shape on the map, or making a selection from the
launchpad "Pick a Geography" section.

Activating a subregion means 1) zooming to the extent of
the subregion (plus a little buffer), 2) turning off the
subregion graphics layers, and 3) triggering an event on
the map to let it know a subregion has been activated.

### DIFF
--- a/src/GeositeFramework/App_Data/regionSchema.json
+++ b/src/GeositeFramework/App_Data/regionSchema.json
@@ -144,6 +144,8 @@
                 "type": "object",
                 "properties": {
                     "id": { "type": "string", "required": true },
+                    "title": { "type": "string", "required": true },
+                    "description": { "type": "string", "required": true },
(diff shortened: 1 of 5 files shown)


caseypt:
Only bc2edd4 is part of this PR.

mmcfarland:
+1

mmcfarland:
Do you think we should hide the toggle subregion plugin when in an active subregion?  Turning it on shows the polygon and may interfere with clicking

caseypt:
That's a good idea. Maybe we can just disable it, and apply some opacity to indicate that. I can do that in the card I'm working on now, which is the subregion header. I'm also going to handle deactivation.

PR #309, opened 2015-02-23
Add subregion header

caseypt:
Show the subregion header when a subregion is active. Also hide the subregion toggle button when a subregion is active.

Some things to test:

- Enter a subregion. The header should appear, and the subregion toggle button will disappear. Leave the subregion via the leave button in the header.
- Enter and leave a subregion with two maps open. Entering a subregion on one map should not effect the other map.
- Enter a subregion on one map, switch to the other map, open a subregion on that map, then switch back to the first map. The subregion status should be preserved on the first map.
- Drag the map from the initial position, then enter a subregion. When you leave the subregion, you should be back in the position you left the map in before entering the subregion.


mmcfarland:
+1

PR #310, opened 2015-02-23
Styled Launchpad

ctaylo37:
-May still need some tweaks, but a good starting point.
-Broke the sub-regions.
-Needs JQuery scrollbars.
-Height of the export modal is cutting off the button.


caseypt:
Looks good, but can you make the base header/content classses not `launchpad`? We are going to use that just for the launchpad modal.

mmcfarland:
Agree with @caseypt about classname `launchpad`, but otherwise +1

ctaylo37:
Yep.  Will Do.

ctaylo37:
Make sure this looks good and I will merge.

mmcfarland:
Looks good to me, +1

PR #325, opened 2015-02-26
Add config option to show launchpad on load.

### LINKED ISSUE #319: Launchpad: Have default option to show/hide launchpad
region.json should allow a launchpad to show on startup, or not.

caseypt:
* Add a config option to the launchpad config settings
that, if true, will show the launchpad by default when the
page loads.
* Make code changes to support this config change.

Closes #319

### DIFF
--- a/src/GeositeFramework/App_Data/regionSchema.json
+++ b/src/GeositeFramework/App_Data/regionSchema.json
@@ -152,6 +152,7 @@
                     "title": { "type": "string", "required": true },
                     "description": { "type": "string", "required": true },
                     "showSubRegions": { "type": "boolean", "required": false },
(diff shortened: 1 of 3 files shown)


caseypt:
Testing:

- With the current settings, the launchpad will open by default.
- Modify the settings file and change `showByDefault` to false. The launchpad shouldn't show when page loads.
- Remove the config settings entirely. The launchpad shouldn't show when page loads.

caseypt:
Also, if there is more than one launchpad in the config with `showByDefault` set to `true`, the last one in the config will be the one that opens because TinyBox is a singleton.

mmcfarland:
+1

PR #326, opened 2015-02-27
Reorder plugins + just use URL hash for long permalink field.

caseypt:
- The subregion toggle and full extent button should be next to each other.
- The base url should no longer be displayed in the long permalink field of the save and share modal.

### DIFF
--- a/src/GeositeFramework/region.json
+++ b/src/GeositeFramework/region.json
@@ -72,9 +72,9 @@
         "layer_selector",
         "custom_layer_selector",
         "zoom_to",
-        "full_extent",
         "measure",
         "legend_display",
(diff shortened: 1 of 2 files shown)


src/GeositeFramework/region.json
```diff
@@ -72,9 +72,9 @@
         "layer_selector",
         "custom_layer_selector",
         "zoom_to",
-        "full_extent",
         "measure",
         "legend_display",
+        "full_extent",
```
mmcfarland: We'll have to remind TNC that this is all they need to do to accomplish this in their own regions.

mmcfarland:
+1

PR #327, opened 2015-03-02
Make single map launchpad issues act on active map

### LINKED ISSUE #317: Launchpad: Scenarios should work activate in the current map
If a single map is active, the launchpad should open scenarios in it (similar to sub-regions).  If in split screen, use the primary map (pane 0).

caseypt:
* If a saved state/launchpad issue contains only one map/pane,
have it update the active map instead of the map the state
was saved for. For example, if the user has the second map
active and they launch an issue that was saved with the first
map active, the second map will now be updated with the saved
state instead of the first map.

Closes #317

### DIFF
--- a/src/GeositeFramework/js/Launchpad.js
+++ b/src/GeositeFramework/js/Launchpad.js
@@ -78,8 +78,17 @@ require(['use!Geosite'], function (N) {
                 issue = _(category.issues).findWhere({id: issueId}) 
             });
 
-            if (issue) { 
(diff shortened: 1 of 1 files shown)


caseypt:
To test this, activate the second map and select one of the launchpad issues (besides the split screen one). The active map should be updated. For best results, let all the callbacks error out before launching the issue. Then go back to the first map, and select an issue, it should update that map.

mmcfarland:
+1  Nice clean code.

caseypt:
Thanks.

PR #329, opened 2015-03-02
Add "Free Explore" button to launchpad

caseypt:
* Add a button to the launchpad that allows the user to close
the launchpad without activating a subregion or issue.
* If a subregion is active, pressing the button deactivates the
subregion and resets the map to the initial view.
* If no subregion is activate, but the map was moved from the
initial view, pressing the button resets the map to the
initial view.
* Pressing the button only effects the active map or the first
map in split screen view.

### DIFF
--- a/src/GeositeFramework/Views/Home/Index.cshtml
+++ b/src/GeositeFramework/Views/Home/Index.cshtml
@@ -380,6 +380,7 @@
                         </div>
                         <div class="thumbs">
                             <ul>
+                                <li class="free-explore"><i class="fa fa-globe"></i><label>Explore the map</label>
                                 <% _.each(subregions.areas, function(subregion) { %>
                                 <li class="subregion" data-subregion-id="<%= subregion.id %>"><i class="fa fa-map-marker"></i><label><%= subregion.display %></label></li>
                                 <% }); %>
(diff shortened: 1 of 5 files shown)


caseypt:
To test:

- Load the app, click the free explore button. Nothing should happen.
- Load the app, select a subregion. Open the launchpad, click free explore. You should exit the subregion, and the map view should be reset to the initial state.
- Load the app, close the launchpad. Drag the map a bit, then select a subregion. Open the launchpad, click free explore, and the map should be reset to the initial state, not the previous panned position.
- Load the app, select a subregion. Activate the second map, and click on a subregion. Open the launchpad, select free explore. The subregion should be deactivated only in the active (second) map. Toggle back to the first map. The subregion should still be active.
- Load the app, select a subregion. Activate split-screen mode. Activate a subregion in the second pane. Open the launchpad, click free explore. The subregion should be deactivated in only the first map.

caseypt:
Just realized that this doesn't reset the extent when an issue was activated. Going to work on this tomorrow.

caseypt:
@mmcfarland: Updated. Clicking "Free Explore" after selecting an issue now resets the map to the initial extent.

mmcfarland:
+1 Looks good.

Issue #314, opened 2015-02-25
Launchpad: Layer selector plugin needs to load


mmcfarland:
Currently, if a scenario is launched from the LP prior to the layer selector loading its layers, it throws an error.

caseypt:
@mmcfarland Not seeing this behavior. Want to double-check on `master`?

mmcfarland:
@caseypt I'm getting layer selector errors on master when selecting Flower scenario.  Do you not see that?

caseypt:
I get the usual errors from layer selector, as well as this one:

![screen shot 2015-03-09 at 8 08 38 pm](https://cloud.githubusercontent.com/assets/1042475/6567067/2ce41e18-c698-11e4-84cb-6f0c60761b26.png)

However, I get that error when I try to open the layer selector in [production](http://maps.coastalresilience.org/us/) before it's done loading. Also, it doesn't prevent the "Multi-Hazard Mortality Risk Index" layer from being added on the map and eventually being selected in the layer selector once the plugin fully loads.

mmcfarland:
@caseypt Nope, not seeing the described behavior in most recent `master`

caseypt:
Okay cool. There is definitely a bug with the layer selector plug-in, but I'm pretty certain it doesn't have to do with the launchpad. Close this and reopen once we have time to diagnose the other bug?

kdeloach:
@caseypt Not seeing this error anymore on latest master. Close?

caseypt:
Yeah.

PR #330, opened 2015-03-04
Activate subregion specified in saveCode

### LINKED ISSUE #315: Launchpad: Set subregions based on saved state
The launchpad scenarios may not respect the saved state for sub-regions.

mmcfarland:
For scenarios and saved states, pass along the active subregion if one is
specified to the framework and to the plugins.

There are some persistant issues with the layer_selector plugin relating
to #314

Fixes #315

### DIFF
--- a/src/GeositeFramework/js/Pane.js
+++ b/src/GeositeFramework/js/Pane.js
@@ -27,12 +27,16 @@
     function initializeSubregionDisplays(mapModel, pane)
     {
         mapModel.on('subregion-activate', function(activeRegion) {
+            pane.set('activeSubregion', activeRegion);
+
             invokeOnPlugins(pane, 'subregionActivated', [activeRegion, pane]);
             invokeOnPlugins(pane, 'hibernate', [activeRegion]);
             setSidebarPluginVisibility(pane, activeRegion.availablePlugins);
         });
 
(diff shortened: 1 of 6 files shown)


caseypt:
+1. Can't test at the moment, but the code looks good.

mmcfarland:
Doesn't seem to be working with a permalink, even without layer_selector

mmcfarland:
@caseypt I could only duplicate the issue you mentioned (failing on permalink version of save code) when the layer selector is in the project.  You said you could reproduce it on other plugins?

caseypt:
Yeah I believe I removed both the layer selector and custom layer selector plug-in folders, reloaded the map, activated a subregion, then used the share code.

caseypt:
I wasn't getting any errors. The map was zooming in, but the subregion was not becoming activated.

mmcfarland:
Updated.  Attempts to further unify the saveState and scenario code to reduce duplication and discrepancies in behavior.  The new scenario saveCode for Flower/Risk should work similarly if activated from the LP or pasted into the hash URL, with the known issue of: #334.

caseypt:
+1. I was able to launch a subregion from a permalink and the launchpad.

PR #331, opened 2015-03-04
Add region dropdown

### LINKED ISSUE #322: Add subregion dropdown to subregion header
From the wireframes:

selection_059

mmcfarland:
Select other regions directly from region header dropdown

Fixes #322

### DIFF
--- a/src/GeositeFramework/Views/Home/Index.cshtml
+++ b/src/GeositeFramework/Views/Home/Index.cshtml
@@ -415,7 +415,15 @@
 
     <script type="text/template" id="template-subregion">
         <div class="subregion-header">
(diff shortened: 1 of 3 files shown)


caseypt:
In split screen mode, selecting a subregion from the drop-down menu on map 2 exits the subregion on that map, without activating the selected subregion.

caseypt:
Besides the aforementioned behavior, this looks good. +1.

mmcfarland:
@caseypt Updated to fix issue.  There was some overlap with the LP region selection that was only activating when the map pane was the main map pane.  Preserved that behavior for LP but let activation happen for any pane from the header.

PR #332, opened 2015-03-04
Activate region while in region

### LINKED ISSUE #313: Handle go to subregion from launchpad when a subregion is already active
If the user is in subregion A, and selects subregion A from the launchpad, the launchpad should close and nothing else should happen. If the user is in subregion A, and selects subregion B from the launchpad, the launchpad should close, then we should exit subregion A and go to subregion B.

mmcfarland:
Only mind ccbe47a

Fixes #313


caseypt:
+1. Tested.

PR #337, opened 2015-03-09
Fix issue with two plugin windows active at once.

### LINKED ISSUE #288: 2 Plugin windows active at once
You can make a situation where 2 plugin windows are active at once.

1. On a fresh page load, open the Layer Selector plugin
2. Drag the window slightly to the side
3. Click any other plugin
4. Both plugin windows are open

This only happens when you open Layer Selector first (or maybe, the first plugin in the collection?).  Once you close and reopen the plugin, the behavior returns to normal.  If you open another plugin first (even a Layer Selector based plugin), the behavior is correct.

I think this is an issue with the ``select`` [link](https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/master/src/GeositeFramework/js/lib/backbone.picky.js#L125) method of backbone.picky.  It appears the event is not available right away, and it throws and suppresses an exception which prevents the following line of code from getting executed:

```js 
this.collection.select(this);
``` 
[link](https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/master/src/GeositeFramework/js/lib/backbone.picky.js#L128)

 leaving the collection thinking that no items in the collection are selected.

This issue seems like it's been around a while.

caseypt:
This fixes the following scenario:

When the map first loaded, if the user activated the layer selector,
and then activated another plugin, and the layer selector was still
retrieving layers, the layer selector window would not close when
the other plugin was activated. Likewise, if a user activated
a different plugin when the map first loaded, then activated
the layer selector plugin, the first plugin window would not be
closed and the layer selector plugin window would open.

Once all of the layer selectors layers were retrieved, the plugins
would behave normally, i.e. activating one plugin would close any
other active plugin window.

It's not totally clear why this was happening, but it maybe has
something to do with the backbone.picky collection select method not
recognizing that the layer selector was selected if it was selected
when it was still loading.

Deferring the running of the event handler functions until after the
current call stack cleared fixes the issue.

Closes #288

### DIFF
--- a/src/GeositeFramework/js/Plugin.js
+++ b/src/GeositeFramework/js/Plugin.js
@@ -211,8 +211,15 @@ require(['use!Geosite',
             var model = view.model;
             view.render();
             model.on('selected deselected', function () {
-                view.render();
-                model.onSelectedChanged();
+                // Pushing this to the end of the call stack
+                // fixes an issue that caused the layer selector
+                // to not be handled properly as part of a picky
+                // collection until all it's layers were loaded.
+                // See #288
+                _.defer(function() {
+                    view.render();
+                    model.onSelectedChanged();
+                });
             });
             model.on('change:active', function () {
                 view.render();


caseypt:
In addition to trying the scenario described in the first paragraph above, test some scenarios to make sure that the layer selector still loads properly and that any active layers are added to the map.

mmcfarland:
+1

PR #339, opened 2015-03-10
Add border to map when subregion is active.

### LINKED ISSUE #335: Add border to subregion

caseypt:
* Per the wireframes, there should be a border around the map
when a subregion is active that matches the color of the
subregion header.

Closes #335


caseypt:
Much like the header, the border should only be added/removed to the map that the subregion was activated/deactivated on.

ctaylo37:
+1

PR #340, opened 2015-03-10
Don't show the launchpad on init for permalinks.

### LINKED ISSUE #333: Launchpad shouldn't autostart if a permalink initiated the page
If a permalink was used (save n share), the launchpad shouldn't start, even if the region setting is set to show by default.

caseypt:
If the user came to the page through a URL that has
a hash, i.e. application state, don't show the launchpad
on initalization even if the showByDefault launchpad
config setting is set to true.

The presumption is that if the user was given a permalink,
the person who shared the link wants the user to go right
to the map, and not clear out that saved state by clicking
on an issue on the launchpad.

Closes #333

### DIFF
--- a/src/GeositeFramework/js/Launchpad.js
+++ b/src/GeositeFramework/js/Launchpad.js
@@ -30,7 +30,7 @@ require(['use!Geosite'], function (N) {
             this.template = N.app.templates['template-launchpad'];
             this.initialExtent = parseExtent(N.app.data.region.initialExtent);
 
-            if (this.model.get('showByDefault')) {
+            if (this.model.get('showByDefault') && N.app.loadedWithState === false) {
                 this.render();
             }
         },
--- a/src/GeositeFramework/js/App.js
+++ b/src/GeositeFramework/js/App.js
@@ -19,6 +19,11 @@
             N.app.data.region = regionData;
             N.plugins = pluginClasses;
 
+            N.app.loadedWithState = false;
+            if (location.hash !== '') {
+                N.app.loadedWithState = true;
+            }
+
             // Set up the google url shortener service
             gapi.client.load('urlshortener', 'v1');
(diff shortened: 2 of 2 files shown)


caseypt:
To test, load the app with no saved state. The launchpad should appear. Now grab the permalink and open the link in a new tab. The launchpad should not appear in this instance.

caseypt:
Anything else to check for here?

mmcfarland:
Nope, nothing else to check.  Looks good.

src/GeositeFramework/js/Launchpad.js
```diff
@@ -30,7 +30,7 @@ require(['use!Geosite'], function (N) {
             this.template = N.app.templates['template-launchpad'];
             this.initialExtent = parseExtent(N.app.data.region.initialExtent);
 
-            if (this.model.get('showByDefault')) {
+            if (this.model.get('showByDefault') && location.hash === '') {
```
mmcfarland: Checking location.hash is a _little_ fragile because of [this fn](https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/master/src/GeositeFramework/js/App.js#L70) gets called before the LP is initialized, the has would always be set.  Right now the hashModel call is [just the next line](https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/master/src/GeositeFramework/js/App.js#L34) after the LP init.  Might be a little more future proof to check for hash contents first thing in App and save that bool as an attr on the LP model.

mmcfarland:
+1

src/GeositeFramework/js/Launchpad.js
caseypt: Will do.

caseypt:
@mmcfarland Updated.

PR #344, opened 2015-03-10
Target correct map on subregion init from savecode

### LINKED ISSUE #338: Subregion state is not preserved/activated correctly from a split-screen scenario.
- Create a split-screen view, activate a subregion on the second map. When you reload the state with the permalink, the subregion is activated on the first map.

- Create a split-screen view, activate a subregion on both maps. When you reload the state with the permalink, the subregion that was activate on the second map is activated on the first map, and no subregion is activated on the second map.

caseypt:
* When using a permalink or loading a scenario with activate
subregions on both maps or the second map, reactivate the
subregion on the correct map(s).
* Swap setTimeout for defer which is slightly cleaner.

Closes #338

### DIFF
--- a/src/GeositeFramework/js/Pane.js
+++ b/src/GeositeFramework/js/Pane.js
@@ -133,12 +133,15 @@
 
         // Drop the plugin state setting until after the stack clears to prevent errors when the
         // map is not in a ready state.
-        setTimeout(function() {
-            var activeSubregion = pane.get('activeSubregion');
+        _.defer(function() {
+            var activeSubregion = pane.get('activeSubregion'),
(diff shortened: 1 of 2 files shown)


caseypt:
- With one active map, activating a subregion from the launchpad should target the active map (map 1 or 2)
- In split screen mode, activating a subregion from the launchpad should target the first map.
- When reactivating a state in split screen mode with a subregion active on each map, both subregions should be activated.
- When reactivating a state in split screen mode with a subregion active on the second map, the subregion should be activated on the second map.

mmcfarland:
+1

PR #345, opened 2015-03-10
Remove unneeded call to stopPropagation

### LINKED ISSUE #342: In IE, subregion activation doesn't zoom in on map.
Running into a syntax error on [this line](https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/master/src/GeositeFramework/js/SubRegion.js#L128), which prevents the rest of the code from running. Should be an easy fix.

caseypt:
* This was causing errors in IE and is no longer needed
since the function it is in is not directly called from
a click event.

Closes #342


mmcfarland:
+1

PR #348, opened 2015-03-10
Fix hiding of launchpad on init for IE.

caseypt:
* IE treats an empty hash differently than Chrome
and Firefox.


caseypt:
Merging in for demo.

PR #350, opened 2015-03-11
Layer Selector uses local regional data

### LINKED ISSUE #320: Extend map_layers to load "local" data set on region change
Add support to allow the map_layer plugins to load an alternative set of data when a subregion is activated

mmcfarland:
Upgrade to the layer selector to take specific action when
entering/exiting a subrgion.  Changes add new attributes to a top-level
layer source (ags or wms) in the layers config and enumerates the
different regions that layer is allowed in.

Fixes #320


mmcfarland:
To test, notice that the layer selector lists a new folder for "ocean" when in the main mode.
* Go to Galveston region, notice ocean is gone, but "Galveston" local is added
* Go to Flower region and see that neither exist
* Go to Ocean and see that "ocean" is back.
* Back to main region, "ocean" exists again

Ocean has been set to Main + Ocean.

caseypt:
Depending on when you open up the layer selector control, the layer list may be cleared out. 

If you very quickly open the app, launch the subregion, then open the layer selector, the layer selector shows the loading gif. When the bad responses come back from the initial layer selector load, the layers appear. When the bad layer responses come back from the layer selector initialization kicked off by the activation of the subregion, the layers are cleared out.

caseypt:
Alternatively, if you wait until the initial bad layer responses come back, then activated the galveston subregion, then wait until the next set of bad layer responses to come back before opening the layer selector, everything is fine.

mmcfarland:
These seem related to the previous issue discussed with the delay in bad layers completing and the init routine.  Maybe I should rebase recent additions to master on it.

mmcfarland:
@caseypt Updated - issue remains?

caseypt:
Yeah, the issue is still there.

caseypt:
Refs #350

src/GeositeFramework/plugins/layer_selector/layers.json
```diff
@@ -1,6 +1,51 @@
 [
     {
         "agsSource": {
+            "availableInRegions": ["galveston"],
+            "url": "http://services.coastalresilience.org:6080/arcgis/rest/services",
+            "folderTitle": "Galvestone Local",
+            "folders": [
+                {
+                    "url": "http://services.coastalresilience.org:6080/arcgis/rest/services/Global",
+                    "name": "",  // Don't show a folder for these services
```
kdeloach: Sadly, I don't think this is valid JSON.

mmcfarland: Yeah, I was surprised to see the comments in there.  These are just copied from the main layer json, which appear to work as far as the parsing and the validator is concerned.

RickMohr: As I recall we parse the JSON in a way that explicitly allows comments.

kdeloach:
If you wait for the layer selector to load completely, and then switch to Flower Garden Hazard, you get a duplicate layers problem. I'm guessing the old layer is not being cleared properly.

image

mmcfarland:
Fixes #320

mmcfarland:
@kdeloach I'm not seeing this behavior, can you show me how to recreate when you have a minute?

kdeloach:
Sorry for lack of a better description.

PR #358, opened 2015-03-17
Fix subregion events

### LINKED ISSUE #353: Clicking on a subregion to activate doesn't work after activating two subregions from the launchpad

caseypt:
See the commit messages for details.

A couple things to test:

- Activate a subregion from the launchpad. Open up the launchpad again, and select a different subregion.
Leave the subregion. Click on any of the subregions on the map. The subregion should activate
- Activate a subregion through any method. Switch subregions using the drop down menu. The map border and map elements should not be moved or removed.
- Activate a subregion. Open the launchpad. Select free explore. The subregion should exit. Click the same same subregion you just exited. The subregion should activate.

Try these things with multiple maps open and with map 2 as the active map.

Also, keep an eye on the console. The messages indicating subregion activation/deactivation should never be duplicated. The pattern should be enter > leave > enter > leave, etc.

Closes #353

### DIFF
--- a/src/GeositeFramework/js/SubRegion.js
+++ b/src/GeositeFramework/js/SubRegion.js
@@ -101,7 +101,6 @@
             var oldRegion = _.findWhere(this.subregions.areas, { id: this.currentRegionId });
             // Use the full extent when exiting this subregion
             extentOnExit = this.initialExtent;
-            deactivateSubRegion(this, this.map.extent, oldRegion);
             if (this.currentHeader) {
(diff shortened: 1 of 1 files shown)


kdeloach:
Do active plugins remain active when switching to another subregion by design?

kdeloach:
+1 Tested and working.

caseypt:
@kdeloach It's not by design. There just isn't a good way to reset the state of everything at the moment.

PR #359, opened 2015-03-17
Stop subregion feature click from propagating.

### LINKED ISSUE #352: Click propagates through subregion to map.
I was wrong about this not having an effect https://github.com/CoastalResilienceNetwork/GeositeFramework/pull/345. 

Clicks are now propagating through to the map.

We need to re-implement what was removed, but have it not cause problems in IE.

caseypt:
* Make sure the event does not propagate to the map
and trigger an identify point event.

Closes #352

### DIFF
--- a/src/GeositeFramework/js/SubRegion.js
+++ b/src/GeositeFramework/js/SubRegion.js
@@ -72,6 +72,8 @@
 
             self.subRegionLayer.on('click', function (e) {
                 self.activateSubRegion(e.graphic, Polygon);
+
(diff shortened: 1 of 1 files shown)


caseypt:
To test, when you click a subregion on the map, it should activate, but an "identify point" window should not open. Make sure to test in IE.

kdeloach:
Was this planned for this sprint? Issue #352 appears in backlog in Waffle. Should it be moved to In Review?

kdeloach:
+1

PR #360, opened 2015-03-18
Make modal header color configurable.

### LINKED ISSUE #351: Replace hard-coded color values for popovers
The new popover style uses hard coded color values associated with the main Coastal Resilience app, but these colors are actually configurable within the region.json file. 

These color values:
https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/master/src/GeositeFramework/css/app.css#L770

Should be set similarly to how the header, etc are set:
https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/master/src/GeositeFramework/Views/Home/Index.cshtml#L86

caseypt:
* Use the primary app color variable to style the modal
header.
* Set a more neutral default for the modal header.

Closes #351

### DIFF
--- a/src/GeositeFramework/Views/Home/Index.cshtml
+++ b/src/GeositeFramework/Views/Home/Index.cshtml
@@ -110,6 +110,9 @@
             border-top: 15px solid @Model.SecondaryColor;
         }
 
+        body .popover-header {
+            background-color: @Model.PrimaryColor;
+        }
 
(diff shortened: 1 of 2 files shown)


caseypt:
Try changing the primary app color in the config. The header color on the modals should change.

kdeloach:
+1 Works.

Issue #356, opened 2015-03-17
Switching between launchpad scenarios should clear previous launchpad


caseypt:

mmcfarland:
@caseypt - did we resolve this in our pairing session last week?

caseypt:
Yeah. PR is still open though.

caseypt:
Implemented in #363

PR #363, opened 2015-03-20
Always refresh app state when launching scenarios.

### LINKED ISSUE #349: Same scenario fails to reactivate after user interaction
- Refresh the page
- Click the "Flower Garden Hazard" scenario
- Move the map or leave the subregion
- Open the launchpad and click the "Flower Garden Hazard" scenario again
- Nothing happens.

If you close the layer selector plugin instead of leaving the subregion or moving the map, the plugin will reopen when the scenario is clicked again, but the active layers are not turned on.

caseypt:
* Turn off all plugins, reset the map extent and basemap,
and exit any active subregions when launching a scenario.
This prevents scenario states from bleeding into one
another. It also allows a user to reactivate the same
scenario no matter what actions were taken on the map
after the scenario was initially launched.

Closes #349, #356

### DIFF
--- a/src/GeositeFramework/js/App.js
+++ b/src/GeositeFramework/js/App.js
@@ -48,7 +48,8 @@
             });
 
             N.app.dispatcher.on('launchpad:activate-scenario', function(state) {
-                self.hashModels.triggerStateChange(state);
+                var forceHashChange = true;
+                self.hashModels.triggerStateChange(state, forceHashChange);
             });
 
             N.app.views.screen = new N.views.Screen({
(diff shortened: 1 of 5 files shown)


caseypt:
To test:

- Launch a scenario, make some changes to the map, then launch the same scenario again. The map should be reset to the initial scenario state.
- Turn on all the plugins, then launch a scenario that only has one of the plugins active. All the plugins will get turned off, then only the one that is active in the scenario should be relaunched.
- Activate a subregion, then launch a scenario without a subregion active. The subregion should be deactivated.
- Activate a subregion, then launch the Galveston scenario. The active subregion should be switched to Galveston.

kdeloach:
+1

PR #373, opened 2015-03-23
Allow initial full selection of saveCode

### LINKED ISSUE #368: Save and Share code should be selected by default
We do the same for the Save and Share Url, makes it easier to copy the whole string.   May not be necessary if we end up implementing #312

mmcfarland:
To allow for easier copying, have the whole thing selected.  Cannot have
multiple inputs with focus, so this steals the focus from the URL box.

Fixes #368


caseypt:
+1 Nice.

PR #374, opened 2015-03-23
Allow map plugins to be selectable in a subregion.

### LINKED ISSUE #361: Launchpad: Map plugins not active on scenario launch
Click any scenario and try to use the map plugins (measure, address, etc).  Plugins are white listed so they probably are deactivated.

caseypt:
* Layer the map plugins above the div element that is used to draw
the subregion border. Previously, they were under the div and
could not be selected.

Closes #361

### DIFF
--- a/src/GeositeFramework/css/app.css
+++ b/src/GeositeFramework/css/app.css
@@ -124,6 +124,7 @@ input.code-like, textarea.code-like {
 .control-container {
     position: absolute;
     width: 100%;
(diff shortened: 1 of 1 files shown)


caseypt:
You should now be able to use the measure tool, full extent, etc. plugins when a subregion is active.

mmcfarland:
+1
I like when they're nice and easy.

PR #379, opened 2015-03-23
Fix saving of map state

### LINKED ISSUE #375: Fix map state for save n' share
There's a regression where the map state is not encoded properly in the state produced by the save n' share.

caseypt:
* It's important to directly call updateHash and trigger a change
event in update and not rely on handleHashChange since it may not fire
a change event if the hash did not change and force was not set to
true.

Fixes #375


caseypt:
The save and share button should work again, and when you launch a saved state, the map extent and basemap should be set properly.

mmcfarland:
+1  
Behaving correctly.

PR #381, opened 2015-03-23
Don't silence change event when updating extent

caseypt:
* This caused the map extent to become stale. It was originally
thought that if the event wasn't silent, the map would get
stuck in an endless loop of saving the extent, firing a change
event, updating the map, then saving the extent, etc. However,
it appears that the ESRI map event 'extent-change' does not
get fired if the extent doesn't actually change.

### DIFF
--- a/src/GeositeFramework/js/Map.js
+++ b/src/GeositeFramework/js/Map.js
@@ -109,7 +109,7 @@ require(['use!Geosite',
         selectBasemap(view);
 
         view.esriMap.on('extent-change', function() {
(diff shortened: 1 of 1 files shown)


caseypt:
Just move the map to a new extent and generate a permalink. When using the permalink, the map should be set to the correct extent.

mmcfarland:
+1

For what it's worth re: your comment, I think it's that Backbone doesn't fire the "change" event if the model attribute doesn't change, rather than the "extent-change" event from jsapi.

caseypt:
Oh I see, yeah, you may be right.

PR #382, opened 2015-03-24
Add config for subregion border and no fill option

### LINKED ISSUE #366: Outline of subregion box (no fill); specify outline color
Not sure if this should be an _option_ for a no-fill outline, or if that's the _only_ way they appear.  @zferdana, can you comment as to which?

caseypt:
* Allow the subregion outline color to be configurable.
* Add an option to specify that a subregion should have
no fill.

Closes #366


caseypt:
The config now shows:

- A subregion that uses the global subregion fill color and it's own outline color.
- A subregion that uses the global subregion outline color and it's own fill color.
- A subregion that uses it's own outline color and specifies that it should have no fill.

caseypt:
selection_066

src/GeositeFramework/js/SubRegion.js
```diff
@@ -146,7 +146,15 @@
         _.each(subregions.areas, function(subregion) {
             var geom = new esri.geometry.Polygon(subregion.shape);
             var symbol = new esri.symbol.SimpleFillSymbol();
+            var outline = new esri.symbol.SimpleLineSymbol(
+                    esri.symbol.SimpleLineSymbol.STYLE_SOLID,
+                    new dojo.Color(subregion.outlineColor || subregions.outlineColor), 2);
```
mmcfarland: If both `subregion` and `subregions` don't have `outlineColor` defined, does this get an error or some default inside of dojo?

caseypt: It defaults to a very faint white:

selection_067

caseypt: ...because the opacity was still set. If the opacity isn't set, here's what it looks like:

selection_068

mmcfarland:
+1

PR #384, opened 2015-03-24
Debounce saving of map state on extent change.

caseypt:
* If the extent was changed too quickly, the ESRI JS API
would throw an error about the tiles not being able to
animate and the tiles wouldn't draw. Debouncing the
saving of the map extent fixes this.


caseypt:
To test, try launching scenarios quickly after one another. The tiles shouldn't disappear/turn gray.

caseypt:
@mmcfarland Updated. This should fix the bouncy split screen and synced map zoom issues.

mmcfarland:
+1 looks good.

PR #385, opened 2015-03-24
Fixes #380.

### LINKED ISSUE #380: Fix scenario label to bottom
Currently, the scenario name label on a launchpad button floats underneath the icon.  But if the icons are set at variable sizes, they labels end up being non-uniform.  They should just stick to a certain spot for all buttons (the bottom?)

ctaylo37:
Added absolute positioning to icons and labels in the launchpad to allow for icon size configuration.
Also, styled the subregion-header dropdoen.


caseypt:
+1

mmcfarland:
+1

PR #386, opened 2015-03-25
Check for pane state before passing active region.

caseypt:
* Fixes an issue that prevented scenarios from completely
activating on Map 2 if it was the active map.


mmcfarland:
+1 I was able to correctly activate the scenario "Sea Rise" when Map 2 was active.  Is that the right test?

caseypt:
Yup! Thanks.

PR #394, opened 2015-03-27
Closes #389: Change color and width of sidebar scroll bar.

### LINKED ISSUE #389: Teal scroll bar on plugin vertical bar needs to be fatter and made bright white

caseypt:
* Scroll bar is now white and a bit wider. However, to
not have the wide scroll bar overlap with the plugin text,
the plugin containers were given some padding so that the
text would wrap.

Closes #389

### DIFF
--- a/src/GeositeFramework/css/app.css
+++ b/src/GeositeFramework/css/app.css
@@ -222,6 +222,10 @@ input.code-like, textarea.code-like {
             border: none;
             color: transparent;
         }
+        .content .sidebar-plugin .plugin-launcher div {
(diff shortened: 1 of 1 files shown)


caseypt:
selection_078

selection_077

caseypt:
The current scroll bar is slightly faded out when it's out of focus, then becomes brighter on focus. I maintained this behavior (as soon in the above screenshots).

Should the scroll bar always be bright white?

mmcfarland:
+ 1 I think maintaining the behavior is good.

PR #396, opened 2015-03-27
Closes #393: Make the subregion row launchpad text configurable.

### LINKED ISSUE #393: Allow configurable "Pick a geography text"
Other sections are editable, but the pick a geo text Ipsum is hard coded in the app.

screenshot from 2015-03-27 09 03 20

caseypt:
* Add an option to the region config file so users can specify
what text is displayed in the launchpad for the subregion row.
* Also fix an issue that prevented the app from loading if the
the key showSubRegions was missing from the region config file.

Closes #393


kdeloach:
+1

PR #397, opened 2015-03-27
Closes #387: Call plugin resize method on continue.

### LINKED ISSUE #387: Infographic continue button should fire resize event for plugin
From George:

In several of my apps I create form elements that are placed out and sized in Javascript rather than just in CSS (they respond to the resize event to update). One problem is that some of these elements, when they are drawn will not draw correctly when their parent element is hidden (display: none). This happens when an infographic is given for a plugin (the main plugin panel is hidden until they click continue.I've coded up a hack that works around this (query for the continue button on the infographic and attach an even to it). However, it would be nice to have this included in the framework code though so I would not have to include it in each plugin that gets built.Really I think just firing resize when ever the main display (plugin-container-inner) is unhidden would work best, and likely is only a line or two of code.

caseypt:
* When the continue button is clicked on an infographic, call
the resize method of the plugin. Most implementations of this
method expect two arguments indicating the new size of the
element. Since nothing was actually resized, it made sense
to me to pass nothing to the method.

Closes #387

### DIFF
--- a/src/GeositeFramework/js/Plugin.js
+++ b/src/GeositeFramework/js/Plugin.js
@@ -525,6 +525,7 @@ require(['use!Geosite',
                     label: "Continue",
                     onClick: function() {
                         pluginModel.set('displayHelp', false);
+                        pluginObject.resize();
                     }
                 }, buttonnode);
             }


caseypt:
@mapossum Can you take a look and verify that this change will meet your need? I looked at what you did [for the multiSelect plugin](https://github.com/CoastalResilienceNetwork/multiSelect/commit/ec159af181e296e06af0c89ecff00f7d542a4680), and I think this should do the same thing, but from the framework. Thanks

mmcfarland:
+1 Looks good to me.

kdeloach:
+1

caseypt:
Merging this in @mapossum. Let us know if it's not what you are looking for.

PR #398, opened 2015-03-30
Fix empty tree

### LINKED ISSUE #391: Community Planning (aka Layer Selector) plugin breaks
Community Planning and Layer Selector broke for some sites after e99eae6f967.

Affected sites:

SE FL, NC have this plugin, also dev.maps.coastalresilience.org/test

Apparently the Layers will not be visible at all, but typing a space then backspace will show them.

kdeloach:
This fixes the case where the layer selector tree appears empty if there
is only 1 configured layer.

Fixes #391


mmcfarland:
How should I test this, just removing the services until there is just one?

kdeloach:
Yeah, you can just verify that the Custom Map Layers plugin is not empty. Here's what it look like on staging:

image

mmcfarland:
+1  Both single and multiple root nodes are showing up.

PR #400, opened 2015-03-30
Closes #388: Block plugins on second map and fix regression

### LINKED ISSUE #388: Blacklist apps on split view/map 2
Configuration for allowing a plugin to only be loaded and exist on Map 1.

caseypt:
* Add a new key to the region config file that allows users
to blacklist plugins from map 2
* Fixes a regression that caused subregions to be activated
then quickly deactivated if the subregion was launched from
a scenario and split screen mode was active/activated.

### DIFF
--- a/src/GeositeFramework/App_Data/regionSchema.json
+++ b/src/GeositeFramework/App_Data/regionSchema.json
@@ -114,6 +114,11 @@
             "type": "array",
             "items": {"type": "string"}
         },
+        "map2PluginBlacklist": {
+            "type": "array",
+            "items": {"type": "string"},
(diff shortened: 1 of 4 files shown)


caseypt:
Custom layer selector is currently blocked on map 2:

selection_081

caseypt:
To test the regression fix, open up the app, activate split screen mode, then launch the Galveston **scenario**. The Galveston subregion should be active on the first map pane.

mmcfarland:
+1 Split screen prevents custom_layer, and activating a subR-Scenario in SS works as expected.

PR #402, opened 2015-04-03
Hide subregion polygons at zoom level

### LINKED ISSUE #392: Hide subregion polygons at high zoom levels
When zooming in via the map or "zoom to extent" from the layer selector, it makes sense to hide the region polygons once a certain threshold has been reached.

mmcfarland:
Allow hiding the subregion shapes from the map when zooming in to a
customized zoom level.  Works sensibly with the toggle region plugin.

To test:
* Zoom in to level 8 and the polygons should disappear
* Zoom back out and they should reappear
* While visible, toggle them off
* Zoom in, stays off, zoom back out - still off
* If `hideByDefault` is set to true initially, zoom in and zoom out, still off.
* While zoomed in and non-visible, toggle back on - should stay visible while zooming below 8 still.

Phew.
Fixes #392

### DIFF
--- a/src/GeositeFramework/App_Data/regionSchema.json
+++ b/src/GeositeFramework/App_Data/regionSchema.json
@@ -69,6 +69,7 @@
             "type": "object",
             "required": false,
             "properties": {
+                "hideAtZoomLevel": {"type": "number", "required": false},
                 "color": {"type": "string", "required": false},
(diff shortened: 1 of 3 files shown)


caseypt:
+1

caseypt:
Everything tested out okay.

PR #403, opened 2015-04-03
Larger default font sizes for LP

### LINKED ISSUE #401: Text sizing for launchpad
From Zach:
I am arriving at 32 for a default icon size for the launchpad. Please hard code this size into explore the map and subregions.

I would also scale up the text underneath to be a bit bigger (match with font size for subtext on the left of each row). The header title and subtext on the left side of each row is good.

mmcfarland:
Makes the default size for fa-icons to be larger and the font size in the button matches the size of the sidebar text.

Fixes #401


caseypt:
+1

PR #411, opened 2015-04-14
Fix error on permalink mouseup

### LINKED ISSUE #409: Save & Share mouse up logs an error
the `_doSelect` method was refactored, and this must have gotten left behind.

https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/master/src/GeositeFramework/js/Permalink.js#L83

mmcfarland:
Call the new method which replaced _doSelect.  Correctly selects text now.

Fixes #409


caseypt:
+1

PR #414, opened 2015-04-22
Always activate plugin when plugin is selected.

### LINKED ISSUE #412: Scenario in Baja doesn't render layer_selector tree
Check out http://maps.coastalresilience.org/baja/

>The one-click maps appear well but the data tree for the layer selector apps are missing. Try Explore Cyclone Risk, minimize the Flood app and open Map Layers. Same with Infrastructure one-click – appears great, but when you open layer selectors for map layers and community planning the data trees are gone. 

Appears to be the filter -> space -> delete bug.

caseypt:
In an earlier attempt to decrease the number of events
and function calls related to activating the layer selector
and plugins in general, a check was added to see if a
plugin needed activation before activating it.

However, this did not account for plugins that are minimized
They are active when they are selected by the user, but still
need their activate function called when they are clicked.

Removing this check ensures that a plugin's activation function
is always called when it is selected.

Fixes #412

### DIFF
--- a/src/GeositeFramework/js/Plugin.js
+++ b/src/GeositeFramework/js/Plugin.js
@@ -142,10 +142,7 @@ require(['use!Geosite',
                     }
 
                     this.set('active', true);
-
-                    if (!active) {
-                        this.get('pluginObject').activate();
-                    }
+                    this.get('pluginObject').activate();
                 } else {
                     this.get('pluginObject').deactivate();
                     this.trigger('plugin:deselected');


caseypt:
**Basic test**
- Open the app and activate the layer selector (or a layer selector descendant).
- Activate a layer.
- Minimize the plugin
- Create a saved state.
- Launch the saved state in a new tab.
- Click the layer selector (which should be active, but minimized). The layer tree should be there and the state should be correct.

**Baja test**
Test with http://maps.coastalresilience.org/baja/ settings. Easiest way I found to do this was clone the repo for the site. Copy the `flood_slr` and `layer_selector` folders from that repo into the dev plugin folder. Then update the dev `region.config` to use the baja config. The behavior described in #412 should be fixed.

mmcfarland:
+1

PR #415, opened 2015-04-22
Split screen clear for scenarios

### LINKED ISSUE #413: Split view should be reset as new scenarios are launched
With the Infrastructure one-click calling split view, when I then go back to Get Started and choose Explore the map split view is stuck on. Seems we still have issue with the previous state not clearing.

mmcfarland:
Ensure that the splitScreen attribute is present in the state string even
if the feature hasn't been activated yet - use the default value (false).

Fixes #413


caseypt:
+1

Issue #423, opened 2015-10-05
Create Print CSS to hide all app features


mmcfarland:
Doing a browser print currently doesn't produce a valuable artifact and we're going to start allowing plugins to provide their own print CSS.  As an initial step, add a print style sheet to the app that would essentially produce a blank sheet when printed.  This may be as simple as hiding `body`, but if so, as a test, see if elements within body can be overridden (from the perspective of a plugin) to show elements within it.

PR #425, opened 2015-10-06
Hide all page content on print, with exceptions

mmcfarland:
By adding a wide-ranging print media CSS selector to hide page content, we
set the stage for allowing active plugins to develop their own print
capabilities.

Connects #423

### DIFF
--- a/src/GeositeFramework/GeositeFramework.csproj
+++ b/src/GeositeFramework/GeositeFramework.csproj
@@ -141,6 +141,7 @@
     <Content Include="css\jquery.dropdown.css" />
     <Content Include="css\jquery.mCustomScrollbar.css" />
     <Content Include="css\pageguide.min.css" />
+    <Content Include="css\print.css" />
     <Content Include="css\TinyBox2.css" />
     <Content Include="Error.html" />
     <Content Include="Error404.html" />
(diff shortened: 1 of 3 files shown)


mmcfarland:
To test:
* When the app loads, Ctrl-P to print preview and see it is blank
* Open a plugin so the container dialog is open and print again, see that the print preview shows the contents.

You can also use Emulation -> Media -> CSS Meda = Print in DevTools
devtools

kdeloach:
I noticed these issues with the print view:
1. There are margins around the top and side
2. The map layers are not visible (don't know if this can be helped)

image

kdeloach:
+1 Tested.

mmcfarland:
Hmm, those basemap tiles are not supposed to be showing up, I'll have to revisit.

kdeloach:
Here's what I see when emulating print mode in Chrome:

image

mmcfarland:
Updated css to be more explicit on map tile hiding, individually.  Looking for something like this when a plugin is opened and print previewed:
tncprint

caseypt:
Are plug-in developers going to want to show the map?

mmcfarland:
They're going to be given a special "print map" that is set to a fixed h/w that they can control layers of, and if they want it shown at all.  The layer selector's print button will divert to the existing map export dialog.

kdeloach:
:+1:

caseypt:
+1. Tested.

Issue #424, opened 2015-10-05
Update plugin config to specify a print bool setting


mmcfarland:
Plugins will be allowed to supply their own print CSS which will be injected into the page when a plugin print button (in the app container) is activated.  Update the schema for plugins to allow this setting, and have the framework draw a print button in the container header when it is present.

PR #429, opened 2015-10-12
Add print setting and button

mmcfarland:
Plugins can specify that they want a print event for their particular app.
The framework provides the UI element and a notification for when the
plugin is in a state for print to be called.

Connects #424 

##### To test:
* Open the sample Identify Point plugin
* Observe that a print button is added to the UI Container
* Clicking the print button will activate the print preview dialog
* Other plugins do not show the print icon

### DIFF
--- a/src/GeositeFramework/js/Plugin.js
+++ b/src/GeositeFramework/js/Plugin.js
@@ -374,7 +374,8 @@ require(['use!Geosite',
                 bindings = {
                     title: pluginObject.toolbarName,
                     id: containerId,
-                    isHelpButtonVisible: isHelpButtonVisible(view)
+                    isHelpButtonVisible: isHelpButtonVisible(view),
+                    hasCustomPrint: pluginObject.hasCustomPrint
                 },
                 $uiContainer = $($.trim(N.app.templates['template-plugin-container'](bindings))),
                 calculatePosition = function ($el) {
@@ -384,7 +385,6 @@ require(['use!Geosite',
                     };
                 };
 
-            $uiContainer = $($.trim(N.app.templates['template-plugin-container'](bindings)));
             view.$uiContainer = $uiContainer;
 
             $uiContainer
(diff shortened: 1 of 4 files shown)


src/GeositeFramework/js/Plugin.js
```diff
@@ -384,7 +385,6 @@ require(['use!Geosite',
                     };
                 };
 
-            $uiContainer = $($.trim(N.app.templates['template-plugin-container'](bindings)));
```
mmcfarland: This line was repeated from 380, above.

src/GeositeFramework/sample_plugins/identify_point/main.js
```diff
@@ -19,6 +20,11 @@ define(["dojo/_base/declare", "framework/PluginBase"],
                 var text = "You clicked on latitude " + mapPoint.getLatitude() + " longitude " + mapPoint.getLongitude(),
                     identifyWidth = 300;
                 processResults(text, identifyWidth);
+            },
+
+            beforePrint: function(printDeferred) {
+                // Prepare plugin markup for print...
+                printDeferred.resolve();
```
kdeloach: Consider making this the default for all plugins?

kdeloach:
+1

src/GeositeFramework/sample_plugins/identify_point/main.js
mmcfarland: Good idea.

caseypt:
+1. Tested.

caseypt:
Also, just FYI. Here's what I see in my print preview:

screen shot 2015-10-13 at 9 59 41 am

mmcfarland:
Added default print resolve in https://github.com/mmcfarland/GeositeFramework/commit/80d34e43326c77866fc4268a78f236d71e38eeac

Issue #426, opened 2015-10-09
Inject plugin supplied print css on print


mmcfarland:
Plugins wanting to participate in the local print operation specified by #424 will supply a print css file which the framework will inject into the page.  [It may be difficult to detect](http://stackoverflow.com/questions/18325025/how-to-detect-window-print-finish) when print operation is completed (the dialog closed) so we should probably just assume the file will be loaded until the next print operation, and clear anything out from a previous operation.

I think the flow is:

1. User clicks container print button
1.  Framework injects or enables the print css for that plugin
3. The Framework notifies the plugin, so it can do whatever element arranging it wants (probably synchronously, so that...)
4.  After the function returns, the framework calls `window.print()`

PR #430, opened 2015-10-12
Add plugin print css

mmcfarland:
Plugins can specify the css that they want enabled during plugin print.

##### To Test
* Open the Identify Point sample plugin, get passed the Info Graphic screen and click print
* The print preview should show the info graphic
* Open Split View and have both maps have an active Identify Point plugin
* Print one of the two plugins and ensure that only one is print previewed

Note that I didn't spend much time on actually positioning the graphic for print nicely, just demonstrating that plugins have the ability to do so.

Only the last three commits are relevant to this PR

Connects #426

### DIFF
--- a/src/GeositeFramework/js/Plugin.js
+++ b/src/GeositeFramework/js/Plugin.js
@@ -404,29 +404,31 @@ require(['use!Geosite',
                 .find('.plugin-print').on('click', function() {
                     var pluginDeferred = $.Deferred(),
                         parseDeferred = $.Deferred(),
-                        pluginPath = model.get('pluginSrcFolder') + '/print.css',
-                        printCssClass = 'plugin-print-css';
+                        pluginCssPath = model.get('pluginSrcFolder') + '/print.css',
+                        printCssClass = 'plugin-print-css',
+                        oppositePaneHideCssPath = 'css/print-hide-map' +
+                            (paneNumber === 0 ? 1 : 0) + '.css';
 
-                    // Clear all .plugin-print-css
+                    // Any previous plugin-prints may have left their print CSS loaded
+                    // clear them and any pane hiding css prior to this new print operation
                     $('.' + printCssClass).remove();
 
                     // Add the plugin css
-                    $('<link>', {
-                        rel: 'stylesheet',
-                        href: pluginPath,
-                        className: printCssClass
-                    }).appendTo('head');
+                    addCss(pluginCssPath, printCssClass);
+
+                    // Hide the possible same plugin from the opposite map pane
+                    addCss(oppositePaneHideCssPath, printCssClass);
 
                     // Give some time for the browser to parse the new CSS, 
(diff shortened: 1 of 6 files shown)


src/GeositeFramework/js/Plugin.js
```diff
@@ -417,6 +445,13 @@ require(['use!Geosite',
             model.set('$uiContainer', $uiContainer);
         }
 
+        function addCss(path, className) {
+            $('<link>', {
+                rel: 'stylesheet',
+                href: path,
+                'class': className
+            }).appendTo('head');
+    }
```
kdeloach: Needs a tab here.

kdeloach:
The map disappears if you follow these steps:
1. Switch to Map 2 (after page loads)
2. Open Identify Point plugin and click print icon
3. Switch to Map 1

kdeloach:
Not sure if this is intentional, but the plugin text does not appear in the print out.

image

image

mmcfarland:
Looking into the disappearing map, good catch.

mmcfarland:
Sadly, the map disappears if you follow the same sequence but just do a Ctrl-P to bring the print dialog up and not interact with any of the new code or plugins.

kdeloach:
Will test to see if this issue exists on master.

kdeloach:
Confirmed this issue is present on master. I'll open an issue.

kdeloach:
+1

caseypt:
After reviewing this, this comment may be more relevant here: https://github.com/CoastalResilienceNetwork/GeositeFramework/pull/429#issuecomment-147723621.

Do we want the selected location in the print preview?

caseypt:
When printing with split screen maps enabled, the infographic is cut off. Not sure what the solution is for this.

Map 1:
screen shot 2015-10-13 at 10 19 06 am

Map 2:
screen shot 2015-10-13 at 10 19 52 am

caseypt:
Regarding ^, maybe, in addition to hiding the second map, expand the width of printed map?

mmcfarland:
Sorry, I should have been more clear.  I'm just using the Sample Plugin to test the workflow, we're not on the hook for it to print anything reasonable and it's not deployed anywhere.  I didn't really spend any time on the print.css for this plugin to make the output look good, mostly just looking to see that it is loaded and printable.

mmcfarland:
Although the pane width issue may be a big deal.  I'll try adjusting the pane width in the framework print css as you suggest

mmcfarland:
Updated whitespace, comments and split screen print width in https://github.com/mmcfarland/GeositeFramework/commit/d0660020113da45d4f628f3f98943f066712bbf4

mmcfarland:
Might hold off - there is a difference between real print preview and the emulate CSS media option, which I was using.

mmcfarland:
Ok, updated and ready for testing.

caseypt:
+1. Looks good. Printing from either map produces a good result.

Issue #427, opened 2015-10-09
Allow print preview map for plugins


mmcfarland:
If plugins specify they want to print, they can also specify if they require a custom print map.  The framework will supply an empty map object to the plugin, which is hidden by normal css rules.  The map should be in an overlay container that has instructions on panning the map and a print button that will call `window.print()`.  It will be the plugin's responsibility to add layers to the map and set the size layout (typically a fixed size reasonable for printing, like 500x500). 

The the map is requested in the config, the sequence described in #426 will branch and not immediately call the print function.

The main map will remain hidden and can be exported by the traditional option, as described in #428

PR #432, opened 2015-10-20
Print map preview

mmcfarland:
Add a print preview map to plugins which request one.  The plugin is responsible for setting the layers of  the map, but lets the user select a zoom and extent that will likely results in a printable map.  The preview map is added to a sandbox element which is only available in `@media print`.  The plugin is also passed a reference to this sandbox so they can place additional elements and construct their own print.css to manage that layout.

Connects #427 
To Test:

The Sample Identify Plugin has been kitted out to support all the new features.
* Launch the Identify Plugin
* Click new Print button (fig 1)
* Navigate the preview map (fig 2)
* Click print to see your map & also a custom style graphic (fig 3)

*fig 1:*
print button

*fig 2:*
preview

*fig 3:*
output print

### DIFF
--- a/src/GeositeFramework/js/Plugin.js
+++ b/src/GeositeFramework/js/Plugin.js
@@ -464,7 +464,9 @@ require(['use!Geosite',
             var mapMarkup = N.app.templates['template-map-preview']({ pluginName: pluginObject.toolbarName }),
                 $mapPrint = $($.trim(mapMarkup)),
                 $printPreview = $('#print-preview-sandbox'),
-                mapReadyDeferred = $.Deferred();
+                mapReadyDeferred = $.Deferred(),
+                mapHeight = pluginObject.previewMapSize[1],
+                mapWidth = pluginObject.previewMapSize[0];
 
             // If the plugin is not set up for map print preview, don't set up a map
             // and resolve any pending print-preview map operations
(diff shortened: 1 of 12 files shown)


mmcfarland:
Note that the sample output isn't supposed to be meaningful, it just illustrates the control a plugin has over its print function.

src/GeositeFramework/js/Plugin.js
```diff
@@ -446,6 +460,56 @@ require(['use!Geosite',
...
+                        // the plugin can mess with it's positioning among its other elements
+                        $(map.container).detach().appendTo($printSandbox);
+                        TINY.box.hide();
+                        $printPreview.hide();
+                        previewDeferred.resolve();
+                    });
+                }
+            });
+
+            return mapReadyDeferred;
```
mmcfarland: The `esriMap` isn't able to be constructed until the container element is on the DOM, which doesn't happen until `openjs` executes, hence the need for this additional deferred.  It's resolved when the map is constructed and is passed the map as an argument.

src/GeositeFramework/js/Plugin.js
```diff
@@ -446,6 +460,56 @@ require(['use!Geosite',
...
+                fixed: true,
+                maskopacity: 40,
+                openjs: function () {
+                    // Remove any calculated size so that it contains the full map
+                    $('#print-preview-container').css({ height: 'initial', width: 'initial' });
+
+                    // Set the supplied height & width on the map
+                    $('#plugin-print-preview-map').css({ height: mapHeight, width: mapWidth });
+
+                    var map = new esri.Map('plugin-print-preview-map', { basemap: 'topo' });
```
caseypt: Consider using the active basemap and extent/center of the current map as defaults for the print map.

caseypt:
+1. Tested in Chrome, Firefox, and IE. This is a complicated workflow. Nice implementation and use of deferreds.

src/GeositeFramework/js/Plugin.js
```diff
@@ -424,10 +428,20 @@ require(['use!Geosite',
...
+                        // is used as a print preview box.
+                        pluginObject.beforePrint(pluginDeferred, $printSandbox, previewMap);
 
-                    $.when(pluginDeferred, parseDeferred).then(function() {
-                        window.print();
+                        // Exectue the browser print when the plugin and print preview (if used)
+                        // have responded, as well as a slight delay for css parsing.
+                        $.when(pluginDeferred, parseDeferred, previewDeferred).then(function() {
+                            window.print();
+                        });
```
kdeloach: If these interactions become any more complex, consider refactoring into a linear sequence (if possible). Ex:

```js
setupPrintableMap(...)
    .then(parseNewCss())
    .then(pluginObject.beforePrint(...))
    .success(window.print);
```

kdeloach:
+1 This feature looks great.

PR #433, opened 2015-10-21
Print enhancements

mmcfarland:
Features and fixes based on previous PR comments and demo with client.  

Test:

* Preview map should have basemap and extent of parent map
* Header popovers have a fixed width (not great for small screens for now)
* Print button visibility can be controlled from inside a plugin

Connects #427 (additionally)


caseypt:
+1. Tested. Nice improvements.

Issue #428, opened 2015-10-09
Layer Selector container print initiates map export


mmcfarland:
The Layer Selector plugin will have special behavior for its container print button, which will initiate the standard existing export dialog.

PR #434, opened 2015-10-22
Initiate Export custom print on layer selector

mmcfarland:
The layer selector specifies a custom print, but in reality just triggers
an export for the map pane it belongs to.

Test:
Each pane has it's own map and layer selector, so it's easiest to test in split screen mode.
- Open split screen and make the basemaps and extents obviously different
- Open the layer selector in either and click the new print button
- Check that the produced pdf is actually from the pane containing the plugin instance you used
- Repeat for the other pane.

Connects #428

### DIFF
--- a/src/GeositeFramework/js/PluginBase.js
+++ b/src/GeositeFramework/js/PluginBase.js
@@ -60,7 +60,10 @@ define(["dojo/_base/declare",
             validate: function () { return true; },
 
             // Auto-resolve the print deferred if the plugin does not implement the function
-            beforePrint: function (printDeferred) { printDeferred.resolve();  },
+            // printDeferred: deferred object to resolve when the printing can commence
+            // $printSandbox: DOM element which the framework provides for printable element arrangement
+            // previewMap: an ESRI map object for the print preview, if `usePrintPreviewMap` is true
+            beforePrint: function (printDeferred, $printSandbox, previewMap) { printDeferred.resolve();  },
 
             // Called when switching from infographic to the primary view or vice versa.
             onContainerVisibilityChanged: function (visible) {},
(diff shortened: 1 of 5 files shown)


src/GeositeFramework/js/PluginBase.js
```diff
@@ -60,7 +60,10 @@ define(["dojo/_base/declare",
             validate: function () { return true; },
 
             // Auto-resolve the print deferred if the plugin does not implement the function
-            beforePrint: function (printDeferred) { printDeferred.resolve();  },
+            // printDeferred: deferred object to resolve when the printing can commenct
```
caseypt: Typo: commenct -> commence.

src/GeositeFramework/plugins/layer_selector/main.js
```diff
@@ -170,6 +171,13 @@ define([
             subregionDeactivated: function(subregion) {
                 this.clearAll();
                 this.initialize(null, 'main');
+            },
+
+            beforePrint: function(printDeferred) {
+                // We can short circuit the plugin print chain by simply
+                // not resolving this deferred object.  Instead, trigger
+                // and export dialog for this pane.
+                this.app.dispatcher.trigger('export-map:pane-' + this.app.paneNumber);
```
caseypt: Does it make sense to call `printDeferred.reject()` here?

mmcfarland: Definitely.  I looked for such a method but didn't find it for some reason.

caseypt:
I get this error when I initiate a print.

```
GET http://localhost:54633/plugins/layer_selector/print.css 404 (Not Found)
```

Not a big deal, but we may want to capture it since (I think) that stylesheet will never exist.

caseypt:
+1. Tested. Nice work.

mmcfarland:
That error is from the browser when I insert a `<link>` tag for the custom print css, don't think it can be captured.  We could add an empty print.css, but not sure if that would be more confusing.  Do you think it's worth doing an ajax request to get the file first and only add the tag if it's not a `404`?

mmcfarland:
Updated from comments.

caseypt:
Sorry. Just saw that comment. I think the empty CSS file with a comment is good solution.

Issue #435, opened 2015-10-22
Turn off Identify from layer selector


mmcfarland:
From Task 6 of the current contract:

> Consensus among the extended Coastal Resilience development team was that Identify functionality should be at the plugin level. Azavea will make the Identify function optional and will turn it off by default at the framework level.

PR #437, opened 2015-10-27
Make Identify optional and off by default

mmcfarland:
TNC would like to transition identify results into the plugins to give
more meaningful results, as the framework version is inherently
context-free and doesn't do a great job of showing meaningful attributes.

* Make Identify a framework option in region.json via `identifyEnabled`
* `identifyEnabled` is false by default

##### Test:
* Set `identifyEnabled` to `true`: Map clicks should attempt an identify
* Remove `identifyEnabled`: Map clicks should not attempt an identify

You'll have to bounce your app or kill the dev IIS process between tests because the region data is cached on app start.

Connects #435

### DIFF
--- a/src/GeositeFramework/App_Data/regionSchema.json
+++ b/src/GeositeFramework/App_Data/regionSchema.json
@@ -151,6 +151,10 @@
             "type": "array",
             "items": { "type": "string" }
         },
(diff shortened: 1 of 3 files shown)


kdeloach:
+1 Tested

Issue #439, opened 2015-10-29
Print: Give map reference only when map is fully loaded


mmcfarland:
Don't call `beforePrint` on a plugin until `map.loaded = true`

PR #444, opened 2015-11-06
Ensure print map is loaded before handoff

mmcfarland:
Plugins may have errors accessing certain map features if `map.loaded` is
not `true`.  Don't resolve the deferred for map ready until the map is
loaded.

Connects #439 

Test:
Open the sample identify plugin, click on the map and then hit the print button.  The preview map should be displayed as usual, but you can test that `main.js/beforePrint:mapObject.loaded === true`


kdeloach:
+1

Issue #441, opened 2015-10-29
Print: Allow sandbox layout to exceed 1 page


mmcfarland:
Currently, print media is somehow specifying that the print be cut off at one page.

Could it be 
* the absolute position of the sandbox contents?
* restricted body size?
* would adding css page breaks help?

Issue #440, opened 2015-10-29
Print: Attempt to hint at print orientation


mmcfarland:
Try to see if a browser can be hinted to change the orientation, and if so, allow the plugin to specify

http://stackoverflow.com/questions/138422/landscape-printing-from-html

PR #445, opened 2015-11-06
Print page layout

mmcfarland:
Shows how to set page orientation (I believe this only works in Chrome, it's a semi-supported standard).  Also, let's overflow page elements print on multiple pages.

Connects #441 
Connects #440 

Test:
Print the sample print dialog from chrome. It should preview in landscape and there should be two pages of content.

### DIFF
--- a/src/GeositeFramework/css/print.css
+++ b/src/GeositeFramework/css/print.css
@@ -5,6 +5,9 @@
         visibility: hidden;
     }
 
+    html, body {
+        overflow: visible !important;
+    }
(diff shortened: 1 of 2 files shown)


kdeloach:
Not sure if this worked for me. I'm using Chrome 45.0.2454.85.

image

kdeloach:
Oh, I forgot to click the plugin print button first. Here's what I see now:
image

kdeloach:
+1

Issue #420, opened 2015-06-02
Metadata links don't work in subregion


mmcfarland:
From Laura Flessner (6/2/15):

> Links to metadata or external sources within the blue info icons for a dataset don’t work when you’re within a region.  For example, in NC http://dev.maps.coastalresilience.org/northcarolina/     if “explore the map” is chosen from the Launchpad > open community planning app > expand Roanoke island folder > click blue icon next to Potential Wetlands (3rd layer down) > this link works fine.  However, if you select the “Roanoke island region” from the launch pad and perform the same steps, the link is no longer live.  I tested other geographies both on dev and production and this is also the case.

laflessn:
Further, it looks like nothing that pops up in a window outside of the app interface works.  The opacity sliders don't work via the orange i buttons if clicked within a region .... it mouse pointer doesn't recognize the pop up window and simply shifts the base map beneath it.

mmcfarland:
This looks like maybe a z-index issue when in SubRegion mode.

PR #450, opened 2015-12-08
Replace subregion border div w/ multiple elements

caseypt:
* Instead of using one div to create the red subregion border
element, use multiple divs to piece together a border. This is
a bit more complicated, but has the effect of not covering up
other elements and causing z-index issues.

**Testing instructions**
- Activate a subregion. Verify that the red border is present, looks correct (check in various browsers) and that no DOM elements are covering up the map. 
- Leave the subregion. Verify that the red border disappears.
- When a subregion is active, open an instance of the layer selector and click one of the info buttons next to a layer. Verify that you can interact with the pop-up that appears.

Connects to #420


mmcfarland:
+1 tested, great solution.

Issue #278, opened 2014-12-11
Particular Tiled Map Service does not show legend


mmcfarland:
1. http://maps.coastalresilience.org/us/
1. Open Layer selector and add Coastal Resilience -> Social & Economic -> Population
1. Layer is loaded but the legend does not appear

http://services.coastalresilience.org:6080/arcgis/rest/services/US/Social_and_Economic/MapServer/legend

The legend is published, but it might it's at the service not the layer level.  Might be responsible.

PR #452, opened 2015-12-09
Add support for ArcGIS tile layer legends

caseypt:
When adding services to the map, track ESRI tile layers in addition to dynamic layers. This fixes an issue that prevented legends for tile layers from appearing on the map. The id for the layer must
be specified in the layer config for the legend to appear.

**Testing instructions**
- Verify that legends appear for some of the layers in the state folders and for layers in the "Coastal Resilience (National)" group, which were added as a sample in this PR.

Connects to #278


caseypt:
Superseded by https://github.com/CoastalResilienceNetwork/GeositeFramework/pull/454

Issue #449, opened 2015-12-03
WMS Layers with Legends published should show in legend


mmcfarland:
Currently, WMS layers are loaded but the WMS Legend, if published, is not used.  Check if either the JSAPI now supports WMS legends, or otherwise get layer legends showing.  They layer selector is being rewritten imminently, so this doesn't have to be a seamless integration, but the new code will need to know the technique.

caseypt:
WMS legends are supported through the JS API. Here's a simplified example. Paste the follow snippet into the ESRI [developer sandbox](http://developers.arcgis.com/javascript/sandbox/sandbox.html?sample=layers_wms). The legend for the cities layer will show up at the bottom of the legend control.

```
<!DOCTYPE html>
<html>
<head>
  <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
  <!--The viewport meta tag is used to improve the presentation and behavior of the samples
    on iOS devices-->
  <meta name="viewport" content="initial-scale=1, maximum-scale=1,user-scalable=no">
  <title>Map with legend</title>

  <link rel="stylesheet" href="http://js.arcgis.com/3.12/dijit/themes/claro/claro.css">
  <link rel="stylesheet" href="http://js.arcgis.com/3.12/esri/css/esri.css">

  <style>
    html, body {
      height: 97%;
      width: 98%;
      margin: 1%;
    }

    #rightPane {
      width: 20%;
    }

    #legendPane {
      border: solid #97DCF2 1px;
    }
  </style>

  <script src="http://js.arcgis.com/3.12/"></script>
  <script>
    var map;
    require([
      "esri/map", "esri/layers/FeatureLayer", "esri/layers/WMSLayer", "esri/dijit/Legend",
      "dojo/_base/array", "dojo/parser",
      "dijit/layout/BorderContainer", "dijit/layout/ContentPane", 
      "dijit/layout/AccordionContainer", "dojo/domReady!"
    ], function(
      Map, FeatureLayer, WMSLayer, Legend,
      arrayUtils, parser
    ) {
      parser.parse();

      map = new Map("map", {
        basemap:"topo",
        center: [-96.53, 38.374],
        zoom: 13
      });
      
      esriConfig.defaults.io.proxyUrl = "/proxy/";

      var rivers = new FeatureLayer("http://sampleserver3.arcgisonline.com/ArcGIS/rest/services/Hydrography/Watershed173811/MapServer/1", {
[...]

PR #453, opened 2015-12-09
WMS layer legend demo

caseypt:
Only 17cc62d is relevant. 

Demos WMS layer legend capability, though it's untested and shouldn't be merged.

screen shot 2015-12-09 at 3 58 12 pm

Connects to #449.


caseypt:
Closing since this shouldn't be merged, but it should be kept around for reference.

Issue #442, opened 2015-11-05
Update JS libraries


mmcfarland:
* [ ] Esri JavaScript API (including, if required, Dojo and ExtJS)
* [ ] Backbone
* [ ] Underscore
* [ ] jQuery

PR #457, opened 2015-12-11
Update JS

mmcfarland:
Updates Underscore, Backbone, ~jQuery, mCustomScrollbar

Connects #442


caseypt:
Missed a passing of options [here](https://github.com/mmcfarland/GeositeFramework/blob/feature/update-js/src/GeositeFramework/js/SubRegion.js#L267).

caseypt:
+1 besides the above error, which prevents subregions from activating. 

Verified that the core libraries have been updated and that the site is working.

```
Backbone.VERSION
"1.2.3"
_.VERSION
"1.8.3"
jQuery.fn.jquery
"1.9.0"
```

mmcfarland:
Good catch, thanks.

mmcfarland:
Updated with SubRegion options.

Issue #448, opened 2015-11-19
Improve Dojo/ESRI load times


mmcfarland:
Try to bundle the required dojo/esri library components and reduce the number of calls on start up.

PR #458, opened 2015-12-11
Add ESRI optimized JS bundle.

caseypt:
This commit adds a minified and bundled JS file containing Dojo, ESRI JS API, and all of the components of those two libraries that were are currently using on the app. It was created using the ESRI JS API Optimizer.

This reduces the number of requests to the ESRI CDN from 49 (current `develop`) to 5, a 90% decrease.

We should be able to get it down to two requests (`dojo.js`, `dojo_en.js`, lose `utils.js`, `geometry.js`, `agscommon.js`), but I believe it would require replacing uses of `esri.sampleNamespace.method` in the code with AMD require statements for `sampleNamespace` and use of `sampleNamespace.method` directly. I replaced some of these to get remove some requests, but there are many uses left to replace throughout the code.

**Testing instructions**
* Compare the number of requests on this branch with the number on `develop`, and verify that there are fewer requests. Make sure to do hard refreshes when switching between branches.
* Verify the site loads with no new errors and generally works.

Thanks to @mmcfarland for finding the optimizer.

Connects to #448


caseypt:
Rebased.

mmcfarland:
+1, esri calls are severly reduced.  Do you think you could make some backlog issues about additional performance ideas that we discussed?

* Remove EXTJS in layer rewrite or get it onto another faster CDN
* Concat framework JS
* Reduce calls to MapServer endpoints in Layer Selector load
* Anything else?

caseypt:
> Do you think you could make some backlog issues about additional performance ideas that we discussed?

Will do.

caseypt:
https://github.com/CoastalResilienceNetwork/GeositeFramework/issues/462
https://github.com/CoastalResilienceNetwork/GeositeFramework/issues/461
https://github.com/CoastalResilienceNetwork/GeositeFramework/issues/460

Issue #466, opened 2015-12-16
Plug-ins are not clickable on sites with many plugins


caseypt:
The updated scrollbar plug-in we are using is creating divs on top of the plug-in buttons, which is prevents the user from activating the plug-ins.

Deleting the #mCSB_1_dragger_vertical element fixes the issue.

PR #470, opened 2015-12-17
Update the css for recent scrollbar plugin

mmcfarland:
This fixes a bug where a new element was introduced but not managed by the
css and caused plugins not to be clickable.

Connects #466 

To test:
* Make the screen small enough that a scrollbar appears for the loaded plugins.
* Ensure you can still click on a plugin


kdeloach:
+1

Issue #468, opened 2015-12-16
Update ESRI JS bundle with additional plug-in dependencies


caseypt:
Add the plugins that are active in the Gulf Mex, Louisiana, and other regional sites to our demo project, and update the JS bundle with the additional plugin dependencies.

PR #474, opened 2015-12-22
Update JS bundle

caseypt:
* Add more components and libraries used by Geosite Framework
plugins to the JS bundle.
* Update JS requirements file to include new dependencies.

This reduces the number of requests for JS resources from arcgis.com from 155 to 6 for the gulfmex region site. This site includes almost all of the Geosite Framework plugins.

**Testing Instructions**:

- Build a local version of the Gulfmex Region site.
- Update the JS bundle used there to version 1.0.8.
- Verify that there are a reduced number of requests for JS resources from arcgis.com

Connects to #468


mmcfarland:
+1

Issue #471, opened 2015-12-18
Legend Flicker Issue


mmcfarland:
After recent commits (either JSAPI upgrade or bundling) the Legend flickers (as if redrawing itself every second) in some browsers.  It's been reported in Chrome on Windows and on Firefox in linux.  To reproduce, add any item from the layer selector to the map and observe the legend. 

Example: http://goo.gl/MY7Zvx

PR #475, opened 2015-12-28
Fix legend flickering.

caseypt:
Adding support for tile layer legends (#452) introduced a bug that made the legend flicker when tile layers were added to the map. This is because 1) tile layers fire multiple events when being added
to the map, which caused our manual layer redrawing to fire multiple times, and 2) our logic for determining the number of visible layers on the map did not properly handle tile layers.

Also, the legend was flickering in older version of the framework, most likely due to changes in the structure of legend elements returned by ArcGIS server, but it wasn't visible because it was happening
very frequently and quickly.

This commit attempts to fix the legend flickering by reducing the amount of times the legend is redrawn. It also attempts to make the logic of the manual legend redrawing hack clearer and easier to follow.

* Try to reduce the number of times updateLegend can be called repeatedly.
* Keep an explicit count of the number of legend nuggets rather than using the length of the legend nugget array. This was changed because we add legends for layers on the same service in groups now rather than breaking them up. This preserves the legend hierarchy and is a littler cleaner and less disruptive to the original structure 
* Rely on a map method and direct property of the layers for determining which layers are visible, so that the logic is more clear.
* Reduce the delay from when update legend is called again if the visible layer count doesn't match. Testing showed that the layers were restored more quickly than 1 second.

**Testing instructions**
* Add multiple layers to the map, and ensure the the legend does not flicker and that when you resize the legend window, the legend items move around to better fit the window's dimensions.
* Refresh the map, and add the following layers to the map (they each represent a different legend structure or layer type): 
- Florida > Social and Economic > Social Vulnerability
- Florida > Florida Infrastructure > Transportation > Airport
- Coastal Resilience (National) > Social and Economic > Population
* With the above layers active, verify the legend does not flicker and the items adjust as the legend window size changes.
* Add breaks points around line 309 in Map.js, and verify that the `updateLegend` function isn't called repeatedly when layers are added and removed. 

Connects to #471

### DIFF
--- a/src/GeositeFramework/js/Map.js
+++ b/src/GeositeFramework/js/Map.js
@@ -238,8 +238,13 @@ require(['use!Geosite',
         // body of `updateLegend` immediately looks at visible layers, while some may
         // still asynchronously be in queue to become visible. As a protection, call
         // `updateLegend` explicitly when a layer has finished adding.
-        dojo.connect(esriMap, 'onUpdateEnd', updateLegend);
-        dojo.connect(esriMap, 'onLayerAdd', updateLegend);
+        //
+        // Add legends for tile layers fires both of these events, which results in
+        // the legend flickering (because of our manual redraw), so make sure the
+        // function is not called repetitively.
+        var debouncedUpdateLegend = _.debounce(updateLegend, 100);
+        dojo.connect(esriMap, 'onUpdateEnd', debouncedUpdateLegend);
+        dojo.connect(esriMap, 'onLayerAdd', debouncedUpdateLegend);
 
         function updateLegend() {
             var services = esriMap.getLayersVisibleAtScale(esriMap.getScale()),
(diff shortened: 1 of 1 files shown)


kdeloach:
I was still able to trigger flickering with these steps:
1. Check "Florida/FEMA/Repetitive Loss Property" and "Costal Resillience (National)/Social and Economic/Population"
2. Uncheck all layers (or Clear All)
3. Re-check "Florida/FEMA/Repetitive Loss Property"

src/GeositeFramework/js/Map.js
```diff
@@ -283,20 +293,21 @@ require(['use!Geosite',
...
-            if (totalVisibleLayers !== legendNuggets.length) {
-                _.delay(updateLegend, 1000);
+            // Get all the layers on the map, and filter out graphic layers (i.e. subregion boundaries).
+            var totalVisibleLayers =_.filter(esriMap.getLayersVisibleAtScale(), function(layer) {
+                return !layer.graphics;
+            });
+
+            // Subtract one for the base layer.
+            if ((totalVisibleLayers.length - 1) !== nuggetCount) {
+                _.delay(updateLegend, 250);
```
kdeloach: Unclear why `updateLegend` calls itself with a delay. This may cause flickering to occur.

src/GeositeFramework/js/Map.js
```diff
@@ -238,8 +238,12 @@ require(['use!Geosite',
...
         // still asynchronously be in queue to become visible. As a protection, call
         // `updateLegend` explicitly when a layer has finished adding.
-        dojo.connect(esriMap, 'onUpdateEnd', updateLegend);
-        dojo.connect(esriMap, 'onLayerAdd', updateLegend);
+        //
+        // Add legends for tile layers fires both of these events, which results in
+        // the legend flickering (because of our manual redraw), so make sure the
+        // function is not called repetitively.
+        dojo.connect(esriMap, 'onUpdateEnd', _.debounce(updateLegend, 100));
+        dojo.connect(esriMap, 'onLayerAdd', _.debounce(updateLegend, 100));
```
kdeloach: Not sure of the intent, but this binds 2 unique debounced functions instead of debouncing all calls to `updateLegend`.

src/GeositeFramework/js/Map.js
caseypt: That was in there, I just shortened the time. Here's the relevant comment:

```
// Compare the total number of visible layers on the map to the number
// of legendNuggets to determine if the legend has gotten info for all
// visible layers. If not, recur with a delay.
//
// This is a hack, because sometimes when loading from browser cache,
// layerAdd events don't fire.
```

src/GeositeFramework/js/Map.js
caseypt: Thanks. I'll fix that.

src/GeositeFramework/js/Map.js
kdeloach: Yeah, just looks suspicious...

caseypt: Agreed.

caseypt: There would be no flickering if we didn't manually rebuild the legend.

caseypt:
PR updated. Fixed the debounce issue.

> I was still able to trigger flickering with these steps: ...

This should be fixed now as well.

kdeloach:
Much improved and I think this fixes the original issue #471. 

I found another flickering edge case that we may want to address in the future:
1. Check "Florida/FEMA/Repetitive..."
2. Check "Habitats/1880 Oyster Reefs"
3. Check "Habitats/Current Oyster Reefs"
4. Repeatedly check & uncheck "Habitats/1880 Oyster Reefs"

kdeloach:
+1

mmcfarland:
+1, Flicker free.

caseypt:
Thanks

PR #476, opened 2015-12-30
Merge develop into master

caseypt:
Includes fixes for legend flickering and an updated JS bundle that supports more framework plugins.


mmcfarland:
+1

Issue #482, opened 2015-12-31
Add support for drawing legends from JSON


caseypt:
Rather than using the default AGS legend elements, render a custom legend based off the legend JSON returned for a service.

Sample response: http://services.coastalresilience.org/arcgis/rest/services/New_Jersey/New_Jersey/MapServer/legend?f=pjson

PR #488, opened 2016-01-13

kdeloach:

Connects #482

### DIFF
--- a/src/GeositeFramework/Views/Home/Index.cshtml
+++ b/src/GeositeFramework/Views/Home/Index.cshtml
@@ -177,12 +177,51 @@
                 <div id="legend-container-<%= paneNumber %>" class="legend">
                     <div class="legend-header">Map Legend
                         <a class="legend-close" title="Hide Legend" href="javascript:;">_</a></div>
-                    <div class="legend-body">
-                        <div id="legend-<%= paneNumber %>"></div>
-                    </div>
+                    <div class="legend-body"></div>
                 </div>
                 <div class="subregion-tooltip"></div>
             </div>
+        </div>
+    </script>
+
+    <script type="text/template" id="template-legend-item-single">
+        <div class="legend-layer single">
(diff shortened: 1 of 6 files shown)


kdeloach:
There may be some styling issues remaining.

To test the "scale" legend items you can select the first layer called "Bathymetry".
image

kdeloach:
This logic may be something that we could use as a reference for the new layer selector https://github.com/CoastalResilienceNetwork/GeositeFramework/pull/488/files#diff-39eb965dbc0c1ad41b4e12dbbc16e00bR209. 

It isn't evident from this PR, but we *could* support lazily adding items to the legend now (before the layer is finished loading). I didn't add a "Loading..." UI for aesthetic purposes. 

I can elaborate on this in-person.

src/GeositeFramework/Views/Home/Index.cshtml
```diff
@@ -177,15 +177,54 @@
...
+            <% }) %>
+        </div>
+    </script>
+
+    <script type="text/template" id="template-legend-item-scale">
+        <div class="legend-layer scale">
+            <div class="legend-title"><%- layerName %></div>
+            <% _.each(legend, function(item) { %>
+                <div class="item">
+                    <img src="data:<%= item.contentType %>;base64,<%= item.imageData %>"
```
mmcfarland: This isn't in the wireframes, but I think we need to add the text label/value for the first and last values in the scale at the beginning/end of the series.

mmcfarland:
This looks great - can you target `develop` for the PR?  We still haven't made the changes to build that lets us make it the default.

kdeloach:
Created a new PR targeting develop.

src/GeositeFramework/Views/Home/Index.cshtml
kdeloach: Oh yeah, that makes sense.

PR #489, opened 2016-01-13

kdeloach:

Connects #482

### DIFF
--- a/src/GeositeFramework/js/Map.js
+++ b/src/GeositeFramework/js/Map.js
@@ -2,16 +2,16 @@
 /*global Backbone, _, $, Geosite, esri, Azavea, setTimeout, dojo, dojox */
 
 require(['use!Geosite',
-         'esri/dijit/Legend',
-         'esri/map',
-         'dojox/layout/ResizeHandle',
-         'framework/widgets/ConstrainedMoveable'
+         'dojo/Deferred',
+         'dojo/request/xhr',
+         'framework/Legend',
+         'esri/map'
         ],
     function(N,
+             Deferred,
+             xhr,
              Legend,
-             Map,
-             ResizeHandle,
-             ConstrainedMoveable) {
+             Map) {
     'use strict';
 
     function getSelectedBasemapLayer(model, esriMap) {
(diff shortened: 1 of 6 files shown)


caseypt:
Even though it's not directly related to the layer selector, should this PR be made against the new `layer-selector-v2` branch?

src/GeositeFramework/region.json
```diff
@@ -159,5 +159,14 @@
...
             ]
         }
-    ]
+    ],
+    "legend": {
+        "layers": [
+            {
+                "serviceUrl": "http://dev.services2.coastalresilience.org:6080/arcgis/rest/services/Gulf_of_Mexico/Florida/MapServer",
+                "layerName": "Bathymetry (m)",
+                "legendType": "scale"
```
caseypt: If a layer doesn't define a `legendType`, will it not have legend?

src/GeositeFramework/js/Map.js
```diff
@@ -201,132 +201,88 @@ require(['use!Geosite',
...
-                return !layer.graphics && layer.visible;
-            });
-
-            // Subtract one for the base layer.
-            if ((totalVisibleLayers.length - 1) !== nuggetCount) {
-                _.delay(updateLegend, 250);
-            }
+        // Filter out layers that are not visible at the current map scale.
+        // Adapted from the ESRI dijit legend source code. (Ref: _isLayerInScale)
+        function isLayerInScale(service, layer) {
```
caseypt: This may not be helpful, but there's a method on `map` to return visible layers: [getLayersVisibleAtScale](https://developers.arcgis.com/javascript/jsapi/map.html#getlayersvisibleatscale).

src/GeositeFramework/js/Map.js
```diff
@@ -201,132 +201,88 @@ require(['use!Geosite',
...
-                handle: $legendContainer.find('.legend-header')[0],
-                within: true
+            regionData = N.app.data.region,
+            id = 'legend-container-' + mapNumber,
+            legend = new Legend(regionData, id);
+
+        var getServiceLegend = (function() {
+            var cache = {};
+            // Memoize so only 1 request is generated per service.
+            var fetch = _.memoize(function(url) {
```
caseypt: Very cool.

src/GeositeFramework/region.json
caseypt: Now I see that it still does. Default looks good.

caseypt:
+1. Tested. Nice work.

mmcfarland:
> should this PR be made against the new layer-selector-v2 branch?

Shoot, that's a good idea.  We probably want to roll these out together and have them be tested together before we do a release for either.  Third time's the charm?

+1

src/GeositeFramework/region.json
kdeloach: It's a required field. Our version of jsonschema doesn't support enums, unfortunately, so this will default to "multiple" if the value is not one of the supported legend types.

src/GeositeFramework/js/Map.js
kdeloach: That function is poorly named, because it actually returns a list of all map services, with each service containing an array of visible layer ids. Unfortunately, visibleLayerIds also contains layers that do not exist in the layer selector, and also layers that are not visible on the map. I discovered that the ESRI dijit legend actually filters out these layers by using this logic so I copied it here.

kdeloach: Forgot to mention we use that function here https://github.com/CoastalResilienceNetwork/GeositeFramework/pull/489#discussion-diff-49731196R238.

kdeloach:
Superseded by #490.

src/GeositeFramework/js/Map.js
caseypt: Right. Forgot it was a pain to use. I had to do something like this to get actually visible layers:

```
var totalVisibleLayers =_.filter(esriMap.getLayersVisibleAtScale(), function(layer) {
    return !layer.graphics && layer.visible;
});
```

kdeloach: We should totally use this if it completes the same objective.

PR #490, opened 2016-01-14

kdeloach:

Connects #482


kdeloach:
Updated the "scale" legend template to display the first and last values.

PR #496, opened 2016-01-15
AgsMeasure updated to work with current AGS Javascript

mapossum:
This file was creating errors and the measure tool was not working.  Looks like it was caused by referencing functions not present in the current version of ArcGIS for Javascript.

### DIFF
--- a/src/GeositeFramework/plugins/measure/AgsMeasure.js
+++ b/src/GeositeFramework/plugins/measure/AgsMeasure.js
@@ -1,4 +1,4 @@
-﻿
+
 define(["jquery",
         "use!underscore",
         "esri/geometry/Polyline",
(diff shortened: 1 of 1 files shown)


src/GeositeFramework/plugins/measure/AgsMeasure.js
```diff
@@ -353,13 +359,13 @@ define(["jquery",
                 
                 // If the user has drawn the polygon ring anti-clockwise, reverse the ring
                 // to make it a valid esri geometry.
-                if (!Geometry.isClockwise(_points)) {
-                    _points = _points.reverse();
-                }
+                //if (!Polygon.isClockwise(_points)) {
```
mmcfarland: Looks like this method has been moved from the class to the object instance, and without it it's producing negative areas for counter-clockwise polygons.  The if statement could be brought back with something like: 

```javascript
if (!polygon.isClockwise(_points) {
    _points = _points.reverse();
}
```

mmcfarland:
Thanks for the patch @mapossum - let me know if you can modify the PR with the commented fix.  If not, we can add an additional commit later.

Issue #492, opened 2016-01-14
Enable truncated legend group


mmcfarland:
For legends with `n` number of elements (legendType == 'multiple'), truncate the list to `n` and provide a "show more" link to expand the grouping.  Unsure of what `n` should be initially, but lets play around with 4 and see how it works.

PR #498, opened 2016-01-18
Make excess legends break visibility toggleable.

caseypt:
* If the legend for an individual layer contains more than 5 breaks/classes, hide the additional items to conserve space.
* Add an element to allow the user to toggle the visibility of the extra items.

**Testing instructions**
- Turn on a layer that has more than 5 breaks/classes. Louisiana > Species > ESI Waterfowl (LDWF 2001) is a good example.
- Verify that only 5 breaks are showing, and that there is an option to show more breaks that toggles the amount of visible breaks.
- Add another layer with more than 5 breaks and verify that you can toggle this layer, and that it doesn't affect the other layer with lots of breaks. Try Florida > Species > Bottlenose Dolphin Abundance.
- Verify that legends with exactly 5 items do not generate a toggle option. Louisiana > Salinity > Low Salinity Season (ppt) has 5.

screen shot 2016-01-18 at 12 26 01 pm
screen shot 2016-01-18 at 12 26 07 pm

Connects to #492

### DIFF
--- a/src/GeositeFramework/Views/Home/Index.cshtml
+++ b/src/GeositeFramework/Views/Home/Index.cshtml
@@ -199,15 +199,28 @@
     <script type="text/template" id="template-legend-item-multiple">
         <div class="legend-layer multiple">
             <div class="legend-title"><%- layerName %></div>
-            <% _.each(legend, function(item) { %>
-                <div class="item">
+            <% _.each(legend, function(item, index) { %>
+                <div class="item <% if (index > 2 && index <= (legend.length - 3)) { %>extra<% } %>">
                     <img src="data:<%= item.contentType %>;base64,<%= item.imageData %>"
                         width="<%- item.width %>"
                         height="<%- item.height %>"
                         title="<%- item.label %>" alt="" />
                     <span class="item-label"><%- item.label %></span>
                 </div>
+                <% if (legend.length > 5 && index === 3) { %>
+                    <div class="item expand">
+                        <span>...</span>
+                    </div>
+                <% } %>
             <% }) %>
+            <% if (legend.length > 5) { %>
+                <div class="item expand">
+                    <a href="#" class="expand-legend">View <% print(legend.length - 5) %> more</a>
+                </div>
+                <div class="item extra collapse">
+                    <a href="#" class="expand-legend">View less</a>
+                </div>
+            <% } %>
         </div>
     </script>
 
--- a/src/GeositeFramework/js/Legend.js
+++ b/src/GeositeFramework/js/Legend.js
@@ -90,6 +90,20 @@ define(['use!Geosite',
             });
 
             this.$el.find('.legend-body').html($container.html());
+            this.assignLegendEvents();
+        },
+
+        assignLegendEvents: function() {
+            this.$el.find('.expand-legend').on('click', $.proxy(function(e) {
+                this.toggleExtraLegendItems(e);
+            }, this));
+        },
+
+        toggleExtraLegendItems: function(e) {
+            var $extraLegendControl = $(e.target),
+                $extraLegendItems = $extraLegendControl.parents('.legend-layer');
+
+            $extraLegendItems.toggleClass('show-extras');
         }
     });
 
--- a/src/GeositeFramework/js/Legend.js
+++ b/src/GeositeFramework/js/Legend.js
@@ -21,6 +21,7 @@ define(['use!Geosite',
                     layerName: layer.name
                 });
             }
+
             return null;
         }
     });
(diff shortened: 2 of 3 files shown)


src/GeositeFramework/js/Legend.js
```diff
@@ -66,23 +68,50 @@ define(['use!Geosite',
...
             this.$el.find('.legend-body').html($container.html());
+            this.assignLegendEvents();
+        },
+
+        assignLegendEvents: function() {
+            var self = this;
+
+            $('.expand-legend').click(function(e) {
+                self.toggleExtraLegendItems(e);
+            });
```
kdeloach: Consider using a live selector and binding this in the constructor. May also want to `this.$el.find` instead of global `$`.

src/GeositeFramework/js/Legend.js
```diff
@@ -66,23 +68,50 @@ define(['use!Geosite',
...
+        },
+
+        toggleExtraLegendItems: function(e) {
+            var $extraLegendControl = $(e.target),
+                $extraLegendItems = $extraLegendControl.parent().parent().find('.extra');
```
kdeloach: Consider using `.parents('.legend-layer')` in case the markup changes?

src/GeositeFramework/Views/Home/Index.cshtml
```diff
@@ -199,8 +199,8 @@
     <script type="text/template" id="template-legend-item-multiple">
         <div class="legend-layer multiple">
             <div class="legend-title"><%- layerName %></div>
-            <% _.each(legend, function(item) { %>
-                <div class="item">
+            <% _.each(legend, function(item, index) { %>
+                <div class="item <% if (index > 4) { %>extra<% } %>">
```
kdeloach: Consider showing the first few and the last few items (hide items in the middle instead of the end).

src/GeositeFramework/js/Legend.js
```diff
@@ -66,23 +68,50 @@ define(['use!Geosite',
...
+
+            $extraLegendItems.toggle();
+
+            if (($extraLegendItems).is(':visible')) {
+                $extraLegendControl.text('View less items');
+            } else {
+                $extraLegendControl.text('View more items');
+            }
```
kdeloach: If you toggle a class on the parent instead of toggling visibility (ex. `.show-extras`), we might be able to move this text into the template.

kdeloach: Ex.
```html
<div class="item expand">
    <a href="#" class="toggle-legend">View more items</a>
</div>
<div class="item extra collapse">
    <a href="#" class="toggle-legend">View less items</a>
</div>
```

```css
.show-extras .extra { display: block; }
.show-extras .expand { display: none; }
```

src/GeositeFramework/Views/Home/Index.cshtml
kdeloach: If we're able to do this-- might want to add a line or something to visually indicate this division.

kdeloach:
+1 Works well

src/GeositeFramework/js/Legend.js
caseypt: I've give this a go.

src/GeositeFramework/js/Legend.js
caseypt: Thanks. Tried this earlier but forgot about parent**s**.

src/GeositeFramework/js/Legend.js
caseypt: Thanks.

src/GeositeFramework/Views/Home/Index.cshtml
caseypt: What do you think @jfrankl? I think this makes sense, but wanted to check with you first.

jfrankl: I like the idea. We'll need to figure out how it should look—let me think on that some more.

src/GeositeFramework/js/Legend.js
caseypt: This was a good call. Added in bfd6d12.

src/GeositeFramework/Views/Home/Index.cshtml
caseypt: I have this working. Here's a simple (but maybe ineffective) way to show there are missing breaks in the middle:

screen shot 2016-01-19 at 9 16 45 am

caseypt: A lot of the layers have 5 breaks, so I made this kick in for any layer with more than 5 breaks.

kdeloach:
:+1:

src/GeositeFramework/Views/Home/Index.cshtml
caseypt: One more update to add the count: 
screen shot 2016-01-19 at 9 33 35 am

caseypt:
The latest:

screen shot 2016-01-19 at 9 33 35 am

Issue #495, opened 2016-01-14
Ensure ordering of legend items correspond to active layers


kdeloach:
> Ensure ordering of legend items as they correspond to active layers (sometime new legend items from plugins get added below existing items, other times they get added above – this even happens within layer selector - we need consistent rules for how items get added across plugins)

Ref: #483

mmcfarland:
The preferred behavior would be for the most recently added layer to be on the top of the legend, which should also correspond to the z-index of the layer.

kdeloach:
Can we wait on adding this until layer selector v2 is complete?

mmcfarland:
@kdeloach Sure

caseypt:
This seems to be somewhat tough to do without #559.

Issue #485, opened 2015-12-31
Don't show legend when no layers are on the map


mmcfarland:
Legend should only be activated with layers are activated (open and blank confuses users).

- Remove the legend icon next to the zoom-slider. When the legend is minimized by the user make the div small in the bottom-right corner. A click on the small div re-opens the legend.

PR #499, opened 2016-01-19
Implement auto-hiding and minimization of legend

caseypt:
* Allow users to minimize the legend when clicking the minimize icon in the legend header.
* If there are no active layers on the map, hide the legend.

Also removes the "legend display" plugin since it is no longer needed.

**Testing instructions**
- When the app loads, the legend should not appear.
- Add a layer, and the legend should appear.
- Click the minimize button in the legend header. The legend should minimize, and the bottom of the legend header should be in line with where the bottom of the legend itself used to be.
- Restore the header. The top of the header should be in the same place as it was when the app was started. 
- Move the legend around and retry the above two steps. 
- Minimize the legend, move it all the way to the top of the window and restore it. Verify that the legend header is not covered by the app title bar.
- Verify that the legend control near the zoom controls is gone.

**Note** I may wait to merge this in until there is a PR for #495, so that the number of times the legend is rendered needlessly can be reduced. 

Connects to #485.


mmcfarland:
+1, tested.

Issue #494, opened 2016-01-14
Dynamic legend dimensions


kdeloach:
> Make the legend div update it’s size dynamically to some agreed-upon min/max height/width.

Ref: #483

PR #501, opened 2016-01-20
Auto-resize legend

caseypt:
Only 0eea00e is part of this PR. Also, I'm still testing out edge cases.

* Resize the legend as layers are turned on so that the legend box grows to fit the additional legends without needing a scroll bar. This continues until the legend hits hard height and width limits. Then a scroll bar is added.

**Testing instructions**
- Add layers and ensure the legend resizes to display all the layers until it reaches `MAX_HEIGHT` and `MAX_WIDTH`.  A scroll bar should be added then.
- Manually make the legend really large. Add and remove layers. It should not resize. The idea here is that if the user manually resized the legend, we shouldn't mess with it.
- Make the legend small again, and add/remove layers. It should resize.

legend

Connects to #494


kdeloach:
+1 Tested

Issue #493, opened 2016-01-14
Verify plugins can still interact with legend


kdeloach:
> Easily prepend a div to the existing legend

See #483

PR #504, opened 2016-01-20
Make sure plugin legends are added to map legend

caseypt:
Only 554e531 is part of this PR.

* When we rewrote the legend, we accidentally dropped the plugin
legends from the container because we are clearing out the
container on render. To preserve the plugin legends, we add
container divs to the legend body, one for layer legends, and one
for plugin legends. Only the layer legend container is rendered
by the legend itself.
* Account for plugin legends when auto-resizing the legend.
* Exercise functionality by adding demo plugin legend to the
new layer selector.

**Testing instructions**
- Turn on the Layer Selector V2 plugin and verify the sample plugin legend is added to the map legend.

Connects to #493


mmcfarland:
+1

kdeloach:
+1

PR #506, opened 2016-01-20
Check for legend config before using

mmcfarland:
The region legend configuration is optional.  Don't fail if it is not
found.


kdeloach:
Thanks, that was an oversight.

Issue #509, opened 2016-01-20
Legend (v2) grows automatically off screen to the right


mmcfarland:
When the legend is pegged to the right edge of the screen, it can auto grow off the right side.

Issue #508, opened 2016-01-20
Legend (v2) grows when minimized and layers are added


mmcfarland:
When the new legend is minimized, add another layer and the legend body will grow until it's un-minimized.  The legends are added correctly at that time.

Issue #507, opened 2016-01-20
Legend (v2) flashes on startup


mmcfarland:
The new legend should be hidden by default via css while it loads so it doesn't appear and then hide on page load.  Keep in mind that a saved state may add layers and it should turn on at the correct time.

PR #511, opened 2016-01-20
Fix legend bugs

caseypt:
Fixes legend bugs reported in the demo. See commit messages for details.

**Testing instructions**
* Verify the behavior reported in each bug is fixed.

Connects to #507 
Connects to #508 
Connects to #509

### DIFF
--- a/src/GeositeFramework/css/app.css
+++ b/src/GeositeFramework/css/app.css
@@ -622,7 +622,6 @@ input.code-like, textarea.code-like {
         overflow: auto;
         height: 100%;
         width: 100%;
-        min-height: 200px;
     }
 
 .legend-layer {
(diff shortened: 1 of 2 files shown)


mmcfarland:
Sorry, this is getting more complicated, found two more edge cases:

* If you add la bunch of layers while the legend is minimized to the point where it will autoResize, when you maximize it the contents will extended outside of the legend body, presumably has to do with the 0 h/w of the body when minimized and doing calculations.
screenshot from 2016-01-21 10 32 53
* There's an issue when sometimes when the screen is relatively narrow (as in split screen), the legend can restore itself from minimized to an area under the plugin.  I kept thinking the legend was being removed, but it was just hidden underneath map layers.  If we could bump the z-index on the legend I think it'd fix this.

mmcfarland:
Fixes provided work great.

caseypt:
Pushed two fixes for the two bugs mentioned above.

caseypt:
Updated with an attempted fix for browsers with big fonts/scrollbars.

mmcfarland:
+1, all working nicely.

caseypt:
Thanks for the thorough review.

Issue #479, opened 2015-12-31
Use sample layer selector config to render interactive layer tree


caseypt:
Using a yet to be determined method (underscore templates?) render the layer selector tree and retrieve AGS service information when needed.

mmcfarland:
Keep in mind that the plugin will also be restoring from a saved state with certain layers already selected and with attributes set (opacity).

PR #512, opened 2016-01-21
Render layer selector from config

kdeloach:
This PR handles schema validation and (basic) rendering the tree to the UI. I will create a separate PR for styling and user interactions.

image

Connects #479

### DIFF
--- a/src/GeositeFramework/plugins/layer_selector_v2/config.js
+++ b/src/GeositeFramework/plugins/layer_selector_v2/config.js
@@ -18,9 +18,47 @@ define([
     ],
     function (declare, JSON, tv4, layerConfigSchema, layerSourcesJson, _) {
 
+        // Wraps raw layer nodes from the plugin configuration to provide
+        // convenience functions for template rendering.
+        // References to parent nodes are preserved to support property
+        // inheritance.
+        var LayerNode = declare(null, {
+            constructor: function(parent, node) {
+                this.parent = parent;
+                this.node = node;
+                this.children = [];
+            },
+            addChild: function(layerNode) {
+                this.children.push(layerNode);
+            },
+            getDisplayName: function() {
+                return this.node.displayName || this.node.name;
+            },
+            getChildren: function() {
+                return this.children;
+            },
+            hasChildren: function() {
+                return this.children.length > 0;
+            },
+            isLoading: function() {
+                return false;
+            }
+        });
+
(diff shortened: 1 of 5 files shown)


mmcfarland:
Looks great and simple.  My recollection was that we switched _from_ `name` to `layerName`/`serviceName`, not the other way around and the high level `layers` was different from the `include/excludeLayers`.  This looks like it works though, and makes sense.  I'll defer to @caseypt if he sees an issue with it.

src/GeositeFramework/plugins/layer_selector_v2/config.js
```diff
@@ -0,0 +1,93 @@
...
+            location: "//cdnjs.cloudflare.com/ajax/libs/underscore.js/1.8.3",
+            main: "underscore-min"
+        }
+    ]
+});
+
+define([
+        "dojo/_base/declare",
+        "dojo/json",
+        "use!tv4",
```
mmcfarland: We were looking at dropping `use` in favor of the dojo shim system that's now available.  Consider upgrading this.

kdeloach:
> Looks great and simple. My recollection was that we switched from name to layerName/serviceName, not the other way around and the high level layers was different from the include/excludeLayers. This looks like it works though, and makes sense. I'll defer to @caseypt if he sees an issue with it.

I suggested using layerName/serviceName but as I was implementing this I realized that the original names were better. The longer names are a little redundant and it was confusing how "server" and "service" is used interchangeably.

caseypt:
This looks great. Config looks good to me.

src/GeositeFramework/plugins/layer_selector_v2/config.js
kdeloach: I'd like to do this in a separate PR when I have a stronger grasp on what this entails.

caseypt: It looks like it's AMD compliant now: https://github.com/geraintluff/tv4#installation

mmcfarland:
Cool, +1

src/GeositeFramework/plugins/layer_selector_v2/config.js
kdeloach: I tried to load the latest release of tv4 as a package (similar to jquery or underscore) but it didn't work as I expected it to.

PR #517, opened 2016-01-22
Load and use ESRI units module.

### LINKED ISSUE #515: Measure Plugin throws on `esri.UNITS.esriMile`
Occasionally when loading the measure plugin, the `esri.UNITS.esriMile` property is undefined (to be specific, `esri.UNITS` is undefined).   This throws an exception which is now suppressed by the framework, but I didn't figure out why that field is sometimes unavailable.   I'm assuming it's related to the new JSAPI version and/or the bundle script.

https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/master/src/GeositeFramework/plugins/measure/AgsMeasure.js#L57

caseypt:
I can't confirm that this fixes #515, because it's so intermittent, but I haven't encountered it since I made this change (I refreshed about 20 times).

**Testing instructions**
- Confirm that the measure tool still works.
- Try and reproduce the gray basemap bug.

### DIFF
--- a/src/GeositeFramework/plugins/measure/AgsMeasure.js
+++ b/src/GeositeFramework/plugins/measure/AgsMeasure.js
@@ -5,8 +5,9 @@ define(["jquery",
         "esri/geometry/geodesicUtils",
         "esri/geometry/webMercatorUtils",
         "esri/tasks/GeometryService",
-	    "esri/geometry/ScreenPoint",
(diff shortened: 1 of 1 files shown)


kdeloach:
+1

mmcfarland:
+1, tested and seems like the right fix.  May want to get this in master, too.

PR #524, opened 2016-01-27
Add ajax utils

kdeloach:
This introduces a design pattern for fetching data from ajax requests
and consuming the result in a synchronous way (without callback
functions).

The implementation for the Legend widget has been generalized into ajax
utils that can be used for other components/plugins.

The purpose of this is to avoid deeply nested async callback trees (think
layer selector). Additionally, the cache is shared throughout the app so
plugins fetching the same data won't make extra ajax requests (Ex. if
you open the same plugin in Split View).

**To test:**
If this works as designed, the legend should work exactly as it did before.

### DIFF
--- a/src/GeositeFramework/js/Map.js
+++ b/src/GeositeFramework/js/Map.js
@@ -2,15 +2,13 @@
 /*global Backbone, _, $, Geosite, esri, Azavea, setTimeout, dojo, dojox */
 
 require(['use!Geosite',
-         'dojo/Deferred',
-         'dojo/request/xhr',
          'framework/Legend',
+         'framework/util/ajax',
          'esri/map'
         ],
     function(N,
-             Deferred,
-             xhr,
              Legend,
+             ajaxUtil,
              Map) {
     'use strict';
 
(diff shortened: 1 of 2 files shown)


kdeloach:
This is related to other changes I'm making to the layer selector.

src/GeositeFramework/js/util/ajax.js
```diff
@@ -0,0 +1,49 @@
...
+            return typeof cache[url] !== 'undefined';
+        }
+
+        function shouldFetch(url) {
+            return !isFetching(url) && !isCached(url);
+        }
+
+        return {
+            get: get,
+            fetch: fetch,
```
kdeloach: Open to suggestions for better naming.

src/GeositeFramework/js/Map.js
```diff
@@ -206,33 +204,18 @@ require(['use!Geosite',
...
-            };
-        }());
+        var redraw = function() {
+            legend.render(getVisibleLayers());
+        };
+
+        function getServiceLegend(service) {
+            var legendUrl = service.url + '/legend',
+                data = ajaxUtil.get(legendUrl);
+            if (ajaxUtil.shouldFetch(legendUrl)) {
```
caseypt: It's not totally clear to me here why `get` is called before the data is fetched. `data` will be populated by `fetch` so that when this function returns it's not undefined if the cache was empty initially?

caseypt: Does it make sense for `get` to call `fetch` internally if there isn't a cached value?

caseypt:
+1. Looks great. Legend still works.

screen shot 2016-01-28 at 10 13 12 am

src/GeositeFramework/js/Map.js
kdeloach: I don't think you can call `fetch` inside `get` because there must be a way to indicate which action to perform after the request completes.

The usefulness of having these 2 distinct operations is more apparent in the layer selector v2 where the usage is a little different. In that case, we only `fetch` when a layer is "activated" (lazy) but we `get` every time the tree is redrawn. Compared to the layer selector, the legend is eager.

Since `get` just returns a cached response, you *could* rewrite this to use `fetch` only, but then the code would be riddled with nested async callbacks and debugging and error handling would become much more complicated.

caseypt: Sounds good to me. That makes sense.

Issue #480, opened 2015-12-31
Add support for retrieving a single layer from AGS and adding it to the map


caseypt:
Selecting a basic layer (not a combined layer, WMS, or layer group) should add it to the map.

PR #526, opened 2016-01-28
Add support for adding AGS layers to the map

kdeloach:
### Changes

* Add support for AGS layers
* Implement layer filter & clear all
* Implement `excludeLayers` blacklist
* Populate on-demand loaded layers 

Connects #480


src/GeositeFramework/plugins/layer_selector_v2/util.js
```diff
@@ -0,0 +1,28 @@
+define(['underscore'], function(_) {
+    // Differs from the underscore `find` in that this version will return
```
kdeloach: TODO strict

src/GeositeFramework/plugins/layer_selector_v2/main.js
```diff
@@ -35,6 +41,71 @@ define([
...
+                if (mapLayer) {
+                    return;
+                }
+
+                // TODO: switch
+                mapLayer = new ArcGISDynamicMapServiceLayer(serviceUrl);
+
+                // Need to assign a deterministic ID, otherwise, the ESRI
+                // JSAPI will generate a unique ID for us.
+                mapLayer.id = serviceUrl;
```
caseypt: Does it not use the id defined in the service?

kdeloach: Layers have ids but map services don't, as far as I can tell.

caseypt:
This is working well so far. Nice job. I can take another look at it once it's out of WIP.

src/GeositeFramework/plugins/layer_selector_v2/main.js
caseypt: Makes sense.

kdeloach:
Rebased and updated the PR description.

mmcfarland:
I can't remove a layer by unselecting it (as opposed to unselecting its parent), is that part of the issue you noted in the PR description?

kdeloach:
I think so. I could update this to only allow leaf nodes to be added to the map to make it easier to test.

kdeloach:
Aside from the UI, all outstanding issues have been addressed in the latest commits.

caseypt:
Taking a look now.

caseypt:
When I filter for "land use" and click on the land use item in the layer selector, I get the following error:

screen shot 2016-02-09 at 12 36 51 pm

caseypt:
+1. This is really impressive. Nice job. There is a lot of code here, but it is all very readable. I've tried to focus on finding bugs, but I've only run into the one I reported above.

caseypt:
Full stack trace

image

kdeloach:
I was able to reproduce that error by doing this:

1. Click "Hurricane Sandy"
2. Apply filter ("land use")
3. Click the "Land Use" layer

kdeloach:
Fixed.

caseypt:
+1. Verified the fix locally.

kdeloach:
Thanks.

Issue #500, opened 2016-01-19
Export page broken


zferdana:
Seems when we created the in-plugin print function that calls the brower's print, it broke the Generate Map within Export page

mapossum:
The error I am getting makes it look like an error with the ArcGIS Javascript version like the error that was reported in the measure tool:

TypeError: esri.tasks.PrintParameters is not a function(…) "TypeError: esri.tasks.PrintParameters is not a function

PR #529, opened 2016-02-02
Fix export and update ESRI JS

mmcfarland:
* Converts failing `esri.` calls in export module to correct syntax to fix the broken export feature
* Remove all existing (and currently functioning) `esri.` calls
* Update the esri JS Bundle

Connects #500 

Test:
* Attempt to export the map
* Attempt to geocode an address
* Attempt to use the Launchpad to activate a subregion


caseypt:
+1. Looks good. Tested. Verified the highlighted app features still work, and that the bundle is capturing all of the required files (besides the few that for whatever reason won't get picked up by the bundle).

Issue #486, opened 2016-01-06
Partner popup `X` close button is hidden


zferdana:
X is white and is invisible when trying to close the partner link along the top bar

zferdana:
Probably best to have a colored top bar that matches the framework color so the X shows up clearly, like the launchpad.

PR #530, opened 2016-02-02
Add stylistic header to popup

mmcfarland:
IFrame based Framework popups, like Partners, were obscuring the close
button and didn't have the header style of other popup types.

Connects #486 

Test:
* Click "GIS" on the test site and see if there is a blue border on the popup header.

popups


caseypt:
Non-iframe based pop-ups now have two headers:

image


image

mmcfarland:
Good catch.  Updated to remove header on close.  Switching between "GIS" and "Get Started" should work appropriately now.

caseypt:
+1. Tested. Looks good.

Issue #502, opened 2016-01-20
Jumping legend


zferdana:
in NYadaptation-region, Habitat Explorer plugin we get a jumping legend while in split view. This is happen when you choose forests as the habitat type.

zferdana:
Jumping legend in Future Habitat also appears in Virginia-region when interacting with Flood plugin

mapossum:
The jumping legend issue appears to be in many of the plugins such as Future Habitat when the plugin creates it's own legend.  Basically when one of these plugins are open and the map layers plugin is open you see it refreshing constantly.  Virginia is a good test case.  Open Future Habitat, then open map layers and turn on a few layers to see the issue.

PR #531, opened 2016-02-03
Custom legend jumps

mmcfarland:
This is a stopgap measure to prevent legend flickering when using the
Habitat Explorer.  The comparison of layers on the map vs legends in the
Legend is expanded to account for possible custom legend.  The need to do
the comparison is a result of the existing unreliability of the JSAPI to
notify of `layerAdd` events.

Connects #502 

#### To test:

* Check that the legend does not flicker in the development build.
* Check that the future habitat plugin in split screen does not flicker (attached)
* Check that after turning on and turning off the Community Rating System plugin, there is no flicker.

#### Test Plugins

(Make sure these extract like `plugins/future_habitat/main.js`, the zip may double up the plugin directory path)
* [future_habitat.zip](https://github.com/CoastalResilienceNetwork/GeositeFramework/files/116013/future_habitat.zip)
* [community-rating-system.zip](https://github.com/CoastalResilienceNetwork/GeositeFramework/files/116309/community-rating-system.zip)

### DIFF
--- a/src/GeositeFramework/js/Map.js
+++ b/src/GeositeFramework/js/Map.js
@@ -322,7 +322,7 @@ require(['use!Geosite',
                 expectedMissingLayerCount = 1 + customLegendsActive + serviceOptOut;
 
             if ((totalVisibleLayers.length - expectedMissingLayerCount) !== nuggetCount) {
-                _.delay(updateLegend, 250);
+                _.delay(updateLegend, 750);
             }
         }
     }
--- a/src/GeositeFramework/js/Map.js
+++ b/src/GeositeFramework/js/Map.js
@@ -307,8 +307,13 @@ require(['use!Geosite',
                 return !layer.graphics && layer.visible;
             });
 
-            // Subtract one for the base layer.
-            if ((totalVisibleLayers.length - 1) !== nuggetCount) {
+            // Subtract one for the base layer and one for each 
+            // custom legend that is visible and has children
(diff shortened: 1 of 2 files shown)


src/GeositeFramework/js/Map.js
```diff
@@ -307,8 +315,13 @@ require(['use!Geosite',
                 return !layer.graphics && layer.visible;
             });
 
-            // Subtract one for the base layer.
-            if ((totalVisibleLayers.length - 1) !== nuggetCount) {
+            // Subtract one for the base layer and one for each 
+            // custom legend that is visible and has children
```
mmcfarland: I'll rebase a change to this comment that includes the description of the `serviceOptOut` after review.

caseypt:
+1. Tested. There is some initial flickering when layers are first added, but in all cases I tested, things settle down in a second or two.

mmcfarland:
Yeah, I noticed that too, do you think it's worth increasing the redraw delay time?

caseypt:
Yeah, that might be a simple way to get around it.

mmcfarland:
Increased from `250 -> 750` in https://github.com/mmcfarland/GeositeFramework/commit/009cf862b504861748e4e5f484f0e1c31f40256c

I don't notice flickering any more.

caseypt:
Just did a couple quick tests and flickering has been significantly reduced.

mmcfarland:
Cool, thanks for the review.  Merging.

Issue #523, opened 2016-01-25
Add description popup to layer


mmcfarland:
The layer options should have an icon link to display a small tool-tip with `description` data, set either from the service or the configuration.

PR #549, opened 2016-02-11
Add description popup to layer

kdeloach:
image

Connects #523

### DIFF
--- a/src/GeositeFramework/plugins/layer_selector_v2/LayerNode.js
+++ b/src/GeositeFramework/plugins/layer_selector_v2/LayerNode.js
@@ -35,7 +35,7 @@ define([
                     return !_.contains(blacklist, layer.getName());
                 }
                 return this.includeAllLayers() ||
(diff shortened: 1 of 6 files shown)


kdeloach:
I mostly copied what already existed for "zoom to extent".

caseypt:
Works well. Two things:

1) Like opacity, they can override the description from the config (there a bug that will prevent us from doing this. I'll open an issue).
2) This is probably outside the scope of the PR, but I think we should add some sort of "loading" indicator. For example, when zooming to an extent or getting the description for a layer that has been shown, there is a slight delay.

caseypt:
Here's the issue I ran into when adding support for opacity: https://github.com/CoastalResilienceNetwork/GeositeFramework/issues/550

caseypt:
+1

kdeloach:
Fixed all bugs. I'll create an issue to add some kind of loading indicator.

caseypt:
+1. Tested the changes.

PR #552, opened 2016-02-11
Rebuild layers after fetching data

kdeloach:
Fixes a small bug where `rebuildLayers` is called  *before* the map service
has been fetched instead of *after* it has been fetched. Some layers
need to be activated twice to add them to the map without this fix.


kdeloach:
You can reproduce this bug by selecting the Hurricane Sandy" layer. Before, you would need to select it twice to add it to the map. Now, it should be added immediately after the map service is fetched.

kdeloach:
Going to merge this in ahead of time for RRP.

Issue #555, opened 2016-02-11
Disable laziness


mmcfarland:
To allow the layer filter to work correctly, consider loading all map services when the plugin is activated.

Issue #522, opened 2016-01-25
Add Layer opacity control


mmcfarland:
The layer options should contain an opacity slider.  The value of o the slider should be included in the state object returned from `getState`

Issue #545, opened 2016-02-10
Layer selector style issues


kdeloach:
Outstanding design issues:
* Need hover & active states for layer tools
* Clicking the visibility icon should toggle the layer on the map
* Layers should appear selected (hover state?) when a layer tool is activated but the layer itself is not selected

PR #556, opened 2016-02-12
Add support for adjusting layer opacity.

caseypt:
Allows users to set the opacity on a layer by layer basis for those
layers that are part of services that support dynamic layers.

* Set individual layer opacity using the LayerDrawingOptions class. This
feature is only supported for layers that are served from a service with
support for dynamic layers turned on. This is not the default setting.
* Save layer opacity settings are part of the state.
* Adjust the layer menu to show or hide the opacity slider depending on
the layer's support for opacity.
* Add opacity to the schema.
* Add NOAA layers to the config to demo the opacity slider.

**Testing instructions**
- Activate one of the NOAA layers (the server is pretty slow, be patient). Confirm that you can adjust the opacity.
- Turn on the other NOAA layer and confirm that it's opacity is not the same as the other layer.
- The opacity slider position should be preserved between opening and closing the menu.
- Select one of the other layers from the New Jersey service and confirm that the layer slider is not present (this may change if the dynamic service option is turned on).
- UPDATE: The first NOAA layer should now start at 80% opacity.

image

Known issues:
- [X] Opacity settings in the config are ignored (working on this).
- [ ] Lazy loading prevents the slider from being present before the layer is loaded even for layers that support opacity. Should be fixed by #555.

Connects to #522

### DIFF
--- a/src/GeositeFramework/plugins/layer_selector_v2/State.js
+++ b/src/GeositeFramework/plugins/layer_selector_v2/State.js
@@ -21,7 +21,8 @@ define([
         var LAYERS_CHANGED = 'change:layers',
             SELECTED_LAYERS_CHANGED = 'change:selectedLayers',
             FILTER_CHANGED = 'change:filter',
-            EVERYTHING_CHANGED = 'change:all';
+            EVERYTHING_CHANGED = 'change:all',
+            OPACITY_CHANGED = 'change:opacity';
 
         // Fetch layer data and return promise.
         function fetch(layer) {
@@ -88,7 +89,9 @@ define([
                     // Selected layerIds (in-order).
                     selectedLayers: [],
                     // Expanded layerIds.
-                    expandedLayers: []
+                    expandedLayers: [],
+                    // List of objects as { layerId: opacityValue }.
+                    layerOpacity: []
                 });
                 this.config = config;
                 this.rebuildLayers();
@@ -209,6 +212,17 @@ define([
 
             getSelectedLayers: function() {
                 return _.map(this.savedState.selectedLayers, this.findLayer, this);
+            },
+
+            getSelectedLayersForService: function(serviceUrl) {
+                var self = this;
+                return _.reduce(this.savedState.selectedLayers, function(memo, layer) {
+                    var layerNode = self.findLayer(layer);
+                    if (layerNode && layerNode.getServiceUrl() === serviceUrl) {
+                        memo.push(layer);
+                    }
+                    return memo;
+                }, []);
             },
 
             clearAll: function() {
@@ -344,6 +358,49 @@ define([
 
             serialize: function() {
                 return this.savedState;
+            },
+
+            getLayerOpacity: function(layerId) {
+                var savedLayerOpacity = _.findWhere(this.savedState.layerOpacity, { layerId: layerId });
+
+                if (savedLayerOpacity) {
+                    return savedLayerOpacity.opacity;
+                } else {
+                    var layer = this.findLayer(layerId),
+                        configOpacity = layer.getOpacity();
+
+                    if (configOpacity) {
+                        return configOpacity;
+                    }
+                }
+
+                return 1;
+            },
+
+            setLayerOpacity: function(layerId, opacity) {
+                var layerItem = _.findWhere(this.savedState.layerOpacity, { layerId: layerId });
+
+                if (layerItem) {
+                    layerItem.opacity = opacity;
+                } else {
+                    this.savedState.layerOpacity.push({
+                        layerId: layerId,
+                        opacity: opacity
+                    });
+                }
+
+                this.emit(OPACITY_CHANGED);
+            },
+
+            serviceSupportsOpacity: function(layerId) {
+                var layer = this.findLayer(layerId),
+                    serviceData = getServiceData(layer);
+
+                if (serviceData && serviceData.supportsDynamicLayers) {
+                    return true;
+                }
+
+                return false;
             }
         });
     }
--- a/src/GeositeFramework/plugins/layer_selector_v2/main.js
+++ b/src/GeositeFramework/plugins/layer_selector_v2/main.js
@@ -21,6 +21,7 @@ define([
         "esri/layers/FeatureLayer",
         "esri/layers/ArcGISDynamicMapServiceLayer",
         "esri/layers/ArcGISTiledMapServiceLayer",
+        "esri/layers/LayerDrawingOptions",
         "framework/PluginBase",
         "./state",
         "./config"
@@ -32,6 +33,7 @@ define([
              FeatureLayer,
              ArcGISDynamicMapServiceLayer,
              ArcGISTiledMapServiceLayer,
+             LayerDrawingOptions,
              PluginBase,
              State,
              Config) {
@@ -70,16 +72,7 @@ define([
                         self.hideLayerInfo();
                     })
                     .on('click', 'a.more', function() {
-                        var $el = $(this),
-                            layerId = self.getClosestLayerId(this),
-                            $menu = self.createLayerMenu(layerId),
-                            $shadow = self.createLayerMenuShadow(),
-                            pos = $el.offset();
-                        $menu.css({
-                            top: pos.top,
-                            left: pos.left
-                        });
-                        $('body').append($shadow).append($menu);
+                        self.showLayerMenu(this);
                     })
                     .on('keyup', 'input.filter', function() {
                         var $el = $(this),
@@ -99,6 +92,11 @@ define([
                     .on('click', '#' + this.layerMenuId + ' a.zoom', function() {
                         self.zoomToLayerExtent(self.getClosestLayerId(this));
                         self.destroyLayerMenu();
+                    })
+                    .on('change', '#' + this.layerMenuId + ' .slider', function() {
+                        var layerId = self.getClosestLayerId(this),
+                            opacity = parseFloat($(this).find('input').val());
+                        self.setLayerOpacity(layerId, opacity);
                     });
             },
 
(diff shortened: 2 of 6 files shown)


caseypt:
I also ran into this issue: https://github.com/CoastalResilienceNetwork/GeositeFramework/issues/557

src/GeositeFramework/plugins/layer_selector_v2/State.js
```diff
@@ -344,6 +375,34 @@ define([
 
             serialize: function() {
                 return this.savedState;
+            },
+
+            setLayerOpacity: function(layerId, opacity) {
+                var layerItem = _.find(this.savedState.layerOpacity, function(layer) {
+                    return layer.layerId === layerId;
+                });
```
kdeloach: Consider replacing with `_.findWhere` to match how it's done in `getLayerOpacity`.

src/GeositeFramework/plugins/layer_selector_v2/main.js
```diff
@@ -98,6 +91,11 @@ define([
                     .on('click', '.layer-selector2-layer-menu a.zoom', function() {
                         self.zoomToLayerExtent(self.getClosestLayerId(this));
                         self.destroyLayerMenu();
+                    })
+                    .on('change', '.layer-selector2-layer-menu .slider', function() {
```
kdeloach: Will need to be rebased to include changes from #558.

src/GeositeFramework/plugins/layer_selector_v2/main.js
```diff
@@ -109,15 +107,36 @@ define([
...
+                    left: pos.left
+                });
+
+                $('body').append($shadow).append($menu);
+            },
+
+            _createLayerMenu: function(layerId) {
                 var layer = this.state.findLayer(layerId),
+                    supportsOpacity = this.state.serviceSupportsOpacity(layerId) || false,
+                    opacity = this.state.getLayerOpacity(layerId) || 1,
```
kdeloach: Consider making `getLayerOpacity` return 1 by default.

src/GeositeFramework/plugins/layer_selector_v2/main.js
```diff
@@ -109,15 +107,36 @@ define([
...
+                    top: top,
+                    left: pos.left
+                });
+
+                $('body').append($shadow).append($menu);
+            },
+
+            _createLayerMenu: function(layerId) {
                 var layer = this.state.findLayer(layerId),
+                    supportsOpacity = this.state.serviceSupportsOpacity(layerId) || false,
```
kdeloach: `serviceSupportsOpacity` returns false by default.

src/GeositeFramework/plugins/layer_selector_v2/main.js
```diff
@@ -160,14 +179,47 @@ define([
...
+
+            _setOpacityForService: function(visibleLayerIds) {
+                // VisibleLayerIds is a map of serviceUrls -> selected layer service Ids
+                _.each(visibleLayerIds, function(layerServiceIds, serviceUrl) {
+                    var mapLayer = this.map.getLayer(serviceUrl),
+                        self = this;
+
+                    if (layerServiceIds.length !== 0) {
+                        var optionsArray = [],
+                            selectedLayers = this.state.getSelectedLayersForService(serviceUrl);
```
kdeloach: This seems a little redundant. `layerServiceIds` should already contain all the selected layers for this service. Right?

src/GeositeFramework/plugins/layer_selector_v2/main.js
```diff
@@ -160,14 +179,47 @@ define([
...
+                                        // to 1 (opaque) so we convert it and invert it here.
+                                        transparency: 100 - (layerOpacity * 100)
+                                    });
+
+                                    var layer = self.state.findLayer(layerId);
+                                    optionsArray[layer.getServiceId()] = drawingOption;
+                                }
+                            });
+
+                            mapLayer.setLayerDrawingOptions(optionsArray);
```
kdeloach: Do we need to set drawing options to empty in the case that all layers for this service are deselected (similar to how visible layers is set to `[-1]` above)?

src/GeositeFramework/plugins/layer_selector_v2/main.js
```diff
@@ -160,14 +179,47 @@ define([
...
+                        self = this;
+
+                    if (layerServiceIds.length !== 0) {
+                        var optionsArray = [],
+                            selectedLayers = this.state.getSelectedLayersForService(serviceUrl);
+
+                            _.each(selectedLayers, function(layerId) {
+                                var layerOpacity = self.state.getLayerOpacity(layerId);
+
+                                if (layerOpacity) {
```
kdeloach: This condition evaluates to false when opacity is 0.

image

src/GeositeFramework/plugins/layer_selector_v2/main.js
```diff
@@ -109,15 +107,36 @@ define([
...
 
-            createLayerMenu: function(layerId) {
+            showLayerMenu: function(el) {
+                var $el = $(el),
+                    layerId = this.getClosestLayerId(el),
+                    $menu = this._createLayerMenu(layerId),
+                    $shadow = this._createLayerMenuShadow(),
+                    supportsOpacity = this.state.serviceSupportsOpacity(layerId),
+                    pos = $el.offset(),
+                    top = supportsOpacity ? pos.top : pos.top + 50;
```
kdeloach: The positioning is a little off for layers that do not support opacity.

image

src/GeositeFramework/plugins/layer_selector_v2/main.js
caseypt: Right, but `selectedLayers` contains the string ids that we need to look up the opacity for each layer.

src/GeositeFramework/plugins/layer_selector_v2/State.js
caseypt: Thanks.

src/GeositeFramework/plugins/layer_selector_v2/main.js
caseypt: Good call.

src/GeositeFramework/plugins/layer_selector_v2/main.js
caseypt: Thanks.

src/GeositeFramework/plugins/layer_selector_v2/main.js
caseypt: Not sure. It doesn't seem useful to set it if the layers aren't visible, so I'll skip this step if that's the case.

src/GeositeFramework/plugins/layer_selector_v2/main.js
caseypt: Thanks. Condition removed (that function now returns 1 by default).

src/GeositeFramework/plugins/layer_selector_v2/main.js
caseypt: Position adjusted.

src/GeositeFramework/plugins/layer_selector_v2/main.js
caseypt: Or rather I'll set it to an empty array.

caseypt:
Rebased and updated.

I also swapped out the NOAA service with the Alabama service. It's way faster, and change opacity is now almost instant.

src/GeositeFramework/plugins/layer_selector_v2/main.js
kdeloach: Oh, I see. Instead of iterating over `visibleLayerIds` you might just want to create another loop based on this code:  https://github.com/caseypt/GeositeFramework/blob/feature/opacity-slider-2/src/GeositeFramework/plugins/layer_selector_v2/main.js#L163.

Another alternative might be something like this?

```
layerByService = groupBy(getSelectedLayers(), layer => layer.getServiceUrl())
getDrawingOpts = layers => reduce(...)
each(layerByService, (layers, serviceUrl) => {
    drawingOpts = getDrawingOpts(layers)
    mapLayer = map.getLayer(serviceUrl)
    mapLayer.setLayerDrawingOptions(drawingOpts)
})

```

kdeloach: Rationale is to reduce nesting and to make the transformation from selected layers to drawing options as simple as possible.

src/GeositeFramework/plugins/layer_selector_v2/main.js
```diff
@@ -110,16 +108,37 @@ define([
...
 
-            createLayerMenu: function(layerId) {
+            showLayerMenu: function(el) {
+                var $el = $(el),
+                    layerId = this.getClosestLayerId(el),
+                    $menu = this._createLayerMenu(layerId),
+                    $shadow = this._createLayerMenuShadow(),
+                    supportsOpacity = this.state.serviceSupportsOpacity(layerId),
+                    pos = $el.offset(),
+                    top = supportsOpacity ? pos.top : pos.top + 55;
```
kdeloach: Not sure why, but the layer tools dissappear when opening the layer menu if the layer is not selected. We could fix this as part of #545.

image

kdeloach:
+1

src/GeositeFramework/plugins/layer_selector_v2/main.js
caseypt: Thanks for the suggestion. Updated with the alternative.

src/GeositeFramework/plugins/layer_selector_v2/State.js
```diff
@@ -211,6 +214,17 @@ define([
...
+            getSelectedLayersForService: function(serviceUrl) {
+                var self = this;
+                return _.reduce(this.savedState.selectedLayers, function(memo, layer) {
+                    var layerNode = self.findLayer(layer);
+                    if (layerNode && layerNode.getServiceUrl() === serviceUrl) {
+                        memo.push(layer);
+                    }
+                    return memo;
+                }, []);
+            },
```
kdeloach: This isn't referenced anywhere now.

caseypt: Thanks. I'll trash it in a future PR.

PR #558, opened 2016-02-15
Prevent events from being handled on both maps.

caseypt:
* Previously, two events were fired from listeners on the
layer menu. I believe this was because each layer selector
instance has it's own set of templates, and because of the
way the listener is set up, both instances of the layer
menu were being captured. To fix this, we give each
instance of the layer menu a unique id that we can bind
events to. We could have also created the layer selector
DOM element as a child of the layer selector, but this made
positioning the element difficult.

**Testing instructions**
- Activate split screen mode and activate the new layer selector on both maps.
- On one of the maps, zoom to the extent of a layer, and verify that only that map is zoomed. Do the same on the other map.

Connects to #553

### DIFF
--- a/src/GeositeFramework/plugins/layer_selector_v2/main.js
+++ b/src/GeositeFramework/plugins/layer_selector_v2/main.js
@@ -53,6 +53,7 @@ define([
                 this.layerTmpl = _.template(this.getTemplateByName('layer'));
                 this.infoBoxTmpl = _.template(this.getTemplateByName('info-box'));
                 this.layerMenuTmpl = _.template(this.getTemplateByName('layer-menu'));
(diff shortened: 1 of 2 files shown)


kdeloach:
Looks like I caused this by creating a global instance of `.layer-selector2-layer-menu`. A future-proof solution might be to use a unique ID class name per instance (ex. `_.uniqueId('layer-selector2-layer-menu-')`) or append the layer menu to the plugin container instead of the body.

caseypt:
Thanks. Updated with the unique id method. I initially tried adding the element as a child of the layer selector container, but it made positioning it difficult.

kdeloach:
+1 Tested

PR #561, opened 2016-02-16
Add unit tests

kdeloach:
This adds a few simple unit tests to describe the interaction between
LayerNode objects.

Test by loading the application normally. There should be a debug statement in the developer console that says "All tests passed!".

Connects #544

### DIFF
--- a/src/GeositeFramework/js/unittest.js
+++ b/src/GeositeFramework/js/unittest.js
@@ -1,0 +1,54 @@
+define([
+        "dojo/Deferred",
+        "dojo/promise/all",
+        "underscore"
+    ],
+    function(Deferred,
+             all,
+             _) {
+        "use strict";
+
+        function runAllTests(tests) {
+            var promise = all(_.map(tests, function(test) {
+                var defer = new Deferred(),
+                    done = function() {
+                        defer.resolve();
+                    };
+
+                // Fail the test if it takes longer than 1000 ms to execute.
+                setTimeout(function() {
+                    if (!defer.isFulfilled()) {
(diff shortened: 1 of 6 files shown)


kdeloach:
Planning to add more comprehensive tests after I finish refactoring `State`.

caseypt:
Cool idea to run the tests by requiring them. However, if during development the tests break because you are adding new functionality or refactoring, say, then an exception from a broken test will prevent the application from loading.

I don't have a better solution at the moment, but ideally you could break a test and keep developing.

caseypt:
Maybe we can use this: https://dojotoolkit.org/reference-guide/1.9/util/doh.html?

kdeloach:
Yeah, that's true. I kind of like that breaking unit tests has serious consequences. On the other hand, I imagine that I will most likely comment out unit tests if it becomes too much of a hindrance.

kdeloach:
If we can get DOH working that would be great. I'm skeptical it will be easy though, considering all the setup and shimming that we have to do to load the application and plugins. The unit test environment may end up being very different from the actual environment. I'll try it out later.

caseypt:
That's true. Forgot about the shims.

caseypt:
If it's too much trouble to try and get DOH to work, we can go with this and just comment out the import during development if tests are failing. In it's current state, this is really nice to have.

kdeloach:
Rebased and disabled unit tests by default. We'll have to uncomment the require statements to run them in development.

caseypt:
+1. Verified that all the tests pass. Nice work finding a way to get tests in here.

Issue #527, opened 2016-02-01
Hide/show subregions toggle button


zferdana:
This button on the map should be removed when we don't have subregions as part of the launchpad

mmcfarland:
This has been clarified that removing the subregion definition from region.json will presently remove the toggle button.

PR #562, opened 2016-02-16
Subregion schema flexibility

mmcfarland:
Some tidying up to ensure that subregions can be easily removed without
exceptions.  The visibility of the button was already tied to the validate
function, checking that the region data had subregions enabled.

Connects #527 

#### Test
* Remove the subregion node from region.js
* See that there are no runtime errors from the launchpad reference
* See that the hide/show toggle is not visible


caseypt:
Just making sure this should be targeted at `master`.

PR #565, opened 2016-02-16
Add support for subregions

kdeloach:
This adds a new layer filter for subregions.

Test by verifying:
* "Coastal Resilience > Boundaries > Watershed Regions" *only* appears in "Galveston Bay"
* "Coastal Resilience > Coastal Hazards > Sea Level Rise Impact Analysis" appears in both **main** and "Somewhere in the ocean"

Connects #547

### DIFF
--- a/src/GeositeFramework/plugins/layer_selector_v2/LayerNode.js
+++ b/src/GeositeFramework/plugins/layer_selector_v2/LayerNode.js
@@ -145,6 +145,10 @@ define([
                 return this.node.combine;
             },
 
+            isAvailableInRegion: function(regionId) {
+                return _.contains(this.node.availableInRegions || [], regionId);
+            },
+
             getOpacity: function() {
                 return this.node.opacity;
             }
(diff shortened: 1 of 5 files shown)


caseypt:
Consider replacing use of `region` with `subregion` throughout the PR to be more clear.

caseypt:
I just noticed that in the old config the settings is called `availableInRegion` (as it is here) and I think I understand why. We may need to replicate this logic: https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/e703afe55e1d62a17296f212d421ae133974bf09/src/GeositeFramework/plugins/layer_selector/LayerManager.js#L43-L50

caseypt:
Code looks good, I'm just not sure if the filtering behavior is correct.

kdeloach:
Good catch. Fixed the subregion filtering and updated the test instructions.

caseypt:
+1. Tested. It's nice to see that we were able to implement this functionality in a much more low impact way than we could with the last layer selector.

Issue #541, opened 2016-02-10
Geocoder plugin closes when Launchpad opens


mmcfarland:
The geocoder/address plugin should start with its UI expanded/opened and usually does.  It appears that regions (gulfmex production) that use the Launchpad see it collapse when the Launchpad opens.

PR #566, opened 2016-02-16
Fix search plugin auto-closing

caseypt:
Remove the hibernate implementation for the search plugin so that
it cannot be closed programmatically. TNC wants it to persist it's
state through subregion and scenario activations/deactivations.

**Testing instructions**
- Verify the plugin remains open when on start-up, when activating scenarios and subregions.
- Verify the "X" button can still clear results and close the plugin.

Connects to #541


mmcfarland:
+1

PR #567, opened 2016-02-16
Only apply drawing options if service support transparency

kdeloach:
Fixes a bug where setting a transparency value of 0 on services that don't
support transparency results in layers not appearing at all.


caseypt:
+1. Tested. Nice clean up and fix.

kdeloach:
Added more details to commit description.

Issue #554, opened 2016-02-11
Add support for loading from saved state


mmcfarland:
Verify that loading from saved state works as designed.

PR #570, opened 2016-02-17
Fix state not being restored properly

kdeloach:
Changes:

* Ensure `updateMap` is called after setting state
* Fetch map service data when restoring state

Test instructions:

* Verify that restoring saved state from "Save & Share" URL restores
selected layers, expanded folders, and the tree filter

Connects #554

### DIFF
--- a/src/GeositeFramework/plugins/layer_selector_v2/State.js
+++ b/src/GeositeFramework/plugins/layer_selector_v2/State.js
@@ -24,15 +24,6 @@ define([
             EVERYTHING_CHANGED = 'change:all',
             OPACITY_CHANGED = 'change:opacity';
 
-        // Fetch layer data and return promise.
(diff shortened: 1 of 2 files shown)


caseypt:
I think this is unrelated, but the filter doesn't work correctly for the "Planes, Trains, and Automobile" layer.

image

image

The layer disappears after typing "Pl".

kdeloach:
Fixed in #565 by https://github.com/kdeloach/GeositeFramework/commit/4a20648c22a0f9ad289f88203ea9bc0ab1bde937.

caseypt:
+1. Verified that layer selection was preserved and layers were activated when restoring state. Also verified that opacity was preserved and restored as well.

kdeloach:
Rebased.

PR #571, opened 2016-02-18
Implement style for layer tools

caseypt:
~~I still want to do a little more testing and shoring up on this.~~ Ready for testing.

Because the layer info box is independent of the layer selector window, and it persists through changes in layer selections, and because the info icon needs to stay highlighted while the info box is open even if another layer is selected, it was simplest to track the box as part of the state (as opposed to after render DOM manipulation) despite the added complexity of tracking such a simple thing.

The layer tools menu is a bit simpler since it is attached to the layer selector items and can be closed when a different layer is selected.

**Testing instructions**
- Open the layer menu for a layer and verify that the layer stays in the active/hover state until the layer menu is closed.
- Open the info box for a layer, and verify that while the info box is open, the layer element is displayed in the active/hover state, and that the "i" icon stays blue.
- Verify that clicking the eye icon toggles the visibility of a layer.

image

image

image

### DIFF
--- a/src/GeositeFramework/plugins/layer_selector_v2/State.js
+++ b/src/GeositeFramework/plugins/layer_selector_v2/State.js
@@ -22,7 +22,8 @@ define([
             SELECTED_LAYERS_CHANGED = 'change:selectedLayers',
             FILTER_CHANGED = 'change:filter',
             EVERYTHING_CHANGED = 'change:all',
-            OPACITY_CHANGED = 'change:opacity';
+            OPACITY_CHANGED = 'change:opacity',
+            LAYER_INFO_ID_CHANGED = 'change:layerInfoId';
 
         // Return true if layer data has been fetched.
         function isLoaded(layer) {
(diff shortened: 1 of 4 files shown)


caseypt:
Connects to #545

kdeloach:
Looks really nice. One minor thing I noticed is that scroll position isn't preserved when opening the layer info. You can reproduce this by expanding a bunch of folders and opening the info panel for the layer on the bottom.

kdeloach:
+1

caseypt:
> One minor thing I noticed is that scroll position isn't preserved when opening the layer info.

Good catch. I'll see if there is a simple way to fix this.

caseypt:
Scroll position is not being preserved because the layer selector is being rerendered when the layer info box opens. I think the only way solid way to solve this is to make the scroll position part of the state. I'll create an issue because that should probably happen after the refactor is merged in.

PR #573, opened 2016-02-18
Fix legend order

caseypt:
* By default, JavaScript sorts numbers as strings, so we need to pass a comparator to correctly sort numbers in ascending order.

**Testing instructions**
- Verify that layers are added to the legend in the order by their id (which should match the z-index of the layers on the map), and that the services are ordered according to the order that their layers were added to the map. This means that if you add a layer from the New Jersey service, then the Alabama service, the New Jersey layer will be on top in the legend. If all of the layers are removed, the JS API still remembers the initial order services were added. So, if these same layers are turned back on in the opposite order, the new legend order will match the initial legend order.

Connects to #495


kdeloach:
+1 I don't really understand why this works, but it does!

PR #575, opened 2016-02-22
Create Tree and AgsService modules

kdeloach:
Refactors the State module by splitting it up into 3 distinct modules:

* State - Represents plugin saved state; contains simple getter/setters

* Tree - Collection of layers; has filter/find functions as well as all
the data necessary to render the tree in the UI

* AgsService - Responsible for all ajax requests

The purpose of this change is to create several smaller units of code that
can be easily tested.

Other changes:

* LayerNode should now be the single source of truth for all layer data.
As a consequence, the tree must be recreated every time there is a state
change or map service data is fetched.

* Events have been removed. These didn't seem necessary anymore and
should make unit testing simpler.

### DIFF
--- a/src/GeositeFramework/plugins/layer_selector_v2/NullService.js
+++ b/src/GeositeFramework/plugins/layer_selector_v2/NullService.js
@@ -1,0 +1,46 @@
+define([
+        "dojo/_base/declare",
+    ],
+    function(declare) {
+        "use strict";
+
+        return declare(null, {
+            getServiceUrl: function() {
+                return null;
+            },
+
+            // Return a promise containing map service data.
+            fetchMapService: function() {
+                throw new Error('Not implemented');
+            },
+
+            // Return a promise with service layer data.
+            fetchLayerDetails: function(tree, layerId) {
+                throw new Error('Not implemented');
+            },
+
+            // Return cached map service data.
+            getServiceData: function() {
+                return null;
+            },
+
+            // Return cached layer details.
+            getLayerDetails: function(serviceLayer) {
+                return null;
+            },
+
+            // Find the corresponding data for `layer` in the map service.
+            findServiceLayer: function(layer) {
+                return null;
+            },
+
+            findServiceLayerById: function(layerId) {
+                return null;
+            },
+
+            supportsOpacity: function() {
+                return false;
+            }
+        });
+    }
+);
--- a/src/GeositeFramework/plugins/layer_selector_v2/State.js
+++ b/src/GeositeFramework/plugins/layer_selector_v2/State.js
@@ -2,21 +2,13 @@ define([
         "dojo/_base/declare",
         "underscore"
     ],
-    function(declare,
-             _) {
+    function(declare, _) {
         "use strict";
 
-        return declare(null, {
+        var State = declare(null, {
             constructor: function(data) {
-                this.setState(data);
-            },
-
-            getState: function() {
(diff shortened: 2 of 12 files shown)


kdeloach:
Test by making sure there have been no major regressions (everything should more or less work as it used to). Unit tests will follow in an upcoming PR.

src/GeositeFramework/plugins/layer_selector_v2/NullService.js
```diff
@@ -0,0 +1,46 @@
...
+                throw new Error('Not implemented');
+            },
+
+            // Return cached map service data.
+            getServiceData: function() {
+                return null;
+            },
+
+            // Return cached layer details.
+            getLayerDetails: function(layer, serviceLayer) {
```
caseypt: Based on `AgsService`, the signature should be `function(serviceLayer)`.

src/GeositeFramework/plugins/layer_selector_v2/NullService.js
```diff
@@ -0,0 +1,46 @@
...
+                return null;
+            },
+
+            // Find the corresponding data for `layer` in the map service.
+            findServiceLayer: function(serviceData, layer) {
```
caseypt: Based on AgsService, this one should be function(layer).

src/GeositeFramework/plugins/layer_selector_v2/NullService.js
```diff
@@ -0,0 +1,46 @@
...
+                return null;
+            },
+
+            findServiceLayerById: function(serviceData, layerId) {
```
caseypt: Only `function(layerId)` here.

src/GeositeFramework/plugins/layer_selector_v2/NullService.js
kdeloach: Good catch, I updated a few other signature mismatches.

src/GeositeFramework/plugins/layer_selector_v2/State.js
```diff
@@ -90,166 +19,47 @@ define([
...
-
-            clearAll: function() {
-                this.filterTree('');
-                this.setSelectedLayers([]);
-                this.setExpandedLayers([]);
-                this.emit(EVERYTHING_CHANGED);
+                return this.savedState.selectedLayers;
             },
 
             isSelected: function(layerId) {
```
caseypt: We talked about this in the in-person review, but it's helpful to have this method on both `State` and `LayerNode` as long as we can ensure they remain in sync.

caseypt:
+1 based on the in-person review and working off of this branch all day (for WMS layer support). I have not noticed any regressions. Nice work.

caseypt:
Also, once WMS support is in and the feature is stable, it might make sense for us to use inheritance for the service types. I've been able to keep the interfaces the same for now, and there are some methods that are the same for all services.

kdeloach:
That's a good idea.

src/GeositeFramework/plugins/layer_selector_v2/State.js
kdeloach: Yeah. Unfortunately, it's very easy to (potentially) get out of sync by forgetting to update references. I'll try to think about design patterns we can use to prevent this from happening.

kdeloach:
Merging!

Issue #519, opened 2016-01-25
Allow WMS layers


mmcfarland:
The layer selector should allow for WMS layers to be added to the configuration and displayed on the map.  If a legend service exists on the WMS server, include it in the legend.

caseypt:
Going to hold off on this until a larger refactor of the state module is complete.

PR #577, opened 2016-02-23
WMS service support

caseypt:
Adds support for WMS layers. See commit messages for details.

TODO

- [X] Map XML to JSON
- [X] Add layers to legend
- [ ] ~~Zoom to extent~~
- [ ] ~~Opacity~~

Known issues:

- In the example [WMS service](http://www.arcgis.com/home/item.html?id=1f80433204814a16964c0bcd3fcefaf8) provided by TNC, there are some layers that aren't found when matching up the layers specified in the config to the layers available in the service. It may have something to do with the fact that many of the names have `:` in them. For example, specifying `preview:Lakes_glwd1` in the config doesn't work, but `Lakes_glwd1` even though the name in the service is `preview:Lakes_glwd1`. Also, lots of layers won't match up with our without the suffix. For example `GAR2015:PGA_1500_yrp` doesn't work as `GAR2015:PGA_1500_yrp` or `PGA_1500_yrp`.
- Zoom to extent does not work: https://github.com/CoastalResilienceNetwork/GeositeFramework/issues/578
- Setting opacity is not currently supported: https://github.com/CoastalResilienceNetwork/GeositeFramework/issues/579

Also, for future consideration, https://github.com/CoastalResilienceNetwork/GeositeFramework/issues/580.

Ready for testing

**Testing instructions**
- Try adding layers from the two sample WMS services and verify that they show up on the map and the legend. The "Global Risk Data Platform" service is very slow.

Connects to #519

### DIFF
--- a/src/GeositeFramework/js/Map.js
+++ b/src/GeositeFramework/js/Map.js
@@ -225,17 +225,19 @@ require(['use!Geosite',
                 if (serviceInfo && service.visible && serviceInfo.pluginObject.showServiceLayersInLegend) {
                     service.visibleLayers.sort(function(a, b) { return a - b; });
                     _.each(service.visibleLayers, function(layerId) {
-                        var layer = _.findWhere(service.layerInfos, {id: layerId});
-                        if (!layer) {
-                            return;
+                        var layer,
+                            legend;
+
+                        if (isWms(service)) {
+                            layer = _getWMSLayer(service, layerId);
+                            if (!layer) { return; }
+                            legend = _getWMSLegend(layer);
+                        } else {
+                            layer = _getAGSLayer(service, layerId);
+                            if (!layer) { return; }
+                            legend = _getAGSLegend(service, layerId);
                         }
 
-                        var serviceLegend = getServiceLegend(service);
-                        if (!serviceLegend) {
-                            return;
-                        }
-
-                        var legend = _.findWhere(serviceLegend, {layerId: layerId});
                         if (isLayerInScale(service, layer)) {
                             result.push({
                                 service: service,
@@ -249,12 +251,41 @@ require(['use!Geosite',
             return result;
         }
 
+        function isWms(service) {
+            if (service.description && service.description.match(/WMS/i)) {
+                return true;
+            }
+            return false;
+        }
+
+        function _getWMSLayer(service, layerId) {
+            return _.findWhere(service.layerInfos, {name: layerId});
+        }
+
+        function _getWMSLegend(layer) {
+            return layer.legendURL;
+        }
+
+        function _getAGSLayer(service, layerId) {
+            return _.findWhere(service.layerInfos, {id: layerId});
+        }
+
+        function _getAGSLegend(service, layerId) {
+            var serviceLegend = getServiceLegend(service);
+
+            if (!serviceLegend) {
+                return;
+            }
+
+            return _.findWhere(serviceLegend, {layerId: layerId});
+        }
+
         // Filter out layers that are not visible at the current map scale.
         // Adapted from the ESRI dijit legend source code. (Ref: _isLayerInScale)
         function isLayerInScale(service, layer) {
             var scale = esriMap.getScale();
-            var minScale = Math.min(service.minScale, layer.minScale) || service.minScale || layer.minScale;
-            var maxScale = Math.max(service.maxScale, layer.maxScale);
+            var minScale = Math.min(service.minScale, layer.minScale) || service.minScale || layer.minScale || 0;
+            var maxScale = Math.max(service.maxScale, layer.maxScale) || 0;
             return minScale === 0 || minScale > scale && maxScale < scale;
         }
 
--- a/src/GeositeFramework/js/util/ajax.js
+++ b/src/GeositeFramework/js/util/ajax.js
@@ -11,14 +11,20 @@ define(['esri/request'],
         }
 
         // Fetch data from `url` and store the result in `cache` (memoized).
-        var fetch = function(url) {
+        var fetch = function(url, options) {
+            var options = options || {},
+                settings = _.defaults(options, {
+                    format: 'json',
+                    content: {f: 'json'}
+                });
+
             if (typeof promises[url] === 'undefined') {
                 promises[url] = request({
                     url: url,
-                    content: {f: 'json'},
-                    handleAs: 'json',
+                    content: settings.content,
+                    handleAs: settings.format,
                     callbackParamName: 'callback',
-                    timeout: 7000
+                    timeout: 20000
                 }).then(function(data) {
                     cache[url] = data;
                 }, function(error) {
--- a/src/GeositeFramework/plugins/layer_selector_v2/WmsService.js
+++ b/src/GeositeFramework/plugins/layer_selector_v2/WmsService.js
@@ -1,0 +1,156 @@
+define([
+        "dojo/_base/declare",
+        "dojo/Deferred",
+        "underscore",
+        "./lib/xmlToJson",
+        "framework/util/ajax",
+        "./util",
+        "./LayerNode",
+    ],
+    function(declare,
+             Deferred,
+             _,
+             xmlToJson,
+             ajaxUtil,
+             util,
+             LayerNode) {
+        "use strict";
+
+        var jsonResponseCache = {};
+
+        function get(url) {
+            return jsonResponseCache[url];
+        }
+
+        return declare(null, {
+            constructor: function(server) {
+                this.server = server;
+                this.defaultAjaxOptions = {
+                    format: 'text',
+                    content: ''
+                };
+            },
+
+            getServiceUrl: function() {
+                return util.urljoin(this._getBaseServiceUrl() + '?request=GetCapabilities&service=WMS');
+            },
+
+            _getBaseServiceUrl: function() {
+                // If name exists, this is likely an AGS WMS service,
+                // as opposed to GeoServer.
+                if (this.server.name) {
+                    return util.urljoin(this.server.url, this.server.name, 'MapServer/WMSServer');
+                } else {
+                    return this.server.url;
+                }
+            },
+
+            // Return a promise containing map service data.
+            fetchMapService: function() {
+                var self = this;
+                return ajaxUtil.fetch(this.getServiceUrl(), this.defaultAjaxOptions).
+                        then(function(response) {
+                            var responseJSON = self._transformServiceData(response);
+                            jsonResponseCache[self.getServiceUrl()] = responseJSON;
+                            return responseJSON;
+                        });
+            },
+
+            // Return cached map service data.
+            getServiceData: function() {
+                return get(this.getServiceUrl());
+            },
+
+            // Return a promise with service layer data.
+            fetchLayerDetails: function(tree, layerId) {
+                // There are no layer details for a
+                // WMS layer.
+                return new Deferred().resolve({});
+            },
+
+            // Return cached layer details.
+            getLayerDetails: function(serviceLayer) {
+                // There are no layer details for a
+                // WMS layer.
+                return {};
+            },
+
+            // Find the corresponding data for `layer` in the map service.
+            findServiceLayer: function(layer) {
+                var serviceData = this.getServiceData();
+
+                if (!serviceData || !layer) {
+                    return null;
+                }
+
+                return _.find(serviceData.layers, function(serviceLayer) {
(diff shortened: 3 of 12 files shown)


src/GeositeFramework/js/Map.js
```diff
@@ -225,23 +225,33 @@ require(['use!Geosite',
...
+                                legend = _.findWhere(serviceLegend, {layerId: layerId});
+                            }
+
+                            if (isLayerInScale(service, layer)) {
+                                result.push({
+                                    service: service,
+                                    layer: layer,
+                                    legend: legend
+                                });
+                            }
```
kdeloach: Consider splitting this out into separate functions like `if (isWms(service) { ... } else if (isAgs(service)) { ... }` etc.

src/GeositeFramework/plugins/layer_selector_v2/WmsService.js
```diff
@@ -0,0 +1,139 @@
...
+
+            findServiceLayerById: function(layerId) {
+                var serviceData = this.getServiceData();
+                if (serviceData) {
+                    return _.findWhere(serviceData.layers, { id: layerId });
+                }
+                return null;
+            },
+
+            supportsOpacity: function(serviceUrl) {
```
kdeloach: Don't think this argument is needed.

src/GeositeFramework/js/util/ajax.js
```diff
@@ -11,16 +11,27 @@ define(['esri/request'],
         }
 
         // Fetch data from `url` and store the result in `cache` (memoized).
-        var fetch = function(url) {
+        var fetch = function(url, format, postprocess) {
```
kdeloach: Consider accepting an `options` argument here instead that we can pass directly to the request function. We could use `_.defaults` to provide some default values.

src/GeositeFramework/plugins/layer_selector_v2/WmsService.js
```diff
@@ -0,0 +1,139 @@
...
+            fetchLayerDetails: function(tree, layerId) {
+                return new Deferred().resolve(function() {
+                    return tree.findLayer(layerId);
+                });
+            },
+
+            // Return cached layer details.
+            getLayerDetails: function(serviceLayer) {
+                return new Deferred().resolve(function() {
+                    return tree.findLayer(layerId);
```
kdeloach: `tree` Isn't in scope here.

src/GeositeFramework/plugins/layer_selector_v2/WmsService.js
```diff
@@ -0,0 +1,139 @@
...
+            // Return cached map service data.
+            getServiceData: function() {
+                return ajaxUtil.get(this.getServiceUrl());
+            },
+
+            // Return a promise with service layer data.
+            fetchLayerDetails: function(tree, layerId) {
+                return new Deferred().resolve(function() {
+                    return tree.findLayer(layerId);
+                });
```
kdeloach: Should this be resolved with a layer or a function that returns a layer?

src/GeositeFramework/js/util/ajax.js
```diff
@@ -11,16 +11,27 @@ define(['esri/request'],
...
-                    handleAs: 'json',
+                    content: content,
+                    handleAs: format,
                     callbackParamName: 'callback',
-                    timeout: 7000
+                    timeout: 20000
                 }).then(function(data) {
-                    cache[url] = data;
+                    if (postprocess && typeof postprocess === 'function') {
+                        cache[url] = postprocess(data);
```
kdeloach: I would vote to remove `postprocess` and let the WMS service handle this transformation and maybe manage its own cache. Since this cache is global I think only raw data should be preserved.

src/GeositeFramework/plugins/layer_selector_v2/WmsService.js
```diff
@@ -0,0 +1,139 @@
...
+                        // Compare not only the name, but the structure as well.
+                        // Protects against an edge case where a map service
+                        // contains a parent and child layer with the same name.
+                        if (layer.hasChildren() && serviceLayer.subLayerIds) {
+                            return true;
+                        } else if (!layer.hasChildren() && !serviceLayer.subLayerIds) {
+                            return true;
+                        }
+                    }
+                    return false;
```
kdeloach: This issue might not exist in WMS so we may be able to simplify this comparison.

src/GeositeFramework/plugins/layer_selector_v2/WmsService.js
```diff
@@ -0,0 +1,139 @@
...
+            constructor: function(server) {
+                this.server = server;
+            },
+
+            getServiceUrl: function() {
+                return util.urljoin(this._getBaseServiceUrl() + '?request=GetCapabilities&service=WMS');
+            },
+
+            _getBaseServiceUrl: function() {
+                // Likely an AGS WMS service
```
kdeloach: What does this comment mean?

caseypt: AGS can publish services as REST services or WMS services. Our module `AgsService` is really `AgsRESTService`. If the WMS service is published on AGS, you need to specify the name of the service. There is no such thing (as far as I can tell) for Geoserver WMS services. I say likely because I'm not totally sure this will always hold up. It seems to be true for the services I tested.

src/GeositeFramework/js/util/ajax.js
caseypt: Good idea. Will do.

src/GeositeFramework/plugins/layer_selector_v2/WmsService.js
caseypt: Yeah, I'm not clear if it exists either. I kept it as the default because if it was inheriting from the parent class, it might make sense to leave it as is until it fails or is too slow or something.

src/GeositeFramework/js/util/ajax.js
caseypt: Will do.

src/GeositeFramework/plugins/layer_selector_v2/WmsService.js
caseypt: I believe we were returning layers, but I'll take another look.

src/GeositeFramework/plugins/layer_selector_v2/WmsService.js
caseypt: Whoops. Thanks.

src/GeositeFramework/plugins/layer_selector_v2/WmsService.js
caseypt: These should probably just return null or an empty object since there are no layer details for a WMS layer.

src/GeositeFramework/plugins/layer_selector_v2/WmsService.js
kdeloach: Ah, thanks. Didn't realize that before.

src/GeositeFramework/js/Map.js
```diff
@@ -225,23 +225,33 @@ require(['use!Geosite',
...
-                            result.push({
-                                service: service,
-                                layer: layer,
-                                legend: legend
-                            });
+                        var layer,
+                            legend;
+
+                        if (_.isUndefined(service.description) === false) {
+                            if (service.description.match(/WMS/i)) {
```
kdeloach: An error occurs when `service.description` is null. Looks like it happens because the legend tries to render before the map service is completely loaded.

Steps:

1. Load WMS layer
2. Load AGS layer (ex. Study Area)

kdeloach:
It's nice how few changes were needed to add support for WMS.

caseypt:
Added two fixup commits. One to address comments on general WMS support, and the other to address problems and comments with the legend.

src/GeositeFramework/js/Map.js
caseypt: This should be fixed by 0edf7f0

src/GeositeFramework/plugins/layer_selector_v2/WmsService.js
```diff
@@ -0,0 +1,157 @@
...
+    function(declare,
+             Deferred,
+             _,
+             xmlToJson,
+             ajaxUtil,
+             util,
+             LayerNode) {
+        "use strict";
+
+        var jsonResponseCache = [];
```
kdeloach: I think this should be `{}`.

caseypt: Whoops. Thanks.

caseypt: Fixed in 7ac6681.

src/GeositeFramework/plugins/layer_selector_v2/WmsService.js
```diff
@@ -0,0 +1,157 @@
...
+                }
+            },
+
+            // Return a promise containing map service data.
+            fetchMapService: function() {
+                var self = this;
+                return ajaxUtil.fetch(this.getServiceUrl(), this.defaultAjaxOptions).
+                        then(function(response) {
+                            jsonResponseCache[self.getServiceUrl()] = self._transformServiceData(response);
+                            return self.responseJSON;
```
kdeloach: How does `self.responseJSON` get populated?

src/GeositeFramework/plugins/layer_selector_v2/WmsService.js
```diff
@@ -0,0 +1,157 @@
...
+                return get(this.getServiceUrl());
+            },
+
+            // Return a promise with service layer data.
+            fetchLayerDetails: function(tree, layerId) {
+                // There are no layer details for a
+                // WMS layer.
+                return new Deferred().resolve(function() {
+                    return {};
+                });
```
kdeloach: I don't know why dojo let's you do this but I think we should always resolve a deferred with a value instead of a function.

src/GeositeFramework/plugins/layer_selector_v2/layers.json
```diff
@@ -150,5 +150,43 @@
...
+        "displayName": "WMS Example Layers",
+        "server": {
+            "type": "wms",
+            "layerType": "dynamic",
+            "url": "http://sampleserver1.arcgisonline.com/ArcGIS/services/Specialty/",
+            "name": "ESRI_StatesCitiesRivers_USA"
+        },
+        "includeLayers": [
+            {
+                "name": "States"
```
kdeloach: Don't know if this is important but no service ID appears next to States when I load the map service.

image

caseypt: That's because it's 0.

src/GeositeFramework/plugins/layer_selector_v2/WmsService.js
caseypt: Okay. I'll switch it up.

src/GeositeFramework/plugins/layer_selector_v2/WmsService.js
caseypt: Fixing...

src/GeositeFramework/plugins/layer_selector_v2/layers.json
```diff
@@ -150,5 +150,43 @@
...
+            {
+                "name": "Cities"
+            },
+            {
+                "name": "Rivers"
+            }
+        ]
+    },
+    {
+        "displayName": "Global Risk Data Platform",
```
kdeloach: The legends for this map service won't load. Should we include this in our sample layers?

image

kdeloach:
+1 Works great.

src/GeositeFramework/plugins/layer_selector_v2/layers.json
caseypt: I included it because it's a service TNC plans on using. It just started having this error when generating legends a couple hours ago:

`java.lang.OutOfMemoryError: GC overhead limit exceeded GC overhead limit exceeded`

kdeloach: Yeah, I noticed that. Maybe we could add the layer name in the image alt attribute to give more context?

caseypt: Good idea. Added in 77289c5

caseypt:
Thanks for the through review.

PR #581, opened 2016-02-23
Remove toggleLayer delay

kdeloach:
When you expand the root folder for slow services (ex. Global Risk) the
tree is not re-rendered until the map service is fetched. This fixes
that by rendering the tree immediately.


kdeloach:
I noticed this while testing #577.

caseypt:
+1. Tested (I turned off preloading to verify this worked).

PR #583, opened 2016-02-24
Fix preloading of WMS layers.

caseypt:
* We need to specify that the response should be handled as text
for WMS layers.

**Testing instructions**
- There should be no errors when the layer selector is launched, and all WMS layers should be preloaded.

### DIFF
--- a/src/GeositeFramework/plugins/layer_selector_v2/main.js
+++ b/src/GeositeFramework/plugins/layer_selector_v2/main.js
@@ -413,7 +413,13 @@ define([
                 // Fetch all map services found.
                 var promise = all(_.map(serviceUrls, function(v, serviceUrl) {
                     // Cache map service response.
-                    return ajaxUtil.fetch(serviceUrl);
+                    var options = {};
+                    if (serviceUrl.match(/WMS/i)) {
+                        options.format = 'text';
+                        options.content = '';
+                    }
+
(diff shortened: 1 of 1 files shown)


src/GeositeFramework/plugins/layer_selector_v2/main.js
```diff
@@ -413,7 +413,13 @@ define([
...
                 var promise = all(_.map(serviceUrls, function(v, serviceUrl) {
                     // Cache map service response.
-                    return ajaxUtil.fetch(serviceUrl);
+                    var options = {};
+                    if (serviceUrl.match(/WMS/i)) {
+                        options.format = 'text';
+                        options.content = '';
+                    }
+
+                    return ajaxUtil.fetch(serviceUrl, options);
```
kdeloach: I think to account for minor differences between REST and WMS (and future services) we should start using those service classes to perform service operations. I'll try to think of a way to do this in the future (something like `createServiceFromUrl(serviceUrl).fetchMapService()` ?)

kdeloach:
+1

src/GeositeFramework/plugins/layer_selector_v2/main.js
caseypt: That makes sense to me.

Issue #578, opened 2016-02-23
Zoom to extent does not work for WMS layers


caseypt:
In #577, we construct an extent object from the WMS layer info that matches the structure of the AGS REST layer extent property. However, when setting the extent, an error occurs within Dojo:

![image](https://cloud.githubusercontent.com/assets/1042475/13260812/457f5ecc-da2b-11e5-8c1f-5d966dc6215f.png)

PR #584, opened 2016-02-24
Fix zoom to extent for WMS layers.

caseypt:
* Previously, we were using the CRS code and not the EPSG code (WMS
services provide both) when constructing the extent object. The JS
API expects EPSG codes.
* Handle differences in the structure of the XML bounding box attribute.

**Testing instructions**
- Try zooming to the extent of WMS layers in both of the demo services.

Connects to #578


kdeloach:
+1

PR #592, opened 2016-03-04
More unit tests

kdeloach:
This adds more unit tests as part of #561.

To test:

1. Uncomment this line https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/layer-selector-v2/src/GeositeFramework/plugins/layer_selector_v2/main.js#L30
2. Run application in browser
3. Verify all tests passed in developer console

image


rajadain:
Could not get tests to run. Firstly, uncommenting the line did not work as expected:

image

It seems that Dojo was interpreting the path to be `/plugins/layer_selector_v2/tests.js` instead of `/plugins/layer_selector_v2/tests/index.js`.

After I changed it from `./tests` to `./tests/index`, I get the following error:

image

kdeloach:
I corrected the path to the unit tests (good catch). Also, it may be possible that slow network speeds are causing this unit test to timeout somehow. I increased the timeout duration from 1 second to 5 seconds. Try again and let me know if this corrects the issue.

kdeloach:
Rebased and fixed failing unit test (rescinded the timeout increase).

rajadain:
Now I see this:

image

rajadain:
Sorry, also had to uncomment https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/layer-selector-v2/src/GeositeFramework/plugins/layer_selector_v2/main.js#L48

Working correctly now:

image

rajadain:
+1 tested.

PR #599, opened 2016-03-16
Update Gulf of Mexico Florida layers

kdeloach:
This fixes a dojo error that occurs when the page loads. The cause is
due to an ID mismatch between layers.json and the map service. Looks
like the map service was updated recently, so this change should bring
the layers.json file back in sync with the map service.


rajadain:
+1 I don't see the error any more. Checked out a handful of Gulf of Mexico layers and they load correctly. Great work finding the root of this!

PR #600, opened 2016-03-16
Update Gulf of Mexico Florida layers

kdeloach:
This fixes a dojo error that occurs when the page loads. The cause is
due to an ID mismatch between layers.json and the map service. Looks
like the map service was updated recently, so this change should bring
the layers.json file back in sync with the map service.


kdeloach:
Same as #599 but targets the layer-selector-v2 branch.

rajadain:
+1, confirmed on this branch as well.

Issue #598, opened 2016-03-14
Internationalize All Text


rajadain:
Following up on #589 and #597, wrap all text in the core app and all core plugins within an `i18next.t()` method so that it is internationalized. Also, add those phrases to the respective `en.json` file.

Issue #597, opened 2016-03-14
Concretize Internationalization Strategy


rajadain:
Continuing from #589, clean up and merge in the work done in https://github.com/rajadain/GeositeFramework/commits/feature/internationalization.

Also consider adding documentation and instructions for plugin developers on how to internationalize their plugins.

Issue #589, opened 2016-02-25
Develop Internationalization scheme


caseypt:
Give http://i18next.com/docs/ a look.

 * [ ] Update region schema
 * [ ] Add i18n module
 * [ ] Replace all hard-coded text/date/number formats
 * [ ] Create sample language string table
 * [ ] Create example Jenkins build?

rajadain:
There are fundamentally three kinds of places where strings need to be replaced:

 1. Direct strings in the native app, such as ["Split View"](https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/master/src/GeositeFramework/Views/Home/Index.cshtml#L199)
 2. Region strings, such as ["Flower Garden Banks Marine Sanctuary"](https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/master/src/GeositeFramework/region.json#L40)
 3. Plugin strings, such as ["Clear All"](https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/master/src/GeositeFramework/plugins/layer_selector/templates.html#L6)

1 and 3 should support multiple languages and locales. The path of localization maps should be devised and followed as a convention, with minimal configuration on the developer's part, so that plugin developers can just stick a translation file in a spot and have the rest of their plugin work as is.

Regarding 2, if the region supports only one language and locale, then nothing needs to be done. However, if a region is to support more than one language or locale, then we'll need to figure out where to put translation strings for those specified in the region's configuration.

rajadain:
A basic test app for internationalization is available here: https://github.com/rajadain/i18next-test

I've been using it to play with the [i18next](http://i18next.com/) library.

rajadain:
The approach to declaratively describe translation keys, like so:

```html
<a class="switch-screen i18n" data-i18n="[title]sidebar.viewOnlyMap1;sidebar.fullMap"></a>
```
Gets translated to

```html
<a class="switch-screen i18n" data-i18n="[title]sidebar.viewOnlyMap1;sidebar.fullMap" title="View Only Map 1">Full Map</a>
```

once

```js
$('.i18n').localize();
```

gets called. Currently, we do that once during app init. However, in the case of templates, this needs to be executed after each render.

kdeloach:
Declarative syntax could be a nice addition if we don't have to call `localize()` each time a template is rendered. At minimum, there should be a function that can handle translations for things outside the DOM (ex. `alert(__("Translate me"))`).

[...]

PR #602, opened 2016-03-18
Internationalization Scheme

rajadain:
## Overview

This is an attempt to devise a concrete internationalization scheme for the app that will allow both the core and plugin apps to be localizable for different locales. The locale to use will be specified in the instance's `region.json` file. Thus, the locale selection is instance-wide, and not user-set.

To facilitate string replacement, we use the [i18next](http://i18next.com) library, which comes with an [ecosystem of plugins](http://i18next.com/docs/ecosystem/) which serve most of the functionality we need. In addition to the core library, we use the following plugins:

 * [i18next-xhr-backend](https://github.com/i18next/i18next-xhr-backend): This allows us to specify a pattern for finding `.json` files on the server, and the `i18next` instance on the front-end to request it dynamically.
 * [i18next-jquery](https://github.com/i18next/jquery-i18next): Allows the use of internationalization in HTML templates in the form of `data-i18n` attributes. In addition to the attribute, the tags should have a `.i18n` class applied to them.
 * [i18next-sprintf-postProcessor](https://github.com/i18next/i18next-sprintf-postProcessor): Provides a `printf` like syntax, allowing the translation to have parameters in different positions given the language's grammar.

## Usage

Using this library, we can translate strings to other languages in the following ways:

 * In JavaScript, using `i18next.t('source phrase')`: Once the app is loaded and `i18next` has been initialized, it is attached to the `window` object and thus available everywhere. Within JavaScript, we can simply use `i18next.t('source phrase')` to translate strings. This will look up the `source phrase` in the supplied translation file, and return the found value. If the phrase is not found in the translation file, it will return `source phrase` itself.
 For strings with dynamic parameters, we can use the `printf` syntax:

  ```js
  i18next.t('Longitude %f, Latitude %f', lng, lat);
  ```

 And the translation file should have corresponding markers:

  ```json
  {
      "Longitude %f, Latitude %f": "Länge %f, Breite %f"
  }
  ```
 * In HTML, using the `data-i18n` attribute: This allows the specification of internationalization text right within the templates themselves. At the end of the first render, we go through all the marked tags in the entire app and localize them:

  ```js
  $('.i18n').localize();
  ```

 This replaces the contents of those tags with the translated value of the contents of `data-i18n`. If we wish to replace the value of another attribute, instead of the tag itself, then we prefix the string with the attribute's name in brackets. For example:

  ```html
  <a href="#" class="i18n" data-i18n="[title]source phrase">Text</a>
  ```

 Which outputs:

  ```html
  <a href="#" class="i18n" data-i18n="[title]source phrase" title="translated phrase">Text</a>
  ```

 The `.i18n` class is used by the `.localize()` call above to find and replace the value in these tags. It must also be called on the view's `$el` every time the view renders, since the replacements in the HTML template will have reset. An example of this can be seen in the `renderSidebar` method in `Pane.js`.

## Translations

The translations themselves are stored as simple JSON files, each named for the locale it translates to. There are three locations for them:

 * `/locales/{{ locale }}.json`: The translations for the core app.
 * `/{{ plugin_path }}/locales/{{ locale }}.json`: Translations for the plugin
 * `/languages/{{ locale }}.json`: Combined file containing all translations in the app

Translations for English are not technically necessary, since we fallback to the phrase if we don't find the key, and the phrases are already in English. However, for the purpose of creating a guide to be used for translating to other languages, we maintain an English translation file which simply translates each phrase to itself. For example, `/locales/en.json` has entries like:

```json
{
    "Map Legend": "Map Legend",
    "Hide Legend": "Hide Legend",
    "View Only Map 1": "View Only Map 1",
    "View Only Map 2": "View Only Map 2",
    "Full Map": "Full Map",
    
}
```

Plugins must provide their own translations. As an example, we have supplied them for `sample_plugins/identify_point` in its `locales/en.json`:

```json
{
    "Identify Point": "Identify Point",
    "Identify point sample plugin": "Identify point sample plugin",
    "Click any point on the map to display Latitude and Longitude": "Click any point on the map to display Latitude and Longitude",
    "You clicked on latitude %f longitude %f": "You clicked on latitude %f longitude %f"
}
```

When the app starts, we go through each plugin's and add them to a master file. Finally, we add the core translations. In case of collisions, later additions will overwrite earlier ones. This ensures that no one plugin's translation can override the presentation of the entire app.

The combined files are stored in `/languages/{{ locale }}.json`. This folder is ignored by git, but looked for by `i18next` when the app starts up, as configured under `internationalize` in `App.js`.

## Testing Instructions

Open the app and ensure that the tooltips and labels for map controls in the bottom left appear and work correctly. Ensure that the Identify Point plugin works and appears correctly.

## Code Review Instructions

Ensure that the JavaScript libraries are loaded in the correct order. Suggest ideas for making the translation-merge more efficient (`PrepareLanguages` in `Global.asax.cs`).

## Next Steps

This effort designs a strategy for internationalizing the app and the plugins. This strategy, once validated, must be used to internationalize the rest of the app and plugins. For this task, we have #598.

Connects #589
Connects #597


caseypt:
Strategy looks good to me. Currently testing out the app.

Consider moving the usage instructions above to the [wiki](https://github.com/CoastalResilienceNetwork/GeositeFramework/wiki).

caseypt:
I get a "Sorry, an error occurred while processing your request." message when trying to load the app. It it's blowing up on line 140 of `Global.asax.cs` (when it tries to delete the locales path). To fix it, I had to create the directory it's trying to delete first.

caseypt:
Regarding the above error, is it safe to assume IIS will always have write permissions?

caseypt:
+1. Nice job. Tested by adding my own translation file.

rajadain:
Re: blowing up on delete, good catch, will ensure that only happens if the directory doesn't already exist.
Re: IIS having write permissions, will add that to the Wiki in addition to the instructions from above.

Thanks for taking a look!

rajadain:
Will now only delete the folder if it exists, in 5f40fc4.

caseypt:
+1. Tested out the fix.

rajadain:
Added above documentation to the Wiki: https://github.com/CoastalResilienceNetwork/GeositeFramework/wiki/Internationalization. I feel like there should be a more obvious section for plugin developers which lists what they have to do step by step, but this should be good enough to give them an overview for now. Merging shortly. Thanks everyone for your feedback and review!

PR #604, opened 2016-03-22
Setup text for translation

caseypt:
This is a first pass at setting up text in the UI so that it can be
translated.

* Wrap UI text in the core framework and plugins that should be
translated.
* Add text to en.json files for core framework and plugins.

**Testing instructions**
- Verify that a sample of the translations are set-up correctly.
- Look for areas in the app that may have been missed.

Connects to #598


src/GeositeFramework/Views/Home/Index.cshtml
```diff
@@ -350,20 +350,20 @@
...
             <div class="popover-content">
                 <div class="row" id="permalink-dialog">
                   <label>Permalink:<input class="code-like permalink-textbox" type="text" value="<%= shortUrl %>"/></label>
-                  <p>Share your map with friends or family, or just bookmark the link for later.</p>
-                  <label>Embed this map in your website:</label>
+                  <p class="i18n" data-i18n="Share your map with friends or family, or just bookmark the link for later.">Share your map with friends or family, or just bookmark the link for later.</p>
+                  <label class="i18n" data-i18n="Embed this map in your website:">Embed this map in your website:</label>
                   <div id="iframe-embed"></div>
-                  <label>Embed Width (px)
+                  <label><span class="i18n" data-i18n="Embed Width(px)">Embed Width (px)</span>
```
rajadain: Missing space between "Width" and "(px)"

src/GeositeFramework/plugins/layer_selector/templates.html
```diff
@@ -2,8 +2,8 @@
 <script type="text/template" id="template-layer-selector">
     <div class="pluginLayerSelector-outer">
         <div class="pluginLayerSelector-search">
-            <input type="text" placeholder="Filter Map Layers"/>
-            <a class="pluginLayerSelector-clear" href="javascript:;">Clear All</a>
+            <input type="text" class="i18n" data-i18n="[placeholder]Filter Map Layers" placeholder="Filter Map Layers"/>
```
rajadain: Cool use of the `placeholder` prefix

src/GeositeFramework/plugins/layer_selector/WmsLoader.js
```diff
@@ -83,7 +83,8 @@ define(["use!underscore", "esri/layers/WMSLayer", "esri/layers/WMSLayerInfo"],
                 });
                 dojo.connect(wmsLayer, "onError", function(err) {
                     esri.config.defaults.io.timeout = 60000;
-                    alert("Error: Unable to load data from this service. Either the service is unavailable or <br> the request can't be completed at the current map scale or browser size.");
+                    var alertMsg = i18next.t("Error: Unable to load data from this service. Either the service is unavailable or <br> the request can't be completed at the current map scale or browser size.");
```
rajadain: I'm not 100% sure if this `<br>` will work or will be HTML escaped ...

rajadain:
One of the things we need to do is call `$(view.$el).localize();` on each render of each internationalizable view, so that the `data-i18n` attributes are processed and applied. You can see an example of this [here](https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/develop/src/GeositeFramework/js/Pane.js#L306-L310):

```js
// If i18n has been initialized,
if ($.i18n) {
    // Internationalize everything with a data-i18n attribute
    $(view.$el).localize();
}
```

We first check if `$.i18n` has been initialized since the view might be rendered the first time before i18next's initialization. Secondly, we use `$(view.$el).localize();` instead of `view.$('.i18n').localize();` which seems on the surface to be more efficient, because when we initialize the `i18n-jquery` plugin, it attaches itself to the globally scoped instance of jQuery, as observable [here](https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/develop/src/GeositeFramework/js/App.js#L158-L160). Thus, we must use the global instance of jQuery for this purpose.

Without it, dynamically rendering views do not get translated:

image

In the above screenshot, I've replaced all translations with `xxx`, as can be seen in the sidebar buttons in the background. The foreground, however, has not been translated.

rajadain:
I will take another look once these comments have been addressed.

src/GeositeFramework/Views/Home/Index.cshtml
caseypt: Good catch.

src/GeositeFramework/plugins/layer_selector/WmsLoader.js
caseypt: Good catch. I tested it out and it doesn't work. I'll remove the `<br>`. This plug-in is being phased out too.

caseypt:
Added a commit to translate some of the dynamically rendered views. I'll pick up where I left off tomorrow.

caseypt:
Updated the fixup commit with more calls to translate plugins as well as more translated text. Ready for another look.

src/GeositeFramework/Views/Home/Index.cshtml
```diff
@@ -250,16 +250,16 @@
...
             <div class="plugin-container">
                 <div class="plugin-container-header">
-                    <h6><%- title %></h6>
+                    <h6 class="i18n" data-i18n="<%- title %>"><%- title %></h6>
                     <a class="plugin-off" href="javascript:;">&#10006;</a>
                     <a class="plugin-close" href="javascript:;">&#95;</a>
 
                     <% if (hasCustomPrint) { %>
-                        <a class="plugin-print" href="javascript:;" title="Print Contents"><i class="fa fa-print"></i></a>
+                        <a class="plugin-print i18n" data-i18n="[title]Print contents" href="javascript:;" title="Print Contents"><i class="fa fa-print"></i></a>
```
rajadain: This doesn't translate correctly because of an inexact string match: should be "Print Contents", is "Print contents"

src/GeositeFramework/Views/Home/Index.cshtml
```diff
@@ -305,19 +305,19 @@
...
-                        <label><input type="radio" name="export-orientation" value="Landscape" checked="checked"/>Landscape</label>
-                        <label><input type="radio" name="export-orientation" value="Portrait"/>Portrait</label>
+                        <h5 class="i18n" data-i18n="Orientation">Orientation</h5>
+                        <label><input type="radio" name="export-orientation" value="Landscape" checked="checked"/><span class="i18n" data-i18n="Landscape">Landscape</label>
+                        <label><input type="radio" name="export-orientation" value="Portrait"/><span class="i18n" data-i18n="Portrait">Portrait</label>
                     </div>
                     <div class="small-4 columns">
-                         <h5>Legend</h5>
-                        <label><input type="checkbox" name="export-include-legend" />Include Layer Legend</label>
+                         <h5 class="i18n" data-18n="Legend">Legend</h5>
```
rajadain: Typo: should be `data-i18n`, is `data-18n`

rajadain:
The Map Zoom button titles are not translated, but I wonder if that is beyond our capacity since they ship directly from ESRI:

image

rajadain:
The app title, menu bar, and map drop down are not translated:

image

But this is okay, since I think all of this is specified in `region.json`, and could be specified in the target language therein.

rajadain:
The title bar for "Save & Share" isn't translating correctly:

image

I'm not sure why. At first I thought it might be the ampersand, but it is handled correctly in the label on the sidebar.

caseypt:
> But this is okay, since I think all of this is specified in region.json, and could be specified in the target language therein.

Yeah, that's what I was thinking.

rajadain:
Identify Point is missing translations in the following two spots:

image

image

rajadain:
Also the "Continue" button in the Identify Point opening dialog.

caseypt:
One spot that I couldn't get to translate after some time trying to was the measure tool dialog. It's all within an ESRI widget, and calling `localize` in various places did not work.

rajadain:
And I imagine these are mostly source from configuration as well, thus could be localized there:

image

rajadain:
Good, very thorough work! I don't think we need to break our backs over the small final details, I just recorded everything I found so that we have a known issues list.

Once those few typos are fixed, I think this is good to go. Friendly reminder to squash before merging.

caseypt:
Pushed one more fixup commit to fix the typos and some of the other translation problems. Other issues have been moved to https://github.com/CoastalResilienceNetwork/GeositeFramework/issues/606.

caseypt:
Thanks for the thorough review @rajadain. I'll rebase and merge shortly.

rajadain:
+1!

Issue #609, opened 2016-03-24
Legend minimize / maximize icon


mmcfarland:
When the legend is minimized, the `_` symbol should either turn to a `+` or perhaps an "expand icon" like a square.

PR #615, opened 2016-03-25
Toggle the min/max icon for legend

mmcfarland:
Swap out a _ for a + when toggling the min/max for Legend.

Connects #609 

To test, minimize the legend and restore it noting the change in icon and title text.


kdeloach:
+1

PR #616, opened 2016-03-28
Merge develop --> layer-selector-2

mmcfarland:
Keeps up with commits made to the develop branch, including internationalization features.  This will allow layer V2 to be internationalized

Test that Layer V2 works and that switching to `es` in region.json initiates the framework internationalization feature.


kdeloach:
Identify point plugin has some `undefined` text.

image

mmcfarland:
That's a bug in `develop` as of this merge, but has been fixed since in https://github.com/CoastalResilienceNetwork/GeositeFramework/commit/d32a144135f066d506138df417b4986e4061e626.

I'm thinking of holding off on more merges since this branch will hopefully soon be worked into develop, and I mostly just wanted to get the base translation stuff into v2 so I could translate that plugin.

kdeloach:
+1 Looks good.

I don't have any `es` locale files so the app throws a Runtime Exception when I use any language besides `en`. No error appears in my logs either. May want to address this in a future PR.

mmcfarland:
Good call, I'll make an issue for mis-configured language settings.

Issue #603, opened 2016-03-21
Translate layer selector v2 templates


caseypt:
When the `layer-selector-v2` branch is merged into `develop`, translate the templates. There isn't too much to do.

caseypt:
This also includes the draw and report plug-in, which is a plug-in to the layer selector. This may require a separate locale file for the layer selector and draw and report. If that is the case, make sure the draw and report file is brought into the compiled languages file.

PR #622, opened 2016-03-29
Translate Layer Selector V2

mmcfarland:
Translates the main template area, draw and report and the draw tooltips.

This commit contains a "test" spanish language translation, which should allow you to set the region language setting to "es" without error.  

Connects #603 

Test that the english text in the plugin that does not come from a server is translated to spanish.

screenshot from 2016-03-29 16 31 28

### DIFF
--- a/src/GeositeFramework/plugins/layer_selector_v2/draw_report/main.js
+++ b/src/GeositeFramework/plugins/layer_selector_v2/draw_report/main.js
@@ -23,7 +23,8 @@ define(["dojo/_base/declare",
         "esri/tasks/Geoprocessor",
         "esri/tasks/FeatureSet",
         "esri/toolbars/draw",
-        "dojo/text!./templates.html"
+        "dojo/text!./templates.html",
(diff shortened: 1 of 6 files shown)


src/GeositeFramework/plugins/layer_selector_v2/draw_report/main.js
```diff
@@ -58,6 +60,16 @@ define(["dojo/_base/declare",
                 this.bindEvents();
             },
 
+            // Translate the built-in draw tool text
+            setupDrawingText: function() {
+                var drawText = esriText.toolbars.draw,
+                    props = ['start', 'resume', 'complete'];
+
+                _.each(props, function(prop) {
+                    drawText[prop] = i18next.t(drawText[prop]);
```
kdeloach: I think we should be using abbreviations for i18n keys. If we ever upgrade to a newer version of the JSAPI and these text values change, our translation files will silently break.

Issue #621, opened 2016-03-29
Translation File Changes


mmcfarland:
* If the desired language setting does not exist, or the resulting file does not parse (JSON syntax error), the site shows a runtime error but no log information.  We should try to write some information to the geosite log.
* The merged region language files are written to disk but have been causing directory permission issues.  We should just hold the language objects in memory and have them recreated on `App_Start` instead of writing to file.

mmcfarland:
Should fix both #610  and #620

Issue #620, opened 2016-03-29
i18n UnauthorizedAccessException


kdeloach:
I'm getting various permissions errors on the i18n files and folders. May be more of a system configuration issue than a coding error.

Running VS debug server throws this error:
![image](https://cloud.githubusercontent.com/assets/43062/14117336/38c1f90e-f5b2-11e5-870f-0c20746f18d1.png)

The languages folder exists but is set to Read-only. It's impossible for me to change this (even as admin):
![image](https://cloud.githubusercontent.com/assets/43062/14117377/701801f0-f5b2-11e5-829b-c018a2439b6f.png)

Additionally, any exception that occurs while assembling language files will throws a runtime error and will not log the exception to the error log.

mmcfarland:
I wonder if we could get past those issues and the one reported in #610 if we didn't write a region-language file but instead just kept it in memory.  I believe we do something like this for the parsed region.json file.  If it's going to re-create the file on `App_Start` anyway, the actual file on disk isn't doing much except cause permission issues.

caseypt:
Yeah, this is probably what is causing #610.

kdeloach:
I like that idea.

Issue #610, opened 2016-03-24
Fix IIS issue with translation file


caseypt:
Currently, our internal build of the GeositeFramework is failing to generate the compiled languages file. 

This could be an issue with IIS or with the old build system.

I tried various changes to the permissions of the directory through IIS, but I was unable to fix this.

One solution may be to upgrade the job to use the new build system. That would involve, however, extracting out the region file from this repo into it's own repo.

PR #624, opened 2016-03-30
Hold language translations in memory

mmcfarland:
To avoid having to write/delete files on disk during App_Start, hold the
translation results as a global static and expose it as a route to the
front end app.

_sorry for the whitespace noise in the commit, new editor_

Connects #610 
Connects #620 
Connects #621 

#### Test
* Ensure you have no `/languages/` directory (delete if necessary)
* Load app and ensure `en` xhr request succeeds (`/languages/en`)
* Change your region language setting to `es`
* Check that translation `es` request succeeds and the app is translated
* Check that a bogus language file logs the error in `/logs/GeositeFramework.log` (ex: http://localhost:54633/languages/ghostwolfe)

### DIFF
--- a/src/GeositeFramework/App_Start/RouteConfig.cs
+++ b/src/GeositeFramework/App_Start/RouteConfig.cs
@@ -24,6 +24,12 @@ namespace GeositeFramework
             );
 
             routes.MapRoute(
+                name: "Language",
+                url: "languages/{locale}",
+                defaults: new { controller = "Language", action = "GetLocale"}
+            );
+
+            routes.MapRoute(
                 name: "Default",
                 url: "{controller}/{action}/{id}",
                 defaults: new { controller = "Home", action = "Index", id = UrlParameter.Optional }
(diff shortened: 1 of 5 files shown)


caseypt:
Re: whitespace, you can turn that off in Settings > Packages > Whitespace > Remove Trailing Whitespace.

rajadain:
Taking a look now ...

rajadain:
+1! Great work, this is how it should have been implemented in the first place. Confirmed that languages are working, that the XHR is correct, and errors are logged correctly:

```
2016-03-31 14:31:56,001 ERROR - System.Collections.Generic.KeyNotFoundException: The given key was not present in the dictionary.
   at System.Collections.Generic.Dictionary`2.get_Item(TKey key)
   at GeositeFramework.Controllers.LanguageController.GetLocale(String locale) in c:\projects\GeositeFramework\src\GeositeFramework\Controllers\LanguageController.cs:line 20
```

When I configured `region.json` incorrectly on purpose, an `HttpException` with the message "Invalid locale setting" was thrown, and the UI fell back to the English key phrases.

Code looks great too! :+1:

rajadain:
The `HttpException` was not written to the log file though. Should it have been?

rajadain:
Oh wait I see we log the actual exception before throwing a new custom one. Nevermind.

Issue #607, opened 2016-03-24
Layer Selector V2 Styling changes


mmcfarland:
A specific list of style considerations were provided which are essentially summed up as "Create more contrast between visual elements on the plugin".  We should use black text and a lighter gray background for highlight states and a more contrasting color for row-icons.  Zach suggests black/white would be fine.

Change display elements
- overall the titles of the layers and symbols are pastel or subtle. The titles should be in black font (the zoom to extent, opacity, etc, are good and bold, and the font seems like the right color black)
- need a better symbol for the data layer, and be bolder. 
- the eye is good, but the identify and more icons need to be bolder, and a different color when activated

jfrankl:
Darken text color to #222:

```css
.layer-selector2 .tree-container li {
    position: relative;
    color: #222;
}
```

Replace database icon with [`circle`](http://fortawesome.github.io/Font-Awesome/icon/circle/) and [`circle-o`](http://fortawesome.github.io/Font-Awesome/icon/circle/). Show `circle-o` on an unselected layer and `circle` on a selected layer.

Remove the eye icon. The new circle icon, with its two states, serves a similar role.

![image](https://cloud.githubusercontent.com/assets/1809908/14213514/5bde78c6-f806-11e5-9950-34c55997e135.png)

PR #625, opened 2016-04-01
Layer selector v2 style updates

caseypt:
Update layer selector v2 style.

* Per instructions from @jfrankl in #607

Connects to #607

image


caseypt:
Tagging @jfrankl as well.

mmcfarland:
+1, looks good.

PR #627, opened 2016-04-04
WIP: Fixed header

caseypt:
This is close, but there are a few problems, and it might be dead-end solution.

I currently have the tabs and filter container fixed, and the layer list is scrollable, however, the bottom of the list is cut off at the bottom. It's tough to tell, but it's illustrated in this screenshot:

image

The bottom of the scroll bar should be visible.

The other problem is that the draw and report results container isn't scrollable at all, despite having a similar structure to the layer list.

### DIFF
--- a/src/GeositeFramework/plugins/layer_selector_v2/draw_report/style.css
+++ b/src/GeositeFramework/plugins/layer_selector_v2/draw_report/style.css
@@ -8,6 +8,13 @@
     padding: 12px;
 }
 
+.layer-selector2 .tab-report .report-intro {
+    overflow: auto;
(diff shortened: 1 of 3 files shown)


kdeloach:
Consider moving this to the regional-planning repo.

caseypt:
I'm going to leave this here for now so it's easier for @jfrankl to take a look at, but I'll be sure to apply the changes to the regional planning repo once they are finalized.

caseypt:
Superseded by https://github.com/CoastalResilienceNetwork/regional-planning/pulls/17

PR #632, opened 2016-04-08
All infoGraphic to be an HTML snippit or an img...

mapossum:
The snippit allows the code to detect if the infoGraphic is specified as an image (.png or .jpg).  If not it assumes that it is a an html snippit and inserts into the same location.  Done at the request of Zach, should not impact current plugins that specify an image.

### DIFF
--- a/src/GeositeFramework/js/Plugin.js
+++ b/src/GeositeFramework/js/Plugin.js
@@ -668,8 +668,13 @@ require(['use!Geosite',
                 var pluginModel = this.model,
                     pluginObject = pluginModel.get('pluginObject');
 
-                var img = $('<img class="graphic" />').attr('src', pluginObject.infoGraphic);
-                this.$el.append(img);
+	       if (pluginObject.infoGraphic.slice(pluginObject.infoGraphic.length - 4, pluginObject.infoGraphic.length) == ".jpg" || pluginObject.infoGraphic.slice(pluginObject.infoGraphic.length - 4, pluginObject.infoGraphic.length) == ".png") {
+		var snippit = $('<img class="graphic" />').attr('src', pluginObject.infoGraphic);
+	       } else {
+		var snippit = $(pluginObject.infoGraphic);
+	       }				
+				
+                this.$el.append(snippit);
 
                 var checkboxnode = $('<span>').get(0);
                 this.$el.append(checkboxnode);


src/GeositeFramework/js/Plugin.js
```diff
@@ -668,8 +668,13 @@ require(['use!Geosite',
                 var pluginModel = this.model,
                     pluginObject = pluginModel.get('pluginObject');
 
-                var img = $('<img class="graphic" />').attr('src', pluginObject.infoGraphic);
-                this.$el.append(img);
+	       if (pluginObject.infoGraphic.slice(pluginObject.infoGraphic.length - 4, pluginObject.infoGraphic.length) == ".jpg" || pluginObject.infoGraphic.slice(pluginObject.infoGraphic.length - 4, pluginObject.infoGraphic.length) == ".png") {
+		var snippit = $('<img class="graphic" />').attr('src', pluginObject.infoGraphic);
+	       } else {
+		var snippit = $(pluginObject.infoGraphic);
+	       }
```
kdeloach: It looks like a mix of spaces and tabs are used for indentation here. Consider using all-spaces for consistency.

src/GeositeFramework/js/Plugin.js
```diff
@@ -668,8 +668,13 @@ require(['use!Geosite',
                 var pluginModel = this.model,
                     pluginObject = pluginModel.get('pluginObject');
 
-                var img = $('<img class="graphic" />').attr('src', pluginObject.infoGraphic);
-                this.$el.append(img);
+	       if (pluginObject.infoGraphic.slice(pluginObject.infoGraphic.length - 4, pluginObject.infoGraphic.length) == ".jpg" || pluginObject.infoGraphic.slice(pluginObject.infoGraphic.length - 4, pluginObject.infoGraphic.length) == ".png") {
```
kdeloach: Consider using the `endsWith` function. So this

```
pluginObject.infoGraphic.slice(pluginObject.infoGraphic.length - 4, pluginObject.infoGraphic.length) == ".jpg"
```

could become

```
pluginObject.infoGraphic.endsWith(".jpg")
```

mmcfarland: We recently got bit by an `endsWith` bug not being supported in any version of IE (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith).  I know @caseypt added a shim, but that might be only in regional-planning.  I think the same concept, but with a substring evaluation could still shorted the code considerably.

Issue #637, opened 2016-04-26
Add es.json files


ajrobbins:
From Zach:
the following links need an es.json file added to them before making this part of the master framework branch:
- https://github.com/CoastalResilienceNetwork/GeositeFramework/tree/develop/src/GeositeFramework/plugins/full_extent/locales
- https://github.com/CoastalResilienceNetwork/GeositeFramework/tree/develop/src/GeositeFramework/plugins/zoom_to

PR #638, opened 2016-04-26
Translate additional plugins

caseypt:
* Adds sample Spanish language files for the full_extent, zoom_to, and subregion_toggle
plugins.

**Testing**
- Set the language of the region to `es` and verify that the plugins are translated.

image

image

image

Connects to #637


mmcfarland:
+1, tested.

Issue #623, opened 2016-03-30
iFrame syntax not working to embed map in web page


zferdana:
We have noticed that the save & share iframe syntax doesn't work properly when embedding the map into a web page. The map has a blank base map and the plugins are also blank. Example to look at include http://coastalresilience.org/test-embed/ and http://slr.stormsmart.org/resilience-ds-tool/

PR #648, opened 2016-04-28
Downgrade jQuery scrollbar library

caseypt:
* Version 3.1.3 of this library (upgraded in d1969e6) causes a JS error
when the site is loaded from within an iframe. The release notes for
3.1.3 mention that a cross-domain iframe issue was fixed by this release,
but it appears that a different issue was introduced. To allow
the site to function, we downgrade to a version of the library
that works in an iframe.

**Testing**
- Create a basic HTML file that has your local site embedded as an iframe. Something like:
```html
<!doctype html>
<html lang="en">
    <head>
        <meta charset="utf-8">
    </head>
    <body>
        <iframe width="500" height="400" src="http://cthom-windev-vm:5001/"></iframe>
    </body>
</html>
```
- Run a local webserver to serve this file and access the site.
- Verify the site loads without any JS errors.

Connects to #623


mmcfarland:
+1, iframes are working.

screenshot from 2016-04-28 13 35 47

Issue #644, opened 2016-04-27
Move framework develop to master


ajrobbins:
Once https://github.com/CoastalResilienceNetwork/GeositeFramework/issues/637 is complete, move internationalization to master.

mmcfarland:
Pre-req:  Update spanish translations from TNC

mmcfarland:
Translations are updated, and we're waiting for 5/31 to "release" to `master` to allow TNC to have time to test things out.

mmcfarland:
This is the card to handle the last bit of release which will be done after any last minute feedback.

PR #649, opened 2016-04-28
Update core translations

mmcfarland:
New Spanish translations from TNC

Connects #644 

##### Test:
See that new translations are shown in framework and measure.


caseypt:
+1. Tested.

PR #652, opened 2016-05-13
Remove layer_selector plugin

mmcfarland:
#### RIP Original layer selector.  
This plugin has been removed as a core plugin, but is still available at
https://github.com/CoastalResilienceNetwork/map-tree and can be included
through a build process.  Additional necessary changes:
* Remove references is core region framework to layer_selctor
* Remove sample_plugin which showed how to extend layer_selector
* Move library `tv4`, which was included with layer_selector, but was
  refereced via `use!` by multiple other plugins, to the core framework.

Connects https://github.com/CoastalResilienceNetwork/GeositeFramework/issues/653

#### Testing
* Verify the site loads and the tests pass
* Verify layer_selector is not found
* Add map tree into a `layer_selector` directory in `plugins/`, 
```
git clone git@github.com:CoastalResilienceNetwork/map-tree.git layer_selector
``` 
  and verify it still works


src/GeositeFramework/region.json
```diff
@@ -1,5 +1,5 @@
 {
-    "language": "en",
+    "language": "es",
```
caseypt: Leftover from debugging?

caseypt:
I think we knew this would happen, but plugins that are dependent on the layer selector don't work once the layer selector is removed from the framework.

caseypt:
I'm getting this error when trying to load the site:

image

mmcfarland:
@caseypt I'm not getting that error.  Can you make sure you kill your appserver/hard refresh and see if you still get it?

caseypt:
Yeah. I closed down Visual Studio, cleared the browser cache, etc.

This is what the `use` config object contains when it tries to find the config for `tv4`.

image

caseypt:
I'm not totally sure what this is actually doing, but this change fixes it for me: 

```diff
diff --git a/src/GeositeFramework/Views/Home/Index.cshtml b/src/GeositeFramework/Views/Home/Index.cshtml
index 45db79b..2d6e2f3 100644
--- a/src/GeositeFramework/Views/Home/Index.cshtml
+++ b/src/GeositeFramework/Views/Home/Index.cshtml
@@ -594,6 +594,7 @@
         var useConfig = { @Html.Raw(Model.ConfigurationForUseJs) };
         useConfig['Azavea'] = { attach: 'Azavea' };
         useConfig['Geosite'] = { attach: 'Geosite' };
+        useConfig['tv4'] = { attach: 'tv4' };

         require({
             async: true,
```

caseypt:
+1 (besides the language change). Tested with the `regional-planning` change. Looks good.

mmcfarland:
Cool, removed the language setting change.  Good catch on that.

rajadain:
+1 tested

PR #654, opened 2016-05-13
Translate `Next` for infographics

mmcfarland:
"Next" replaced "Continue" and needs to be translated.

Test that infographics on `es` are translated.

screenshot from 2016-05-13 17 49 00
screenshot from 2016-05-13 17 47 44


caseypt:
+1. Tested.

Issue #657, opened 2016-06-01
custom legend


zferdana:
The following seems to be a common theme since migrating over to the new framework and legend:

I’m having a hard time figuring out the best way to override the new legend plugin modifications.  It isn’t picking up our custom class breaks like the old one did, but is instead using the default style from the server.  I can’t find a clean way with the plugin API to override the legend container and legend title for our plugin.  Is there a built in way that this should be done?
 
I can create custom content in the legend using this method:
 
$(this.legendContainer).html("Custom legend here");
 
But that doesn’t get rid of the layer-legend that is automatically loaded when we add the layer to the map, which we don’t want to show as is.
 
I’d appreciate any pointers you might have on how we are supposed to do this, before I try and find an “alternative” solution.

kdeloach:
Could you provide the name of a reference plugin we could use to reproduce the problem?

zferdana:
sure, the Natural Coastal Protection plugin on http://dev.maps.oceanwealth.org/ is experiencing this problem

dharlan:
HydroFlows legend has boxes around each legend item, causing the legend to get larger than needed.    This can been see here http://54.241.26.242/mississippi-dev/.

mmcfarland:
@zferdana This appears to be due to the use of a custom renderer for the layer.  Since we ditched the confining ESRI Legend widget we are making requests directly to the legend endpoint, which are not reflecting the custom class breaks.  This may be more than a quick fix to integrate with the custom renderer.

An interim solution using the available framework features might be:
1. Disable the framework from auto showing the plugin layer in the legend by [setting `showServiceLayersInLegend = false`](https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/71c05132c2762be06b439c09abb72cd116d31e5d/src/GeositeFramework/js/PluginBase.js#L30) in your main.js file
2.  Add a custom legend as demonstrated here:  https://github.com/CoastalResilienceNetwork/regional-planning/commit/b7098729032b0a9ce1ad733dc009ef37639c50df

mmcfarland:
[...]

PR #658, opened 2016-06-02
Override styles for tables in legend

mmcfarland:
Plugins which use the ESRI Legend Dijit internally may place table markup
in the custom legend container.  These have excessive borders and
paddings.  This commit removes those styles so that legend items appear
uniform.

Partially connects #657 

#### Testing
Testing is difficult since one cannot clone HydroFlow into a dev framework and have it work, it needs special configuration.  Easiest test is to apply the css rule provided here to developer tools on http://54.241.26.242/mississippi-dev/ and ensure the legend styling is enhanced.

### DIFF
--- a/src/GeositeFramework/css/app.css
+++ b/src/GeositeFramework/css/app.css
@@ -668,6 +668,16 @@ input.code-like, textarea.code-like {
         margin-bottom: 5px;
     }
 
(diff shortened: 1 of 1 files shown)


mmcfarland:
Updated with more specific css.

caseypt:
+1. Tested.

Issue #661, opened 2016-07-21
Services with string ids for layers fail to load legends


caseypt:
It seems that in all cases, the `layerId` property for layers in `service.layerInfos` are numbers. 

However, in some cases, the `layerId` for layers that are returned by `esriMap.getLayersVisibleAtScale` are strings while other times they are numbers. When they are strings, this statement fails to return any fails because the id comparison fails: https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/master/src/GeositeFramework/js/Map.js#L277.

If we ensure that `layerId` is a number (by using `parseInt`) in the context of the above function, the legend works for the Restoration Explorer plugin.

According to the documentation the id on the layer will always be a string (https://developers.arcgis.com/javascript/3/jsapi/layer-amd.html#id), but the id on the layer in a layerInfo will always be a number (https://developers.arcgis.com/javascript/3/jsapi/layerinfo-amd.html#id).

That doesn't appear to be the case. Note the data type of `layerId` in the screenshots below.

Sample layer for the New Jersey service:

![image](https://cloud.githubusercontent.com/assets/1042475/17036815/e47fbc42-4f5b-11e6-98b3-d95b6b923c69.png)

Sample layer from the Restoration Explorer service:

![image](https://cloud.githubusercontent.com/assets/1042475/17036865/1d532086-4f5c-11e6-9ee0-303030cd0e58.png)

PR #663, opened 2016-07-29
Ensure layerId is always an int in getVisibleLayers

caseypt:
Convert layerId to an int when assembling visible layers. In most cases, it was already in int. In some cases, it was a string, which caused look-ups further down in the getVisibleLayers function to fail.

**Testing**
- Add the [restoration explorer](https://github.com/CoastalResilienceNetwork/restoration-explorer) plugin to your local development framework installation.
- Visit the site, and activate the plugin.
- Select some of options from the plugin to add layers.
- Ensure those layers show up in the legend.
- Ensure layers added in the Regional Planning plugin still get added to the legend.
- [Risk explorer](https://github.com/CoastalResilienceNetwork/explorer) also experienced the same problem, so you may want to test that it works now too.

image

Connects to #661

### DIFF
--- a/src/GeositeFramework/js/Map.js
+++ b/src/GeositeFramework/js/Map.js
@@ -232,7 +232,8 @@ require(['use!Geosite',
                     service.visibleLayers.sort(function(a, b) { return a - b; });
                     _.each(service.visibleLayers, function(layerId) {
                         var layer,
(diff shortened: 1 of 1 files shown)


mmcfarland:
+1, tested.

Issue #665, opened 2016-08-11
Legend not appearing since newer Framework (Summer 2016).


mapossum:
In the plugin Regional Resilience Project (Connecticut dev and prod http://dev.maps.coastalresilience.org/connecticut/) the legend does not appear until after some user interaction.  For example things like repeated zooming in and out, turning a layer from another plugin and activating a sub-region or even resizing the entire browser window all seem to get the legend to appear.  Is there a way for plugins to force activation of the legend?  The repo for this plugin is called "filter-select".  

The legend for this plugin did not have this issue prior to the framework upgrade and has been recently tested with an older version of the framework and it still works with that one.

PR #667, opened 2016-08-31
Redraw legend upon map resize

kdeloach:
Allow plugins to redraw the legend by calling `map.resize()`.

This is a workaround for the fact that adding custom legend HTML does
not trigger any events we listen to for redrawing the legend.

If custom legend HTML is created while the legend is minimized, it will
stay minimized.

I decided that it would be easier for plugins to trigger an event on the
`map` instance (that each plugin already has access to) instead of
updating the API to pass in the current `Legend` instance.

This was previously not an issue because the legend container remained
open at all times, even when it was empty. Plugins that rely on this old
legend behavior, and use custom legend content, will need to be updated
to call `map.resize()` each time custom legend content is updated.

**Testing**

1. Clone and checkout https://github.com/CoastalResilienceNetwork/filterSelect/pull/1 into your `plugins` folder
2. Load the `filterSelect` plugin
3. Select a project type

    image
4. Verify that the custom legend appears (without needing to pan/zoom the map)

Connects #665

### DIFF
--- a/src/GeositeFramework/js/Map.js
+++ b/src/GeositeFramework/js/Map.js
@@ -301,6 +301,8 @@ require(['use!Geosite',
         dojo.connect(esriMap, 'onLayerAdd', redraw);
         dojo.connect(esriMap, 'onLayerRemove', redraw);
         dojo.connect(esriMap, 'onLayerSuspend', redraw);
+        // Allow plugins to trigger a legend redraw by calling map.resize()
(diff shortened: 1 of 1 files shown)


caseypt:
Asking because I'm not sure: should this target `develop` or `master`?

kdeloach:
Oh, that was a mistake. Let me open a new PR.

Issue #660, opened 2016-07-08
Plugins which fail to load shouldn't break framework


mmcfarland:
When a plugin throws an exception when starting up, the error is often lost and the framework typically only renders the header bar, leaving a white element underneath.  Ideally, in that error case the behavior would be:
* Error is logged to server and/or console
* Offending plugin parsing/loading is abandoned
* All other plugins are loaded normally

PR #673, opened 2016-09-16
Improve handling of bad plugins

caseypt:
To prevent plugins with syntax errors from stopping the framework from loading, we also now capture errors when creating the plugin objects. This is before the plugins are initialized, where we were already capturing errors.

**Testing**
- In any plugin, create a syntax error in the initialize function.
- Load the framework. You should see an error message like this in the console:
image
- The framework should still load.

- In that same plugin, remove the syntax error, and add a new syntax error in the meta information above the initialize function.
- Refresh the framework. You should see an error message like this in the console:
image
- The framework should still load, and the launcher for that particular plugin should be missing from the sidebar.

Connects #660

### DIFF
--- a/src/GeositeFramework/js/Pane.js
+++ b/src/GeositeFramework/js/Pane.js
@@ -104,26 +104,34 @@ require([
             var plugins = new N.collections.Plugins();
 
             _.each(N.plugins, function(PluginClass, i) {
-                var pluginObject = new PluginClass(),
-                    plugin = new N.models.Plugin({
-                        pluginObject: pluginObject,
-                        pluginSrcFolder: model.get('regionData').pluginFolderNames[i]
-                    });
+                try {
+                    var pluginObject = new PluginClass(),
+                        plugin = new N.models.Plugin({
+                            pluginObject: pluginObject,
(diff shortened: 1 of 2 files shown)


caseypt:
Re: some errors not being caught in the `declare` block of plugins (for example, when underscore is undefined [here](https://github.com/CoastalResilienceNetwork/regional-planning/blob/master/main.js#L66)), I'm having trouble figuring out a way to catch these errors.

The plugins are first referenced [here](https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/master/src/GeositeFramework/Views/Home/Index.cshtml#L685). Wrapping this section in a try/catch, which is not really what we want, doesn't catch `not defined` errors, which I don't quite understand.

There is a handler for errors thrown during `require`, (we set it up [here](https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/master/src/GeositeFramework/Views/Home/Index.cshtml#L660)), however, that seems to only catch errors thrown when one of the required modules is not found.

I'm currently looking into if there is a way to adjust the plugin architecture.

mmcfarland:
+1

PR #677, opened 2016-09-19
Show infographic on start

mapossum:
This pull request adds code and a default parameter to Plugin Framework that will change the default so that plugins do no show their infographic by default unless the user clicks the help "?" button.  Per Zachs requests this request also allows the plugin developer to override this behavior by setting showInfographicOnStart: true in the plugin code.

### DIFF
--- a/src/GeositeFramework/js/Plugin.js
+++ b/src/GeositeFramework/js/Plugin.js
@@ -236,8 +236,7 @@ require(['use!Geosite',
                 if (typeof localStorage[showValueKey] !== 'undefined') {
                     return localStorage[showValueKey] === 'true';
                 } else {
-					localStorage.setItem(showValueKey,pluginObject.showInfographicOnStart);
-					return localStorage[showValueKey] === 'true';
+					return pluginObject.showInfographicOnStart;
 				}
             },
 
--- a/src/GeositeFramework/js/Plugin.js
+++ b/src/GeositeFramework/js/Plugin.js
@@ -1,4 +1,4 @@
-﻿/*jslint nomen:true, devel:true */
+/*jslint nomen:true, devel:true */
 /*global Backbone, _, $ */
 
 // A plugin wraps around a plugin object and manages it in backbone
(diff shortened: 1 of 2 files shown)


src/GeositeFramework/js/Plugin.js
```diff
@@ -235,8 +235,10 @@ require(['use!Geosite',
                     showValueKey = pluginObject.toolbarName + " showinfographic";
                 if (typeof localStorage[showValueKey] !== 'undefined') {
                     return localStorage[showValueKey] === 'true';
-                }
-                return true;
+                } else {
+					localStorage.setItem(showValueKey,pluginObject.showInfographicOnStart);
+					return localStorage[showValueKey] === 'true';
+				}
```
kdeloach: This can be simplified to:

```js
return pluginObject.showInfographicOnStart;
```

mapossum: Ok sounds good.  You mean:

return localStorage[showValueKey] === 'true';

Can be replaced with:

return pluginObject.showInfographicOnStart;

We still need set the localStorage variable though.

Do I make the change?

Thanks

caseypt: Hi @mapossum. I believe @kdeloach is suggesting that

```js
localStorage.setItem(showValueKey,pluginObject.showInfographicOnStart);
return localStorage[showValueKey] === 'true';
```

can be replaced with 

```
return pluginObject.showInfographicOnStart;
```

I tested out this PR with and without @kdeloach's suggested change, and they are functionally similar. I think setting the local storage variable has no functional effect because it is only used immediately after it is set from the value in the plugin config. Returning the value from the plugin config is a more direct way to achieve the desired change.

caseypt: I tested out @kdeloach's suggestion a few times, and here is the behavior I'm seeing.

If the user unchecks the "Don't show this on start" option after clicking the "?" button on a plugin, the local storage value is set and the infographic opens the next time the plugin is activated. If the user then rechecks the option in the infographic, the local storage value is updated, and the infographic isn't shown again until the "?" is clicked.

If the user never interacts with the "Don't show this on start" option, the local storage option is never set, and the plugin override value is used.

The change @kdeloach suggested also has the side benefit of not overwriting the user's preferred settings. If they have previously hid the infographic, that setting will not be overridden by the plugin setting.

caseypt: That's all to say, keeping this as is works just the same as the suggested change, so it's up to you if you want to edit it before merging.

mapossum:
Ok I made the change and update and created a revised pull request.

kdeloach:
+1

Issue #692, opened 2016-10-17
Implement installation drop-down menu


caseypt:
- This is currently hard-coded in the prototype. How should this list be configured?

![image](https://cloud.githubusercontent.com/assets/1042475/19445343/dd2837bc-9461-11e6-856d-90724117c01c.png)

mmcfarland:
This is currently configured as a `headerLink`, we should probably reuse the parsing and config, but promote it from a generic link to a specific region list link.

https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/master/src/GeositeFramework/region.json#L14-L20

caseypt:
I think `headerLink` now represents this:

![image](https://cloud.githubusercontent.com/assets/1042475/19445974/a76d5938-9464-11e6-8162-548c66cdebf8.png)

So maybe add a `regionLink` list?

mmcfarland:
Sorry, I meant that's how it's configured currently in the regions: https://github.com/CoastalResilienceNetwork/gulfmex-region/blob/master/region.json#L9-L21

I think we should pull it out from the generic `headerLink` to a specific `regionLink` as you suggest.

caseypt:
Ah got it. Didn't notice it in the header before. Thanks.

PR #697, opened 2016-10-18
Configure installation dropdown menu

kellyi:
This PR sets up the region installation dropdown menu to configure dynamically from the links nested beneath `region.json`'s `"GO TO"` item, which is how those links seem to be initially configured in the regional sites:

screen shot 2016-10-18 at 11 45 08 am
 
& ->

screen shot 2016-10-18 at 11 45 25 am

The changes here update the parsing in `Geosite.cs` to create a `RegionLinks` component in addition to the existing `HeaderLinks` component, by filtering the objects based on whether or not they've got the name `"GO TO"`. From there, the list of `RegionLinks` gets passed through a `RenderRegionLinks` helper method which creates the links in the dropdown. 

**Testing**
- Get this branch, build the solution in Visual Studio, then view it in the web browser
- Confirm that you see the new dropdown menu with clickable links ordered in a way that matches the `region.json` file
- Confirm that the "Geosite Framework Sample" dropdown includes its own list of links and does not include any of the links under "Sample Region"

Connects #692

### DIFF
--- a/src/GeositeFramework/Models/Geosite.cs
+++ b/src/GeositeFramework/Models/Geosite.cs
@@ -53,6 +53,7 @@ namespace GeositeFramework.Models
         public Link TitleMain { get; private set; }
         public Link TitleDetail { get; private set; }
         public List<Link> HeaderLinks { get; private set; }
+        public List<Link> RegionLinks { get; private set; }
         public string RegionDataJson { get; private set; }
         public List<string> PluginFolderNames { get; private set; }
         public string PluginModuleIdentifiers { get; private set; }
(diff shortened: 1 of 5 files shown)


src/GeositeFramework/Models/Geosite.cs
```diff
@@ -152,7 +153,14 @@ private static List<JsonData> GetPluginConfigurationData(IEnumerable<string> plu
             if (jsonObj["headerLinks"] != null)
             {
                 HeaderLinks = jsonObj["headerLinks"]
-                    .Select(ExtractLinkFromJson).ToList();
+                    .Where(element => element["text"].Value<String>() != "GO TO")
```
mmcfarland: This seems a little too lose, I think that we should update the region schema to allow for an explicit type of `regionList` and not infer it from the label text.

kellyi: Ah okay -- happy to do that! I didn't know whether we needed to stick with the existing schema, but I'll modify it and make the corresponding change here.

kellyi:
Pushed a fixup which updates the schema, then uses the new schema to configure the installations dropdown.

caseypt:
@jfrankl Should the links in this dropdown open in the same window or new tabs?

kellyi:

kellyi:
I squashed the fixup commit and revised the commit message to note the changes to the schema. Going to merge.

Issue #694, opened 2016-10-18
Implement plugin layer hiding functionality


caseypt:
Clicking the eye icon should hide any layers added by the plugin.

![image](https://cloud.githubusercontent.com/assets/1042475/19477411/01c77e8e-950c-11e6-829c-594947c66d8f.png)

kellyi:
Per @jfrankl this button should toggle the layers off and make the corresponding changes to the legend, but it should leave the plugin open. That means it's not really either `.hibernate` or `.close`, exactly, so I'll rework what I did in #702 to handle toggling.

PR #702, opened 2016-10-18
Call plugin `hibernate` method on eye icon click

kellyi:
This PR wires up the eye-icon button in the plugin navbar so that it will call the plugin's `hibernate` method. I've tested it against the master branch of `regional-planning`, and it calling that method will clear the layers the plugin has added while keeping the plugin's sidebar open.

The second commit here removes some drag-to-move functionality which was added to the sidebar in the prior interface for the app. In the `prototype` branch, that feature had the effect of banishing the sidebar out of the viewport altogether, and it would interfere with clicking the navbar header buttons in the case that a click on one of the icons happened in tandem with some motion of the mouse or touchpad.

**Testing**
- get this branch, install the `regional-planning` plugin, then build the app in Visual Studio and view it in Chrome
- click the regional planning plugin to bring up the sidebar
- add some of the Alabama layers to see them highlighted in the plugin, the legend, and on the map (other layers will work, too, but those are directly in the viewport)
- click the eye icon to clear all the layers, and verify that they're removed from the map, the legend, and the layer selector
- repeat a few times to verify that it doesn't end up in a non-working state
- try to drag the header to move the sidebar and verify that it won't move
- try dragging the legend to confirm that it's still movable 

Connects #694

### DIFF
--- a/src/GeositeFramework/js/Plugin.js
+++ b/src/GeositeFramework/js/Plugin.js
@@ -8,7 +8,6 @@ require(['use!Geosite',
          'esri/layers/ArcGISDynamicMapServiceLayer',
          'framework/Logger',
          'dojo/dom-style',
-         'framework/widgets/ConstrainedMoveable',
          'dojox/layout/ResizeHandle',
          'dijit/form/CheckBox',
          'dijit/form/Button'
@@ -18,7 +17,6 @@ require(['use!Geosite',
              ArcGISDynamicMapServiceLayer,
              Logger,
              domStyle,
-             ConstrainedMoveable,
              ResizeHandle,
              CheckBox,
              Button) {
@@ -469,11 +467,6 @@ require(['use!Geosite',
             view.$el.parents().find('.content .nav-apps').after($uiContainer);
 
             setResizable(view, pluginObject.resizable);
-
-            new ConstrainedMoveable($uiContainer[0], {
-                handle: $uiContainer.find('.sidebar-nav')[0],
-                within: true
-            });
 
             // Tell the model about $uiContainer so it can pass it to the plugin object
             model.set('$uiContainer', $uiContainer);
--- a/src/GeositeFramework/js/Plugin.js
+++ b/src/GeositeFramework/js/Plugin.js
@@ -163,6 +163,10 @@ require(['use!Geosite',
                 }
             },
 
+            hibernate: function () {
+                this.get('pluginObject').hibernate();
+            },
+
             turnOff: function (callback) {
                 var self = this,
                     initiallySelected = this.selected,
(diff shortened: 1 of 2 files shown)


src/GeositeFramework/js/Plugin.js
```diff
@@ -401,6 +405,10 @@ require(['use!Geosite',
             view.$uiContainer = $uiContainer;
 
             $uiContainer
+                // Call `hibernate` method on plugin when the eye icon is clicked
+                .find('.plugin-eye').on('click', function () {
+                    model.hibernate();
+                }).end()
```
kdeloach: I would check to make sure this isn't already satisfied by https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/c3d2e103da69463456dbaf738bb0e86ec86fc657/src/GeositeFramework/js/Plugin.js#L417.

kdeloach: At least, this is the code path executed on the `develop` branch when the minimize button is pressed.

src/GeositeFramework/js/Plugin.js
```diff
@@ -470,11 +470,6 @@ require(['use!Geosite',
 
             setResizable(view, pluginObject.resizable);
 
-            new ConstrainedMoveable($uiContainer[0], {
-                handle: $uiContainer.find('.sidebar-nav')[0],
-                within: true
-            });
-
```
kdeloach: Consider removing `ConstrainedMoveable` from the define statement as well, if it's not used anywhere else in this file.

src/GeositeFramework/js/Plugin.js
kdeloach: This change looks good, but considering that the existing framework (on `develop`) already has a "minimize" button which causes the plugin to hibernate, I'm wondering if this is necessary. It's possible that some code became orphaned while implementing the prototype. Or maybe the "minimize" code was deleted?

kellyi: It looks like the `.deselect` method called in the `.plugin-close` case also removes the sidebar rather than just having it remove the layers. If we do want it the eye-icon button to close the sidebar too, I can have it use `.deselect` method after `.hibernate`.

I think either way I'll cut the `.plugin-close` case, as that's no longer a button used in the app.

kdeloach: Sounds like a regression defect then. For example, if you add some layers using the Map Layers plugin in [production](http://maps.coastalresilience.org/gulfmex/), then minimize the plugin, the layers remain on the map.

Basically, we just need to keep the existing behavior, but replace the minimize icon with an eye ball.

Just pointing this out to ensure that we keep things as backwards-compatible as possible for existing plugins.

kellyi:
I'm going to close this PR and open a new one about toggling the layers on and off per specification in #694

src/GeositeFramework/js/Plugin.js
kdeloach: Disregard all this. Misunderstood the requirements.

Issue #704, opened 2016-10-19
Implement TNC UI feedback


ajrobbins:
- Some regions may not want the region name dropdown to facilitate switching away from their region so easily. We should add a configuration option in which there is no region dropdown (need to figure out where other regions will live in header/navigation), and potentially one in which the dropdown has subregions.
- Plugin menu should collapse when plugin is selected, like it is on small screens/tablets. The active plugin icons should stay highlighted.
- Need to show which plugins are active - icon a different color? Vs which plugin is being shown in the sidebar - background highlighted?
- In the responsive design for tablets, the plugin menu is collapsed so only the icons are visible. The plugin name is shown in a tooltip on hover. How will this work with touch?
- ~~How to show that a plugin is activated but not visible (eye icon) in the plugin menu - currently, there's no way to see the eye icon/visibility status from the plugin menu~~
- ~~Need to figure out how the eye icon works - currently, when layers are turned on, it remains grey. It turns blue on hover, but when clicked it removes the layers from the map. This seems confusing. Need a way to have the eye icon show that layers are on the map, and figure out visual signal that it's showing layers off vs on~~

kellyi:
eye-close icon to display when the layers are toggled off: 

http://fontawesome.io/3.2.1/icon/eye-close/

otherwise, show the eye-open icon

caseypt:
> Some regions may not want the region name dropdown to facilitate switching away from their region so easily. We should add a configuration option in which there is no region dropdown (need to figure out where other regions will live in header/navigation), and potentially one in which the dropdown has subregions.

The region dropdown can be removed by not including the `regionLinks` object in `region.json`.

mmcfarland:
> The region dropdown can be removed by not including the regionLinks object in region.json

I think the better behavior would be to still display the region label, just without the dropdown.

mmcfarland:
These tasks have been broken out to other issues or completed.

PR #705, opened 2016-10-20
Toggle plugin layers on eye-icon button click

kellyi:
This PR wires up the eye-icon button to show and hide a plugin's layers on clicking the eye-icon button. Per #694, the desired behavior for this button is that it toggles a plugin's layers on and off on the map and in the legend on click, but does not otherwise reset the state of the plugin. 

To handle this, I added a new `toggleLayers` method to `plugins.js`, plus some variables to store the state of the toggling and the layers to return to the map when they're toggled back on. The result currently works like this:

*After adding layers from ECA and Regional Planning plugins:*

screen shot 2016-10-20 at 10 08 04 am

*After clicking the eye-icon button in the Regional Planning plugin:*

screen shot 2016-10-20 at 10 08 20 am

*After clicking the Regional Planning eye-icon button again:*

screen shot 2016-10-20 at 10 08 31 am

The second commit here removes a drag event listener from the sidebar plugins which would cause them to be banished out of the viewport if a user dragged while clicking. This functionality's no longer necessary since the sidebar plugins now occupy a fixed spot in the viewport when selected.

**Testing**
- get this branch, install the ECA and Regional Planning plugins, then build the project in Visual Studio
- view the project in Google Chrome and open the browser console
- toggle some layers on in ECA and some in Regional Planning
- click the eye-icon in Regional Planning and verify that its layers have been removed from the map and the legend but that the state of the plugin itself has not been reset and that the layers from ECA remain on the map and in the legend
- click the eye-icon button again to toggle its layers back on and confirm that they show in the map and legend
- repeat these steps with ECA
- try some additional combos of toggling on and off between plugins and verify that the app doesn't end up in a non-working state
- try to click and drag the header bar for the plugins and verify that nothing happens

Connects #694

### DIFF
--- a/src/GeositeFramework/js/Plugin.js
+++ b/src/GeositeFramework/js/Plugin.js
@@ -8,7 +8,6 @@ require(['use!Geosite',
          'esri/layers/ArcGISDynamicMapServiceLayer',
          'framework/Logger',
          'dojo/dom-style',
-         'framework/widgets/ConstrainedMoveable',
          'dojox/layout/ResizeHandle',
          'dijit/form/CheckBox',
          'dijit/form/Button'
@@ -18,7 +17,6 @@ require(['use!Geosite',
              ArcGISDynamicMapServiceLayer,
              Logger,
              domStyle,
-             ConstrainedMoveable,
              ResizeHandle,
              CheckBox,
              Button) {
@@ -482,11 +480,6 @@ require(['use!Geosite',
             view.$el.parents().find('.content .nav-apps').after($uiContainer);
 
             setResizable(view, pluginObject.resizable);
-
-            new ConstrainedMoveable($uiContainer[0], {
-                handle: $uiContainer.find('.sidebar-nav')[0],
-                within: true
-            });
 
             // Tell the model about $uiContainer so it can pass it to the plugin object
             model.set('$uiContainer', $uiContainer);
--- a/src/GeositeFramework/js/Plugin.js
+++ b/src/GeositeFramework/js/Plugin.js
@@ -130,7 +130,9 @@ require(['use!Geosite',
             defaults: {
                 pluginObject: null,
                 active: false,
-                displayHelp: false
+                displayHelp: false,
+                pluginLayers: {},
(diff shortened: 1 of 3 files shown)


caseypt:
Toggling works well, however, the eye icon stays highlighted after the first click.

image

image

src/GeositeFramework/js/Plugin.js
```diff
@@ -163,6 +165,21 @@ require(['use!Geosite',
                 }
             },
 
+            toggleLayers: function () {
+                var currentPlugin = this.get('pluginObject'),
+                    pluginLayersVisible = this.get('pluginLayersVisible');
+
+                if (pluginLayersVisible) {
+                    this.set('currentPluginLayers', currentPlugin.map.getMyLayers());
```
caseypt: `currentPluginLayers` is set here, but `pluginLayers` is initialized above.

kellyi: Good catch. I'll update this to use `pluginLayers`.

kellyi:
re: the highlighting: the blue highlight currently happens on the icon because it's the style given when the button's selected. I wasn't sure what other updates to make to the UI to handle making it known that the layers were toggled on and off, but one possibility is that the method could add and remove classes to that button on toggles.

caseypt:
@kellyi Yeah, I think adding and removing classes is the way to go.

kellyi:
Updated to change the icon button's color when the layers are toggled on and off:

*On*:
screen shot 2016-10-20 at 11 49 15 am

*Off*:
screen shot 2016-10-20 at 11 49 08 am

Additional test for this change would be to verify that the eye icon color updates properly in tandem with the state of the toggling and that toggling the layers off in one plugin (thereby changing the icon color) does not change the color of the icon in another plugin.

kellyi:
Spoke with @ajrobbins and I'm going to cut the color change stuff altogether for now as she made #704 to handle creating a visual signal that the layers from a plugin have been toggled on and off.

caseypt:
Sounds good.

Issue #688, opened 2016-10-17
Add ellipse map control


caseypt:
- Add ellipse map control
- Move measure and zoom to extent to ellipse control
- Convert print and save and share to map controls.

![image](https://cloud.githubusercontent.com/assets/1042475/19445124/0a8b56f4-9461-11e6-98f3-e0b49d837492.png)

kdeloach:
Leaving 2 points on this from last sprint to complete #708 (the changes were more involved than I anticipated).

PR #708, opened 2016-10-20
Create map_utils plugin

kdeloach:
#### Overview

This plugin displays a popup menu to allow users to select a variety of
map actions from a single location.

Two existing map plugins have been merged into map_utils (measure & full_extent). The other two commands for "save & share" and "export page" use the app dispatcher to launch the appropriate modal dialogs.

#### Demo

screenshot 2016-10-24 at 14 58 15

#### Notes

The "measure" command styling is messed up due to the prototype styles. We should create a new issue to fix this.

#### Test

Verify that the plugin works as expected. 

Connects #688


kdeloach:
I think this is ready for review now.

caseypt:
Consider updating the VS solution file with the new plugin file paths.

src/GeositeFramework/css/app.css
```diff
@@ -262,8 +262,8 @@ input.code-like, textarea.code-like {
             }
         .content .map .tools {
             position: absolute;
-            top: 92px;
-            left: 20px;
+            top: 75px;
+            right: 20px;
             z-index: 1;
```
caseypt: I'm still seeing the zoom buttons on the left. Should they have been moved as well?

kdeloach: I didn't move those because they are part of the ESRI map not plugins, but I guess it does make sense to move them too.

src/GeositeFramework/plugins/map_utils/export/main.js
```diff
@@ -0,0 +1,20 @@
...
+        constructor: function(args) {
+            declare.safeMixin(this, args);
+        },
+
+        execute: function() {
+            this.app.dispatcher.trigger('export-map:pane-0');
+        },
+
+        cancel: function() {
+            // NOP
```
caseypt: Typo

src/GeositeFramework/plugins/map_utils/share/main.js
```diff
@@ -0,0 +1,20 @@
...
+        constructor: function(args) {
+            declare.safeMixin(this, args);
+        },
+
+        execute: function() {
+            this.app.dispatcher.trigger('save-share');
+        },
+
+        cancel: function() {
+            // NOP
```
caseypt: Same typo here as well.

src/GeositeFramework/plugins/map_utils/export/main.js
kdeloach: It's a stand-in for "no operation" but maybe I'll just delete these placeholder functions instead.

caseypt: Ah. I thought it was supposed to be `// NOOP`.

src/GeositeFramework/css/app.css
kdeloach: I was able to position the zoom controls to the top-right. However, I noticed that resizing the page will cause the zoom plugins to move independently from the rest of our plugins.

image

It "snaps" back into the correct position once you are done resizing the page. To avoid this, we might consider leaving the zoom controls in the top-left position.

kdeloach:
Rebased, updated VS project file, and moved zoom controls to top right.

PR #709, opened 2016-10-20
Hide region menu if no regions are specified

caseypt:
If no region links were specified in the site's region.json, don't show
the region selection menu. This also allows the site to load if the region
links setting is not included in region.json.

**Testing**
- Load the site. The region selection menu should be in the header.
- Remove the region links section from `region.json`, and reload the site.
- The region selection menu should be gone.

Connects to #692

### DIFF
--- a/src/GeositeFramework/Views/Home/Index.cshtml
+++ b/src/GeositeFramework/Views/Home/Index.cshtml
@@ -59,10 +59,12 @@
 }
 
 @helper RenderRegionLinks(IList<Geosite.Link> links) {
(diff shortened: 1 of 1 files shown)


kellyi:
With `regionLinks` present:

screen shot 2016-10-20 at 1 49 34 pm

Without `regionLinks`:

screen shot 2016-10-20 at 1 50 41 pm

kellyi:
+1

caseypt:
Thanks

Issue #700, opened 2016-10-18
Fix map plugin activation


caseypt:
Clicking on this plugin doesn't appear to be activating them.

ajrobbins:
Closing in favor of https://github.com/CoastalResilienceNetwork/GeositeFramework/issues/688

kdeloach:
I'd like to reopen this issue because it's still happening for me and it doesn't seem caused by #688 (nor does it seem within the scope of #688 to fix).

PR #712, opened 2016-10-21
Fix map plugin activation

caseypt:
Use a less specific selector for plugin launcher event binding. The
sidebar plugins use button elements now, but the topbar/map plugins use
divs.

image

**Testing**
- Try activating all of the map plugins. They should now work.
- Try activating some sidebar plugins. They should still work.

Connects to #700


kdeloach:
+1 Tested

Issue #699, opened 2016-10-18
Fix printing from a plugin


caseypt:
- The plugin print workflow is complicated, and is currently not working on the `prototype` branch. Sometimes when you click the print button of the regional planning plugin, nothing happens~~, and sometimes the plugin window disappears~~ (fixed in #705). I spent some time debugging this but I couldn't figure out what's wrong.
- To debug this, it might be helpful to open up the framework on the `develop` branch and compare behavior to the `prototype` branch.

Issue #687, opened 2016-10-17
Fix launchpad


caseypt:
- The launchpad is currently blank and can't be closed.

mmcfarland:
There's actually a separate design task and milestone for reimplementing the launchpad so not worth the effort to necessarily fix the display on this branch.

PR #713, opened 2016-10-21
Remove Bootstrap popover CSS rules

caseypt:
These rules conflict with the Tiny Box rules, the modal library we are
currently using.

**Testing**
- When starting the app, the launchpad should load, and it should be displayed properly.
- Open a plugin that has a print functionality. Click the print icon. The export map dialog should appear and it should render correctly.

Connects to #699
Connects to #687


mmcfarland:
On IdentifyPoint, the printer icon is delayed in rendering for some reason, and once it loads, the print preview map is larger than the container:

screenshot from 2016-10-21 15 23 17

caseypt:
As discussed, the identify point print icon is intentionally delayed.

Fixed the print preview map in 7482a4b.

image

Issue #701, opened 2016-10-18
Move search control to header


caseypt:

kellyi:
I hacked around with this by moving the DOM elements around in the HTML source of the page and adjusting some CSS rules and:

<img width="1197" alt="screen shot 2016-10-19 at 5 32 17 pm" src="https://cloud.githubusercontent.com/assets/4165523/19538462/200dca94-9622-11e6-913c-b2c6ba011e1d.png">

This control's currently implemented as a plugin. Do we have any existing plugins which occupy the navbar like this? If not, would we want to make it possible to add plugins to the header (at least this one) or should this just be set up as a non-plugin feature?

kdeloach:
I think this should become integrated into the framework rather than exist as a plugin.

caseypt:
Per @mmcfarland, I'm going to replace our search control with the built in ESRI search control (demo here: https://developers.arcgis.com/javascript/3/sandbox/sandbox.html?sample=search_basic). The search control will be baked into the framework and will no longer be a plugin.

Issue #690, opened 2016-10-17
Minor UI issues


caseypt:
~~Close button on map controls (like the search) is not circular. The search control close button is hard to click as well.~~ This was replaced by a new control.

**Add styles for "Get Started" button**

Currently:

![image](https://cloud.githubusercontent.com/assets/1042475/19487222/d355edd2-952f-11e6-8c00-c667585b6b82.png)

Should be:

![image](https://cloud.githubusercontent.com/assets/1042475/19487228/d9bc03f0-952f-11e6-84f1-83186ee7f086.png)

**Plugin header doesn't adjust to different plugin window sizes**
![image](https://cloud.githubusercontent.com/assets/1042475/19566845/aec12122-96b9-11e6-9534-8d2419e0799e.png)

**Plugin container padding cuts off plugin**
![image](https://cloud.githubusercontent.com/assets/1042475/19567010/46a0e5cc-96ba-11e6-92ba-1dbe23ce9bbc.png)

**Search control overlaps with region menu on small screens**
![image](https://cloud.githubusercontent.com/assets/1042475/19664460/71d086ce-9a0e-11e6-8f7b-8d6deee24a83.png)

**Export and save and share modals relied on Foundation, and needs to be reformatted**
![image](https://cloud.githubusercontent.com/assets/1042475/19732637/5ad3fe1c-9b6f-11e6-9c49-995196546db1.png)

kdeloach:
Activating a subregion hides the "toggle subregion" map plugin. However, it leaves a large gap because we are only hiding the launcher button, instead of the entire plugin container.

![image](https://cloud.githubusercontent.com/assets/43062/19738503/b91e73ec-9b85-11e6-9168-eb50d8fb51d6.png)

mmcfarland:
`<li>` elements retained their bullet style.  This can be seen on the "overflow map tool menu" and also the plugin list before the sidebar is fully rendered.

![screenshot 2016-11-03 at 9 15 07 am](https://cloud.githubusercontent.com/assets/1014341/19967542/24412d46-a1a7-11e6-960e-47f778830f1c.png)

PR #714, opened 2016-10-24
Remove "zoom_to" plugin and add ESRI search control

caseypt:
Replaces the `zoom_to` plugin with the ESRI search control.

image

**Testing**
- Visit the app and try out the search control in the header.
- Verify all references to the `zoom_to` plugin have been removed.

**Notes**
There are some display issues on small screens. I'm going to update https://github.com/CoastalResilienceNetwork/GeositeFramework/issues/690.

Connects to #701


src/GeositeFramework/js/Map.js
```diff
@@ -7,15 +7,17 @@ require(['use!Geosite',
          'esri/map',
          'esri/layers/ArcGISTiledMapServiceLayer',
          'esri/geometry/Extent',
-         'esri/SpatialReference'
+         'esri/SpatialReference',
+         'esri/dijit/Search',
```
kdeloach: I think we have run into issues in the past with trailing commas in some versions of IE.

kdeloach:
This diff looks great.

caseypt:
> I think the input control should always be open, but we can let @jfrankl weigh in.

Yeah, I wasn't sure. It's a boolean option when initializing the control, so it's easy to update.

> We'll need to get it worked into the responsive layout. Presently, it gets smashed up with the "region dropdown". Probably best to create an issue for that, and we'll work it into the design sprint when we accumulate additional tasks.

I added this to #690.

src/GeositeFramework/js/Map.js
caseypt: Thanks

jfrankl:
To confirm what Matt said: the input should always be open.

caseypt:
Pushed a fixup commit to ensure the search control is always open.

caseypt:
Going to merge since this has been reviewed and approved.

Issue #695, opened 2016-10-18
Integrate custom colors into new style rules


caseypt:
This [section of rules](https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/master/src/GeositeFramework/Views/Home/Index.cshtml#L90-L128) should be updated to work with the new style rules.

Issue #691, opened 2016-10-17
Consolidate and clean-up css.


caseypt:
- ~~Prettify `main.css`.~~ Done in #698 
- Consolidate `main.css` and `app.css`.
- Remove unused styles.
- Remove foundation JS and CSS

PR #719, opened 2016-10-24
Implement custom colors in new design

caseypt:
The CSS rules in the index document that use the custom colors were
updated with selectors that are used in the new design.

Also, a third color was added to the region config. This color is used
to configure the color of the sidebar.

The default colors were updated to match the new design.

Current theme:
image

For demonstration purposes:
image

**Testing**
- Modify the colors in `region.json` and verify the app theme is updated accordingly.

**Note**
- The links in the title bar are currently being colored by Foundation, which will be removed in #691.
- Some color choices will result in a color scheme that doesn't work well.

Connects to #695

### DIFF
--- a/src/GeositeFramework/App_Data/regionSchema.json
+++ b/src/GeositeFramework/App_Data/regionSchema.json
@@ -107,7 +107,7 @@
             }
         },
         "basemaps": {
-            "type": "array", 
+            "type": "array",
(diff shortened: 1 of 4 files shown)


jfrankl:
Looks good to me

caseypt:
Thanks!

PR #720, opened 2016-10-26
Consolidate stylesheets, remove unused JS and CSS libraries

caseypt:
This PR consolidates the two main stylesheets used in the framework (app.css, from the old design, and main.css, from the new design). It also fixes a few other small UI issues.

**Testing**
- Launch the app. It may be easier to close Visual Studio, reopen it, and then open the solution file. Several files are removed, which sometimes causes Visual Studio to get tripped up when it tries to update.
- Browse around the app, and verify that components of the app don't look unstyled.
- Verify the the app title and subtitle are not links.
- Use the measure tool, and verify that the tooltips are readable and that the completed tooltip doesn't have a blue background image.

image

**Notes**
- The Dijit/Dojo controls used by most plugins work intermittently. I haven't really spent any time tracking down why this is, but sometimes they work, and sometimes they don't.
- The export modal style is messed up because it relied on some of the base Foundation classes. I've add this to #690.

Connects to #691


kdeloach:
Regional planning styles may need some attention:

image

kdeloach:
Thanks for fixing the measure plugin styling. I noticed that while creating the map_utils plugin but forgot to create an issue.

kdeloach:
+1

caseypt:
Working on fixing the conflict.

caseypt:
Conflicts resolved.

caseypt:
> Regional planning styles may need some attention:

Thanks. Opened https://github.com/CoastalResilienceNetwork/regional-planning/issues/59

mmcfarland:
Couple of issues I noticed:
* Regional Planning has bad tabs - is the idea to port foundation based styles to that repo?
* The new font seems to be unapplied, things are now styled with `tahoma`
* There are a number of css errors in the console, not entirely sure if they're specific to this branch:  
screenshot from 2016-10-26 15 50 08

I think the latter may be related to how I have `layer_selector` is configured on my branch.

caseypt:
> Regional Planning has bad tabs - is the idea to port foundation based styles to that repo?

I opened CoastalResilienceNetwork/regional-planning#59 to track that. I think we could import the parts of Foundation we need directly from that plugin rather than including it in the framework.

> The new font seems to be unapplied, things are now styled with `tahoma`

Thanks. I'll take a look.

> There are a number of css errors in the console, not entirely sure if they're specific to this branch

I think some of those are related to the layer_selector plugin. Are we still supporting that plugin in the new design? If so, I'll look into the issues. I'm going to look into the `measure.css` errors.

mmcfarland:
To follow up, `map-layers` works fine on the `prototype` when configured well.  Only issue I see are the font and measure.css.

caseypt:
The `measure.css` error is related to the `future_habitat` plugin requesting `measure.css` in it's previous location (`plugins/measure/measure.css` vs `plugins/map_utils/measure/style.css`). I'll open an issue on that plugin.

caseypt:
Updated with a commit to protect against EXT styles.

mmcfarland:
+1, works with `map-tree` now

caseypt:
Thanks!

Issue #717, opened 2016-10-24
Keep region name when regionLinks is not included in region.json


caseypt:
Currently, the menu and title are removed when the config block is missing.

Relevant code: https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/prototype/src/GeositeFramework/Views/Home/Index.cshtml#L560-L572

caseypt:
Also, the region name should not be a link.

caseypt:
Shares a point with https://github.com/CoastalResilienceNetwork/GeositeFramework/issues/718

Issue #706, opened 2016-10-20
Update documentation


caseypt:
- Plugin sizing options
- Region installation menu config

kellyi:
I updated "Plugin Settings" on the wiki to document the new `size` option (and change `width` and remove `height`) here: https://github.com/CoastalResilienceNetwork/GeositeFramework/wiki/Plugin-Settings

I also updated the "Region.json Settings" page to document using the `regionLinks` array to configure that dropdown here: https://github.com/CoastalResilienceNetwork/GeositeFramework/wiki/Region.json-Settings

PR #721, opened 2016-10-26
Always show framework region name

caseypt:
In the event that no region links have been added to the site's region.json file, still show the region name, but don't render the region link list.

image
*No region links provided*

**Testing**
- Load the app. Verify the region menu is present.
- Edit `region.json`, and remove the `regionLink` section.
- Kill and restart the VS debug server.
- Reload the app. The region name should still be there ("Sample Region"), but the dropdown menu, including the arrow that indicates there is a menu, should be gone.

Connects to #717

### DIFF
--- a/src/GeositeFramework/Views/Home/Index.cshtml
+++ b/src/GeositeFramework/Views/Home/Index.cshtml
@@ -554,20 +554,22 @@
                 @RenderLinks(Model.HeaderLinks)
             </ul>
         </div>
-        @if(Model.RegionLinks != null) {
         <div class="dropdown header-dropdown nav-locations">
             <button class="button dropdown-toggle nav-main-button" type="button" id="installation-menu" data-toggle="dropdown" aria-haspopup="true" aria-expanded="true">
                 <div class="nav-locations-content">
                     @Model.TitleDetail.Text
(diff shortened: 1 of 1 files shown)


kellyi:
Region name's here; dropdown menu isn't:

screen shot 2016-10-26 at 5 30 16 pm

It seems like the link for the "Sample Region" isn't opening the link, but instead seems to add and remove an `open` class to the `dropdown header-dropdown nav-locations` div:

screen shot 2016-10-26 at 5 36 18 pm

&-> 

screen shot 2016-10-26 at 5 36 29 pm



I also noticed that the `titleDetail` and `titleMain` objects for `region.json` aren't specified in the Wiki -- https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/master/src/GeositeFramework/region.json -- so I'll update them as part of #706.

caseypt:
Sorry, the region name shouldn't actually be a link. I fixed that in https://github.com/CoastalResilienceNetwork/GeositeFramework/pull/720/commits/cc4878f810c64b8648932905ad0d06eb2e730d73.

kellyi:
Ah, got it. +1 then!

caseypt:
Thanks!

Issue #718, opened 2016-10-24
Bring back the plugin close button


caseypt:
This button should replace the layer hide button (the eye icon) and should turn off a plugin.

caseypt:
Shares a point with https://github.com/CoastalResilienceNetwork/GeositeFramework/issues/717

PR #722, opened 2016-10-26
Replace the hide layers button with the close button

caseypt:
After further consideration, we thought it made more sense to bring back
the plugin close button instead of having a button that would hide active
plugin layers. I left the layer hiding code in case we need it later.

image

**Testing**
- Launch the app and open the Regional Planning plugin.
- Expand some of the layer trees and turn on some layers.
- Click the plugin close button.
- Verify that the layers are removed.
- Open the Regional Planning plugin again. It should be back to the initial state.

Connects to #718

### DIFF
--- a/src/GeositeFramework/Views/Home/Index.cshtml
+++ b/src/GeositeFramework/Views/Home/Index.cshtml
@@ -343,7 +343,7 @@
                         </button>
                     <% } %>
 
-                    <button class="button-link plugin-eye"><i class="icon-eye"></i></button>
+                    <button class="button-link plugin-off"><i class="icon-cancel"></i></button>
                 </div>
             </div>
             <div class="sidebar-content">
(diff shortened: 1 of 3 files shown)


src/GeositeFramework/Views/Home/Index.cshtml
```diff
@@ -343,7 +343,7 @@
                         </button>
                     <% } %>
 
-                    <button class="button-link plugin-eye"><i class="icon-eye"></i></button>
+                    <button class="button-link plugin-off"><i class="icon-cancel"></i></button>
```
mmcfarland: Glad to see it back.  Can we add an i18n `title` attribute about what it does?

caseypt:
Added a commit to translate the button,

kdeloach:
+1

caseypt:
Thanks!

PR #723, opened 2016-10-27
Add `underscore` import for use in map-utils

kellyi:
This PR fixes a bug whereby `underscore` wasn't available in the measure plugin. The changes here add `underscore` to the `main.js` file for the map-utils plugins, and call `use` on the import to make it available.

**Testing**
- get this branch and launch it without any additional plugins installed
- build the project in Visual Studio
- view the site in the web browser and verify that it loads as it should and that you don't see the 

```
Uncaught TypeError: Module 'underscore' is undefined or does not have a 'use' config. ...
```

error in the console.
- verify that the rest of the app works as it should


kdeloach:
+1

kellyi:
Thanks!

PR #733, opened 2016-11-15
Fix minor UI issues

jfrankl:
This pull request fixes all of the minor UI issues outlined in #690.

image

image

image
 
 Connects #690

### DIFF
--- a/src/GeositeFramework/Views/Home/Index.cshtml
+++ b/src/GeositeFramework/Views/Home/Index.cshtml
@@ -419,23 +419,27 @@
             </div>
             <div class="popover-content">
                 <div class="row">
-                    <h5 class="i18n" data-i18n="Title (Optional):">Title (Optional):</h5> <input type="text" id="export-title"/>
-                    <div class="small-4 columns">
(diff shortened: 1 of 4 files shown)


caseypt:
The app now has a horizontal scroll bar for me, and the plugin title scrolls with the page:

image

caseypt:
The close button that appeared when you activated the map utils plugin has disappeared:

image


Previously (on `prototype`):

image

caseypt:
Talked with @jfrankl re: the two above items.

Re: horizontal scroll bar, this is due to the new map print sandbox. It currently has crude styling for demonstration purposes. We'll be changing this soon, so no changes are needed here.

Re: map utils close button, we thought it was best to not have the "X" and to implement a more intuitive close workflow (clicking away or clicking the plugin again should close it). I'll open an issue.

jfrankl:
Thanks @caseypt. Anything else I should do before I merge this?

caseypt:
Nope. Looks good. +1.

kdeloach:
Re: map utils... It might be easier just to eliminate the concept of map plugins, then we would have full control over this behavior. There are no other map plugins besides the ones in core, as far as I know.

Issue #725, opened 2016-11-03
Allow print preview


mmcfarland:
Based on first steps #724, add a new area for browser print DOM to hold a fixed size map based on what will fit on an 11.5x8 page.  Since the print map size will differ from the device viewport, allow for a print preview similar to the plugin-print to allow map adjustments.  If the true map size won't fit, allow the preview to at least allow the appropriate center.

I imagine this working by allowing a clone or detach/attach of the existing map so that all the layers, zoom level and extent are preserved.

PR #735, opened 2016-11-16
Add map to print preview window & print sandbox

jtarricone:
This adds a map to the print-preview window and the final printed output. By moving the DOM element directly, all the map layers and settings remain unchanged, apart from where altered by specific style rules.

To test:
* clone this branch and bring up the app
* navigate to `http://localhost:54633`
* click on the options button on the right and then on "Export Page"
* a window should appear that resembles this -->
download1
* you should be able to move and zoom the map, and any basemap or layers from the main page should be identical --> 
download3
* add a test title in the textbox
* click "Generate Map Export" to move to the print preview
* check that the print preview has the map, title, and correct branding
* repeat for a couple combinations of basemaps and layers
* repeat but exit without clicking "Generate Map Export", and check that the map is in the main app

Connects #725

### DIFF
--- a/src/GeositeFramework/Views/Home/Index.cshtml
+++ b/src/GeositeFramework/Views/Home/Index.cshtml
@@ -91,7 +91,6 @@
     <link rel="stylesheet" href="css/TinyBox2.css">
     <link rel="stylesheet" href="css/pageguide.min.css"/>
     <link rel="stylesheet" href="css/main.css">
-    <link rel="stylesheet" href="css/print.css">
     <link rel="stylesheet" href="//maxcdn.bootstrapcdn.com/font-awesome/4.4.0/css/font-awesome.min.css">
     <link href="https://fonts.googleapis.com/css?family=Roboto:500,900" rel="stylesheet">
 
@@ -433,6 +432,10 @@
                         <label><input type="checkbox" name="export-include-legend" /> <span class="i18n" data-i18n="Include Layer Legend">Include Layer Legend</span></label>
                     </div>
                 </div>
+                <div id="export-print-preview-container">
+                    <p class="instructions i18n" data-i18n="Zoom to the view that you want printed in your map.">Zoom to the view that you want printed in your map.</p>
+                    <div id="export-print-preview-map"></div>
+                </div>
                 <div class="row">
                     <div class="left-col">
                         <div class="export-output-area">
@@ -625,7 +628,7 @@
             <img src="@Model.PrintHeaderLogo" title="Header Logo" class="logo" />
             <h1></h1>
         </div>
-        <div id="print-map-container">Map goes here</div>
+        <div id="print-map-container"></div>
     </div>
 
     <!-- LEFT CONTENT AREA -->
--- a/src/GeositeFramework/js/Export.js
+++ b/src/GeositeFramework/js/Export.js
@@ -1,12 +1,91 @@
-require(['use!Geosite',
-         'dojo/Deferred',
-         'dojo/request',
-         'framework/Logger'],
-    function(N,
-             Deferred,
-             request,
-             Logger) {
+/*global Backbone, _, $ */
+
+require(['use!Geosite'],
+    function(N) {
     "use strict";
+        function render(view) {
+            $('#plugin-print-sandbox').empty();
+            $('.plugin-print-css').remove();
+            $('.base-plugin-print-css').remove();
+            $('<link>', {
+                rel: 'stylesheet',
+                href: 'css/app-print.css',
+                'class': '.app-print-css' 
+            }).appendTo('head');
+            var html = N.app.templates['template-export-window']();
+            view.$el.empty().append(html);
+            if (view.$uiContainer) {
+                view.$uiContainer.show();
+            }
+
+            return view;
+        }
+
+        function createUiContainer(view) {
+            var model = view.model,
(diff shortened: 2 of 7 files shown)


jtarricone:
It's worth noting that the styling on the print job itself is a bit off. See the attached sample.

[Geosite Framework Sample _ Sample Region.pdf](https://github.com/CoastalResilienceNetwork/GeositeFramework/files/595676/Geosite.Framework.Sample._.Sample.Region.pdf)

caseypt:
I think we should reduce the size of the map print modal to match the size of the plugin print modal. On small screens, the modal is too tall, and the export button is cut-off.

image

src/GeositeFramework/js/Export.js
```diff
@@ -1,12 +1,121 @@
...
+                printCssClass = 'plugin-print-css',
+                oppositePaneHideCssPath = 'css/print-hide-map' +
+                    (paneNumber === 0 ? 1 : 0) + '.css';
+
+
+            // Add the plugin css
+            addCss(pluginCssPath, printCssClass);
+
+            // Hide the possible same plugin from the opposite map pane
+            addCss(oppositePaneHideCssPath, printCssClass);
```
caseypt: This is now unnecessary since there is only one map pane. The previous version of the app allowed of split-screen map mode, which this line is protecting against.

src/GeositeFramework/js/Export.js
```diff
@@ -1,12 +1,121 @@
...
+                $uiContainer = $($.trim(N.app.templates['template-export-window']({id: 'export-print-preview'})));
+
+            view.$uiContainer = $uiContainer;
+
+            var pluginCssPath = 'css/app-print.css',
+                printCssClass = 'plugin-print-css',
+                oppositePaneHideCssPath = 'css/print-hide-map' +
+                    (paneNumber === 0 ? 1 : 0) + '.css';
+
+
```
caseypt: There should be a call here to remove any plugin print stylesheets:
Something like: 

```
$('.plugin-print-css').remove();
```

src/GeositeFramework/js/Export.js
```diff
@@ -1,12 +1,121 @@
+/*global Backbone, _, $ */
+
 require(['use!Geosite',
+         'esri/map',
+         'esri/geometry/Extent',
+         'esri/layers/ArcGISDynamicMapServiceLayer',
+         'esri/layers/ArcGISTiledMapServiceLayer',
```
caseypt: These map layer classes don't appear to be used.

src/GeositeFramework/js/Export.js
```diff
@@ -1,12 +1,121 @@
...
+                        var $printSandbox = $('#map-print-sandbox'),
+                            parseDeferred = $.Deferred(),
+                            previewDeferred = $.Deferred();
+                        $('.print-sandbox-header h1').text($("#export-title").val());
+
+                        _.delay(parseDeferred.resolve, 500);
+
+                        // Move the map from the plugin print preview dialog to the sandbox where
+                        // the plugin can mess with it's positioning among its other elements
+                        //var mapReadyDeferred = this.printableMap($printSandbox, previewDeferred);
```
caseypt: Consider updating this comment to reflect the use of this code in print the main map and not a plugin. Also, can the commented out code line be removed?

src/GeositeFramework/Views/Home/Index.cshtml
```diff
@@ -90,6 +90,7 @@
 
     <link rel="stylesheet" href="css/main.css">
     <link rel="stylesheet" href="css/print.css">
+    <link rel="stylesheet" href="css/app-print.css">
```
jtarricone: I'm not sure if I know what's wrong with that screen capture, but I think that since the node is appended to a specifically named element, `export-print-preview-map`, which is explicitly named in the CSS file as the parent, then the rules can be fairly precisely applied. The reason I opted for that was that there seemed to be inconsistency with applied styles when the DOM node was moved around, or when events acted upon it.

caseypt:
Heads up that there is a conflict.

jtarricone:
After I rebased and made the additional style change to the modal, this is the result -->
download4

I think it's more expedient to just merge this in as is anyway, since there will presumably be some design work that's out of scope.

caseypt:
Taking another look.

src/GeositeFramework/js/Plugin.js
```diff
@@ -450,6 +450,9 @@ require(['use!Geosite',
                     $('.' + printCssClass).remove();
                     $printSandbox.empty();
 
+                    // add base plugin css
+                    addCss('css/print.css', 'base-plugin-print-css');
```
caseypt: `print.css` was intended to always be on the page, and not something that needs to be added dynamically. It should hold the base print rules that apply both to plugin print and map printing.

caseypt:
Looks good and works well. Just curious about the changes to `print.css`.

src/GeositeFramework/js/Plugin.js
jtarricone: I cut it because after the rules were added dynamically for showing the header, the styles went into a weird state such that the header would always show for plugins. It wouldn't happen if only the plugin output was printed, but once an export brought in the rules they stuck around, even after using jQuery to remove the class.

caseypt: Got it. Thanks for the explanation.

jtarricone:
Thanks, I'll reword the commit message and add a couple issues.

Issue #726, opened 2016-11-03
Add legend to browser print


mmcfarland:
The legend DOM should be included in the sandbox for browser print

PR #740, opened 2016-11-17
Add layer legend to printed map

jtarricone:
This adds the legend to the printed map, per the users option. The legend style is manipulated directly since it's a child of the map node that gets moved around.

To test:
* clone this branch and bring up the app
* select a layer, then click "Export"
* check the "Include Layer Legend" box and then click "Generate" -->
download
* the printed output should resemble the attached PDF-->
[Geosite Framework Sample _ Sample Region2.pdf](https://github.com/CoastalResilienceNetwork/GeositeFramework/files/598060/Geosite.Framework.Sample._.Sample.Region2.pdf)
* go through the same steps but without checking the box, there should be no legend

Connects #726

### DIFF
--- a/src/GeositeFramework/js/Export.js
+++ b/src/GeositeFramework/js/Export.js
@@ -64,11 +64,28 @@ require(['use!Geosite'],
 
                         mapReadyDeferred.then(function () {
                             $("#export-print-preview-map").detach().appendTo($("#print-map-container"));
+
+                            if ($("[name='export-include-legend']").is(":checked")) {
+                                // show & expand all legend items
+                                $("#legend-container-0").css({ visibility: "visible" });
+                                _.each(
+                                    $(".item.expand>.expand-legend"),
+                                    function (el) {
+                                        el.click();
+                                    });
+                                $(".item.extra.collapse").hide();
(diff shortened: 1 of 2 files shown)


src/GeositeFramework/js/Export.js
```diff
@@ -64,11 +64,24 @@ require(['use!Geosite'],
 
                         mapReadyDeferred.then(function () {
                             $("#export-print-preview-map").detach().appendTo($("#print-map-container"));
+
+                            if ($("[name='export-include-legend']").is(":checked")) {
+                                // show & expand all legend items
+                                $("#legend-container-0").css({ visibility: "visible" });
+                                $(".expand-legend").each(function(x) { x.click; });
```
caseypt: Minor, but a more common variable name for cases like this is `el`.

caseypt:
I took a look at the changes. The problem I reported above appears to have been fixed.

I think this was working before, but now layers with many legend items are showing up as compressed in the print preview, regardless of their state before initiating the export:

image

jtarricone:
I just pushed a fix for the collapsed items.

Issue #754, opened 2016-11-28
Ensure launchpad is always at top of plugin list


caseypt:
Our test prototype NJ spanish region shows this problem:

![image](https://cloud.githubusercontent.com/assets/1042475/20685026/93803d2c-b580-11e6-8a93-13b209e5b060.png)

Issue #750, opened 2016-11-22
Shutdown launchpad when another plugin or subregion is activated


caseypt:
Originally from https://github.com/CoastalResilienceNetwork/GeositeFramework/issues/736

> Unlike plugins, it will never be "active" when it is not displayed ...

Most plugins "hibernate" when another plugin is activated. They only "turn off" when the "X" button is closed. However, the launchpad plugin should "turn off" instead of "hibernate" when another plugin is activated.

A naive attempt at this was calling `this.plugin.turnOff` from the `launchpad.hibernate`, but this creates a max callstack exceeded error because `turnOff` calls `hibernate`.

PR #760, opened 2016-12-06
Launchpad updates

mmcfarland:
#### Overview
* Adds an icon to the launchpad plugin
* Ensures that launchpad will always be first in the list of plugins

Connects #754


mmcfarland:
Thanks.  Looks like Casey already made an issue for that https://github.com/CoastalResilienceNetwork/GeositeFramework/issues/750

Issue #747, opened 2016-11-22
Activate the launchpad by default


caseypt:
Unlike other plugins, the launchpad should be activated by default.

I attempted an implementation of this that started the launchpad from the `Plugin.js` `initialize` method. However, launching the plugin needed to be delayed, otherwise the plugin would get closed by the initialization of other plugins (I think).

If you delay the launch long enough, the launchpad activates, but this isn't a great solution:

```diff
diff --git a/src/GeositeFramework/js/Plugin.js b/src/GeositeFramework/js/Plugin.js
index 125e46b..eec1d87 100644
--- a/src/GeositeFramework/js/Plugin.js
+++ b/src/GeositeFramework/js/Plugin.js
@@ -335,6 +335,12 @@ require(['use!Geosite',
                         self.handleLaunch();
                     }
                 });
+
+                window.setTimeout(function() {
+                    if (self.model.getId()) {
+                        self.handleLaunch();
+                    }
+                }, 5000);
             },
```

We should find a better way to implement this.

PR #761, opened 2016-12-06
Initialize Launchpad by default

mmcfarland:
After all plugins are created, activate the launchpad if it has been
included and if there is no other saved state that would contradict that
activation.

Does not use a timeout based approach as described in #747.

Also fixes the region.json schema to check the correct property for launchpad, and uses the presence of a `launchpad` config key as an indication to create or not create a launchpad instance.

#### Testing
* With the `region.json` configured as is (with a launchpad), verify that the launchpad is opened on app start
* Open regional planning, select some layers and position the map - generate a save 'n share link
* Open that save 'n share link and verify that the launchpad is not visible, but the regional planning plugin is.
* Rename the `launchpad` config key so as it remove it.  Restart the site such that the region.json is re-parsed and verify that the launchpad is not visible.

Connects #747

### DIFF
--- a/src/GeositeFramework/App_Data/regionSchema.json
+++ b/src/GeositeFramework/App_Data/regionSchema.json
@@ -169,7 +169,7 @@
             "type": "boolean",
             "required": false
         },
-        "launchpads": {
+        "launchpad": {
             "type": "object",
             "required": false,
             "properties": {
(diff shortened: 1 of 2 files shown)


jtarricone:
+1 Tested.

It's worth noting that the sidebar launchpad entry looks a little weird on IE11 -->
shot-2016-12-09_14-47-06

Issue #744, opened 2016-11-17
Work out print and preview sizing area


mmcfarland:
Complete work on the print preview and the printable areas
* ~~Preview map size relating to print size~~
* Remove map print preview.  TNC would like to just use the browser print preview as the whole preview.  The center of the map should be preserved in the printed version, and screen map that is outside the container for print will will just be clipped.
* Legend and map formatting
* Landscape vs Portrait

jtarricone:
To be clear, point 2 means undoing #735 right?

mmcfarland:
Correct.  The map should be centered in the print element the same as the display element, meaning that the extent will be different for the two sizes.

PR #763, opened 2016-12-07
Feature/jmt/refine print preview

jtarricone:
This PR fixes a jQuery-related bug, refines the styling of the printed document, and updates the "print preview" form to reflect that the preview map is gone.

To test:
* clone this branch and bring up the environment
* navigate to the page
* activate some layers or markers
* export/print the map, checking that the layers are printed, and the legend is selected

Here are some sample prints:
[Geosite Framework Sample _ Sample Region_portrait.pdf](https://github.com/CoastalResilienceNetwork/GeositeFramework/files/637297/Geosite.Framework.Sample._.Sample.Region_portrait.pdf)
[Geosite Framework Sample _ Sample Region_landscape.pdf](https://github.com/CoastalResilienceNetwork/GeositeFramework/files/637298/Geosite.Framework.Sample._.Sample.Region_landscape.pdf)

Connects #744


mmcfarland:
Some inconsistent behavior I'm noticing, mostly due to the legend control being collapsed or not when the export is initiated.

### Portrait, no legend (legend expanded): Does not seem to use the whole map area.
screenshot 2016-12-07 at 1 26 17 pm
### Portrait, no legend or yes legend (legend minimized): Uses even less of the map area
screenshot 2016-12-07 at 1 27 57 pm

jtarricone:
I don't see anything like that. Is it possible your browser is using cached css files?

jtarricone:
I did notice that if the legend is minimized, and the "Include Layer Legend" option is selected, the legend won't appear when printed, since it's in a minimized state. Should the print option override the UI state?

mmcfarland:
* Yeah - if the legend is requested to be in the print, I think it should show up regardless of the state of the UI legend.
* Whether the legend is included or not, in Portrait mode, it always leaves whitespace if the UI legend is minimized.  This issue may be addressed by addressing the above point.

mmcfarland:
The center of the map is not maintained, nor is it easy to figure out how to adjust the map on the UI to get it to show what your inteneded print would be.  Here's a map with NJ centered and what the map on the print preview looks like:
#### In app:
screenshot from 2016-12-08 10 12 09

#### In print preview:
screenshot from 2016-12-08 10 14 12

The example is contrived, but is more pronounced when you're actually zoomed into some features, those are usually then not on the screen.

jtarricone:
Testing in my environment -->

download 1

PDF -->
[Geosite Framework Sample _ Sample Region_test.pdf](https://github.com/CoastalResilienceNetwork/GeositeFramework/files/639849/Geosite.Framework.Sample._.Sample.Region_test.pdf)


It looks slightly off-center, but I'm more curious why there would be such a pronounced difference. 

The white-space issue with the legend was due I think to a rebase problem, there was a CSS rule that wasn't being overwritten.

mmcfarland:
Yeah, that looks much better than mine.  I'll test again once the rebased commit is fixed.

jtarricone:
I did that test without having done any updates; I would expect both to look the same, I'm wondering why they aren't. Which browser are you using?

mmcfarland:
Chrome 55 on Windows.  I cleared the browser cache, deleted the branch, pulled the latest `prototype` and re-checked out your branch.  Here's portrait, no legend selected, with minimized legend:

screenshot from 2016-12-08 10 58 21

and PDF: 
[Geosite Framework Sample _ Sample Region.pdf](https://github.com/CoastalResilienceNetwork/GeositeFramework/files/639903/Geosite.Framework.Sample._.Sample.Region.pdf)

Let me know if you want to look at my machine.  I'm wondering if it's because you used `px` and `%` units instead of `in` for your print css.  Since we're targeting 8.5x11 output, that may be more consistent.

jtarricone:
It may have to do with the browser zoom level, that also really throws the style off.

jtarricone:
I'm going to just redo it using absolute units, I'll push an update shortly.

jtarricone:
I pushed an update to the branch that should address the issue. Changing the styles to absolute units  didn't solve the biggest problem, which I think ended up basically stemming from window size, so I added transform rules that appear to enforce consistent positions.

mmcfarland:
Map centering is working much better now, although there's some cruft at the bottom, from clipping.  I'd suggest:
* Fill the map to the border, or just bring the border up to the bottom of the map to keep it better contained
* Add CSS rules to hide the attribution in print because it's clipped.  Bonus if we can reposition that element to display on the page.

screenshot from 2016-12-08 14 21 34

mmcfarland:
You can also still get excessive bottom whitespace by doing portrait mode when there is no legend, the map should be using that space.

screenshot from 2016-12-08 14 40 07

jtarricone:
I only see that when the window is pretty small, where it looks like the map isn't big enough to fit the print area. It might be doable by resizing the element but that might be pretty delicate.

jtarricone:
I pushed a refactor that uses ESRI class methods wherever possible. It seems to work well, I tested it on different screen sizes and it was pretty consistent as far as I could tell. One thing I wondered: could the time delay required for the ESRI invocations and/or CSS loads vary from one client to the next, due to either browser or hardware?

jtarricone:
The showstopping IE11 bug appears to be fixed, but the printed output is cropped, so an IE-specific process will have to be added at some point.

mmcfarland:
Print output is looking good!  The amount of revision has left a lot of commits that I don't think have an impact on the final diff, making it a little difficult to parse.  Could you squash some of them down to reduce the noise of the various refactors?

src/GeositeFramework/js/Export.js
```diff
@@ -54,63 +68,79 @@ require(['use!Geosite'],
...
-                                // wrap all items in a div to style separately from header
-                            } else {
-                                // if the style rule is changed via jquery, that state seems to
-                                // "stick", regardless of what's in the stylesheet
-                                $("#legend-container-0").css({ visibility: "hidden" });
-                            }
-                            var pageOrientation = $("[name='export-orientation']").filter(function () { return this.checked; }).map(function () { return this.value; }).first()[0];
-                            var orientDeferred = $.Deferred();
 
+                            var pageOrientation = _.head($("[name='export-orientation']")
```
mmcfarland: This whole 4-line chain can be replaced with:
 ```js
$("[name='export-orientation']:checked").val()
```

File: src/GeositeFramework/js/Export.js [hunk 2]
```diff
@@ -54,63 +68,79 @@ require(['use!Geosite'],
...
+                                .filter(function () { return this.checked; })
+                                .map(function () { return this.value; })
+                                .first());
                             if (pageOrientation === "Landscape") {
```
mmcfarland: These two blocks for adding the orientation css link are nearly identical.  Consider calling it out into a function that generates the string change, or just templates in the variable:

```js
$('<link>', {
    rel: 'stylesheet',
    href: 'css/print-' + pageOrientation + '.css',
    'class': '.print-orientation-css',
 }).appendTo('head');
```

src/GeositeFramework/js/Export.js
```diff
@@ -54,63 +68,79 @@ require(['use!Geosite'],
...
+                                map.resize(true);
+                                map.centerAt(mapDimensions.extent.getCenter());
+                                _.delay(resizeDeferred.resolve, 500);
+
+                                // the legend items are affected by the map resize, so
+                                // they are manipulated just before printing
+                                $.when(resizeDeferred).then(function() {
+                                    // expand legend container if minimized
+                                    if ($("#legend-container-0").hasClass("minimized")) {
+                                        _.head($(".legend-close")).click();
```
mmcfarland: The use of `_.head` here and below is superfolous, jQuery operates on collections natively.

src/GeositeFramework/js/Export.js
```diff
@@ -54,63 +68,79 @@ require(['use!Geosite'],
...
+                                        postPrintAction = function() {
+                                            _.head($(".legend-close")).click();
+                                        };
+                                    }
+
+                                    if ($("[name='export-include-legend']").is(":checked")) {
+                                        // show & expand all legend items
+                                        $("#legend-container-0").css({ visibility: "visible" });
+                                        _.each(
```
mmcfarland: Here, too - jQuery will execute on collections be default.  This `_.each` can be replaced with:

```js
$(".item.expand>.expand-legend").click()
```

src/GeositeFramework/js/Export.js
```diff
@@ -54,63 +68,79 @@ require(['use!Geosite'],
...
                                     'class': '.print-orientation-css',
                                 }).appendTo('head');
-                                _.delay(orientDeferred.resolve, 750);
+                                _.delay(orientDeferred.resolve, 500);
                             }
 
                             $.when(orientDeferred).then(function () {
-                                window.print();
-                                previewDeferred.resolve();
+                                map.width = parseFloat($("#export-print-preview-map").css("width"));
```
mmcfarland: Instead of re-selecting `$("#export-print-preview-map")` throughout this code, save the reference to a variable which can be re-used without further DOM querying.  I think it'll also help out in readability as well.

src/GeositeFramework/js/Export.js
```diff
@@ -124,13 +154,31 @@ require(['use!Geosite'],
...
-                closejs: function() {
-                    var mapNode = $("#map-0").detach()[0];
-                    mapNodeParent.append(mapNode);
-                    $("#export-print-preview-map").detach().appendTo($("#export-print-preview-container"));
+                closejs: function () {
                     $('.app-print-css').remove();
                     $('.print-orientation-css').remove();
+
                     $("#legend-container-0").css({ visibility: "visible" });
+                    _.delay(restoreCssDeferred.resolve, 500);
```
mmcfarland: It's unfortunate that we have to rely so heavily on `delay` in order to have things updated in time, but the use of `Deferreds` makes it fairly understandable.  I assume the result of not using it is that the css would not be applied and rendered by the time the print preview is displayed?  In both the close and open functions, do the delays need to be incremental, or can they be accumulated into a single delay (ie, all the properties are set, css loaded, etc - then a delay - then the print call)?

src/GeositeFramework/js/Export.js
```diff
@@ -34,12 +34,26 @@ require(['use!Geosite'],
...
                 $printPreview = $('#print-preview-sandbox'),
                 mapReadyDeferred = $.Deferred(),
-                mapNodeParent = $("#map-0")[0].parentElement;
+                previewDeferred = $.Deferred(),
+                orientDeferred = $.Deferred(),
+                resizeDeferred = $.Deferred(),
+                restoreMapDeferred = $.Deferred(),
+                restoreCssDeferred = $.Deferred(),
+                restoreNodeDeferred = $.Deferred(),
+                postPrintAction = function() {},
```
mmcfarland: Using `_.noop` here may be a little more descriptive of your intention.

src/GeositeFramework/js/Export.js
```diff
@@ -54,63 +68,79 @@ require(['use!Geosite'],
                 maskopacity: 40,
                 openjs: function () {
```
mmcfarland: Both the `openjs` and `closejs` functions are long and complex enough that I think they warrant being broken out into their own named functions - and rescoping some of the above variables to just the functions where they're needed..  This would also reduce the level of indentation in the nested conditional.

jtarricone:
I think i addressed all the points of review.

src/GeositeFramework/js/Export.js
```diff
@@ -34,71 +34,162 @@ require(['use!Geosite'],
...
+                    $.when(orientDeferred).then(function () {
+                        context.map.width = parseFloat(exportMap.css("width"));
+                        context.map.height = parseFloat(exportMap.css("height"));
+                        context.map.resize(true);
+                        context.map.centerAt(context.mapDimensions.extent.getCenter());
+                        _.delay(resizeDeferred.resolve, 500);
+
+                        // the legend items are affected by the map resize, so
+                        // they are manipulated just before printing
+                        $.when(resizeDeferred).then(function() {
```
kdeloach: Does each deferred action need to be nested inside the previous action?

src/GeositeFramework/js/Export.js
```diff
@@ -34,71 +34,162 @@ require(['use!Geosite'],
...
-                                $("#legend-container-0").css({ visibility: "hidden" });
+                                context.legend.css({ visibility: "hidden" });
                             }
+                            // wrap all legend items in a div to style separately from header
+                            $(".legend-layer").each(
+                                function(el) {
+                                    $(this)
+                                        .children(".item")
+                                        .wrapAll("<div class='print-legend-coll'></div>");
+                            });
```
kdeloach: Do we need to adjust this markup dynamically? Could we update the legend template instead to avoid this?

jtarricone: There may be a way to do the same by modifying the template if the rendering of legend items is open to modification. As long as the styles only come in attached to temporary stylesheets, hopefully the elements would revert back.

src/GeositeFramework/js/Export.js
jtarricone: Typically they do, because steps have to occur in a specific sequence.

src/GeositeFramework/js/Export.js
kdeloach: I guess I don't really see the point to wrapping these. You should be able to override the legend item styles in the print stylesheet. It seems a little redundant to wrap each legend item div with another div.

kdeloach:
The legend header and minimize button are visible for me:

image

jtarricone:
>The legend header and minimize button are visible for me:

I can fix that as part of #770, along with any other minor things that may come up.

src/GeositeFramework/js/Export.js
jtarricone: I misunderstood what you meant by "nested." I think this could be updated to move totally separate parts out of blocks, I'll take a stab at it and push an update if it's a quick fix.

kdeloach: Whoops, I could have worded that better.

jtarricone:
I pushed a small update that reduces the unneeded indents in `setupExport`.

src/GeositeFramework/js/Export.js
jtarricone: The point is to manipulate the legend items separately from everything else in the legend. In the `legend-layer` element the `item`s aren't grouped, so achieving something similar to grouping them and applying `display:flex` to the container would be onerous.

jtarricone:
Is there any chance this can be approved and merged today? I have other changes that are lining up behind this.

Issue #745, opened 2016-11-21
Rename "export page" to "Create Map"


ajrobbins:
Currently, the map print function is called "Export page". We should rename it in the map functions overflow list and in the print modal:

![screen shot 2016-11-21 at 12 38 51 pm](https://cloud.githubusercontent.com/assets/12868342/20493626/a1c27510-afe7-11e6-9da3-1b689c87e26b.png)

![screen shot 2016-11-21 at 12 38 57 pm](https://cloud.githubusercontent.com/assets/12868342/20493630/a4bbe4cc-afe7-11e6-92c0-d783019e1f31.png)

PR #764, opened 2016-12-07
Replace print function name with "Create Map"

jtarricone:
This replaces all instances of "Export Page" and "Map Export" with "Create Map", on the overflow list on the main page, and on the print modal.

To test:
* clone this branch, bring up the environment & open the app
* check anywhere the old text would be, ensure that it has been replaced with "Create Map"
* check that no functionality has been affected, the change should be superficial

Connects #745

### DIFF
--- a/src/GeositeFramework/locales/en.json
+++ b/src/GeositeFramework/locales/en.json
@@ -15,7 +15,7 @@
     "Force maps to show the same extent": "Force maps to show the same extent",
     "Link Maps": "Link Maps",
     "Receive a printable copy of the current map": "Receive a printable copy of the current map",
-    "Export Page": "Export Page",
+    "Create Map": "Create Map",
     "Print Contents": "Print Contents",
     "View the info-graphic": "View the info-graphic",
     "Title (Optional):": "Title (Optional):",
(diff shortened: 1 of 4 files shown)


src/GeositeFramework/locales/en.json
```diff
@@ -15,7 +15,7 @@
     "Force maps to show the same extent": "Force maps to show the same extent",
     "Link Maps": "Link Maps",
     "Receive a printable copy of the current map": "Receive a printable copy of the current map",
-    "Export Page": "Export Page",
+    "Create Map": "Create Map",
```
kdeloach: The corresponding values in `es.json` file should be updated as well.

jtarricone: I'm guessing it's this one? -->
https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/prototype/src/GeositeFramework/locales/es.json#L45

kdeloach: They keys should be the same between both files (although it's possible for keys to be missing if we forgot to add a translated version).

https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/prototype/src/GeositeFramework/locales/es.json#L18

Issue #767, opened 2016-12-08
Export fix ups


mmcfarland:
* The export title should have a longer input box
* Title on print should be centered in the header area
* Include Legend should be checked by default
* Test and fixup other browsers (IE, specifically)

![screenshot from 2016-12-08 14 27 10](https://cloud.githubusercontent.com/assets/1014341/21024412/6b4ced4e-bd52-11e6-86d5-a2c85694c6a5.png)

![screenshot from 2016-12-08 14 27 38](https://cloud.githubusercontent.com/assets/1014341/21024436/7c65e0ea-bd52-11e6-9c28-b9a06ab7b4d8.png)

PR #770, opened 2016-12-13
Map Export UI fixes & browser-specific print process

jtarricone:
This fixes some UI issues that are part of the map export process, and addresses differences in print process between browsers.

To test:
* clone this branch, bring up the environment, and open the app
* select some combination of layers so that the legend pops up
* export the map, and check that: the textbox for the title is wider; the "Export Legend" checkbox is checked by default; in non-Chrome browsers, a message pops up to remind users about printing differences
* check that the exported map looks correct with respect to the given settings, e.g.  the legend should be printed, the page orientation should be correct, etc.

Connects #767

### DIFF
--- a/src/GeositeFramework/Views/Home/Index.cshtml
+++ b/src/GeositeFramework/Views/Home/Index.cshtml
@@ -443,10 +443,11 @@
                         <h3 class="i18n" data-i18n="Orientation">Orientation</h3>
                         <label><input type="radio" name="export-orientation" value="Landscape" checked="checked"/><span class="i18n" data-i18n="Landscape">Landscape</label>
                         <label><input type="radio" name="export-orientation" value="Portrait"/><span class="i18n" data-i18n="Portrait">Portrait</label>
+                        <p class="font-italic small i18n" data-i18n="Browser Print">* You may need to alter your browser print settings</p>
                     </div>
                     <div class="popover-section">
                          <h3 class="i18n" data-i18n="Legend">Legend</h3>
-                        <label><input type="checkbox" name="export-include-legend" /> <span class="i18n" data-i18n="Include Layer Legend">Include Layer Legend</span></label>
+                        <label><input type="checkbox" name="export-include-legend" checked/> <span class="i18n" data-i18n="Include Layer Legend">Include Layer Legend</span></label>
                     </div>
                 </div>
                 <div id="export-print-preview-container">
--- a/src/GeositeFramework/css/app-print.css
+++ b/src/GeositeFramework/css/app-print.css
@@ -1,4 +1,24 @@
 ﻿@media print {
+    .print-sandbox-header h1 {
+        width: 100%;
+        padding-top: 0.3in;
+        text-align: center;
+    }
+
+    .esriScalebar {
+        visibility: hidden;
+    }
+
+    .print-sandbox-header {
+        display: flex;
+        flex-direction: row;
+        height: 0.95in;
+    }
+
+    #print-map-container {
+        overflow: hidden !important;
+    }
+
     #map-print-sandbox {
         visibility: visible !important;
     }
@@ -10,6 +30,32 @@
     #export-print-preview-container {
         padding-bottom: 5px !important;
         visibility: visible;
+    }
+
+    #legend-container-0 {
+        background-color: #fff !important;
+    }
+
+    .legend-close {
+        visibility: hidden;
+    }
+    
+    #print-map-container > #export-print-preview-map > #map-0 {
+        height: 100% !important;
+        width: 100% !important;
+        position: relative;
+    }
+    
+    #legend-container-0 {
+        margin-left: 0px !important;
+        left: 0px !important;
+        width: 100% !important;
+        position: absolute !important;
+        z-index: 10;
+    }
+    
+    .dojoResizeHandle {
+        visibility: hidden !important;
     }
 
     #left-pane {
@@ -55,6 +101,49 @@
     .logo-med {
         display: none !important;
     }
+    
+    .legend-header {
+        height: 15%;
+        font-size: 11px;
+    }
+
+    .legend-body {
+        display: flex;
+        flex-wrap: wrap;
+        width: 100% !important;
+        height: 85% !important;
+    }
+
+    .layer-legends {
+        height: 100% !important;
+        width: 100% !important;
+        display: flex;
+        flex-direction: column;
+        flex-wrap: wrap;
+    }
+
+    .legend-layer {
+        width: auto !important;
+        font-size: 8px;
+        height: auto !important;
+        display: flex;
+        flex-direction: column;
+        flex-shrink: 1;
+    }
+
+    .legend-title {
+        max-width: 75%;
+    }
+
+    .print-legend-coll {
+        display: flex !important;
+        flex-direction: column;
+    }
+
+    .legend-layer img {
+        width: 10px;
+        height: 10px;
+    }
 }
 
 #export-print-preview-container {
@@ -95,4 +184,44 @@
 
 .instructions {
     display: none;
+}
+
+.legend-close {
+    visibility: hidden;
+}
+
+.dojoResizeHandle {
+    visibility: hidden !important;
+}
+
+#export-title {
+    width: 100%;
+}
+.print-sandbox-header .hdr-container {
+    position: fixed;
+    height: 0.85in;
+    text-align: center;
+    padding-left: 160px;
+    display: flex;
+    flex-direction: row;
+}
+
+.print-sandbox-header h1 {
+    width: 100%;
+    padding-top: 0.3in;
+    text-align: center;
+}
+
+.print-sandbox-header {
+    display: flex;
+    flex-direction: row;
+    height: 0.95in;
+}
+
+.esriScalebar {
+    visibility: hidden;
+}
+    
+#print-map-container {
+    overflow: hidden !important;
 }
--- a/src/GeositeFramework/css/print-landscape.css
+++ b/src/GeositeFramework/css/print-landscape.css
@@ -8,12 +8,7 @@
         height: 6in !important;
     }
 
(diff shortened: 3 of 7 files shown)


jtarricone:
This should be postponed until #763 is merged.

kdeloach:
Do we need the page orientation option anymore?

Considering that we are no longer generating static printouts from ArcGIS API (as far as I know) and also users can adjust the page orientation in the print dialog settings.

screenshot 2016-12-13 at 15 03 44

jtarricone:
There are differences between the portrait and landscape stylesheets, and there was some weirdness with IE11 that I noticed, such as that it at least appeared that the `media` section wasn't the one whose rules were actually applied when printed.

kdeloach:
I can test this today.

src/GeositeFramework/js/Export.js
```diff
@@ -105,6 +129,20 @@ require(['use!Geosite'],
                                 .wrapAll("<div class='print-legend-coll'></div>");
                         });
 
+                    function afterPrintHandler() {
+                        _.delay(previewDeferred.resolve, 250);
+                    }
+
+                    $.when(alertDeferred).then(function () {
+                        if (!!window.onafterprint) {//ie, ff, safari?
+                            window.onafterprint = afterPrintHandler;
```
kdeloach: Consider adding some documentation to explain why we have to perform browser sniffing. What is the difference between IE/FF/Safari and Chrome?

src/GeositeFramework/css/app-print.css
```diff
@@ -95,4 +119,16 @@
...
+
+.legend-close {
+    visibility: hidden;
+}
+
+.dojoResizeHandle {
+    visibility: hidden !important;
+}
+
+#export-title {
```
kdeloach: Wasn't sure where to put this comment, but according to issue #767, the print title should be center aligned.

image

jtarricone: Good catch, I missed that one. I'm guessing #767 meant centered vertically and laterally, is that correct @mmcfarland ?

src/GeositeFramework/js/Export.js
```diff
@@ -105,6 +129,20 @@ require(['use!Geosite'],
...
+                        }
+                        window.print();
```
jtarricone: Did the alert box appear after you clicked through the main export dialog?

jtarricone: It may be worth checking that your IE11 is up to date, when I tested against it with a base install there were some weird jQuery bugs that made the print feature unusable. Installing some security patch cleared it up.

A quick way to check, at least through my experience, is to open the IE11 developer tools and see if anything comes up. Before the patch there was no content in any of the panes.

src/GeositeFramework/js/Export.js
```diff
@@ -41,6 +41,19 @@ require(['use!Geosite'],
                 return 'css/print-portrait.css';
             }
         }
+        
+        function canSetPageLayout() {
```
jtarricone: To add, non-Chrome browsers expose an `onafterprint` event, which is what is used to destroy the modified styles and put the map back in place. The core problem is that `window.print` didn't consistently block across browsers.

src/GeositeFramework/css/app-print.css
mmcfarland: That's correct.

src/GeositeFramework/js/Export.js
kdeloach: Lines 146 and 147 look suspicious. Were they supposed to be deleted? That might explain the cause behind this issue.

src/GeositeFramework/js/Export.js
jtarricone: That may be a simpler route, I can test it out and see how it goes. I'll also add a blurb as a comment paraphrasing what I mentioned earlier.

src/GeositeFramework/js/Export.js
jtarricone: Good catch, those must have slipped in when I rebased.

kdeloach:
More of a general comment, but the print preview already looks pretty good on the prototype branch even without targeting Chrome specific features. Switching between landscape/portrait on the system dialog works pretty well.

image

image

jtarricone:
For purposes of comparison, here's Firefox's default print output (map was centered on Philadelphia)-->
scrn2
scrn1

IE11 -->
scrn4
scrn3

jtarricone:
I made a couple updates. The first was to hide the scalebar that was recently added in the printed output. The second was to group common style rules into `app-print` and try to leave the orientation-specific sheets with only unique rules.

kdeloach:
This works much better in IE now. However, I noticed that the "Your browser does not automatically detect page layout" popup now appears in Chrome. Clicking "Okay" triggers the print preview like normal.

Another interesting issue is that If you press Escape or close the "page set" popup, the map is partially blanked out (fixed by resizing the window). This happens in Chrome but not IE.

image

kdeloach:
I'm also not sure why we need to display a special popup for page orientation either. It's unlikely that users will know what this means or why it's being displayed.

Need a second opinion.

@mmcfarland ^

image

jtarricone:
The message is obviously open for edits to make it clearer, but regardless there should probably be some type of message, or, barring that, maybe the landscape/portrait selector can be reworded on the previous form. The reason I added it is that there's a clear disconnect between what a user would reasonably expect by selecting the layout on the prior form, and what they would likely get with default browser print settings.

jtarricone:
I think the issue in Chrome was caused by removing the explicit check for Chrome, and instead relying on `window.afterprint`. The check has to be whether that event is, specifically, `undefined`, as both `undefined` (Chrome) and `null` (IE/FF/...) will result in the same evaluation of `!!window.onafterprint`. I made a change that should fix that, but my VM isn't running Chrome at the moment so I havent tested it yet.

ajrobbins:
I agree that the page orientation warning popup is confusing. Does the print preview not reflect the landscape/portrait selection as is?

jtarricone:
No, actual printed output will be whatever the browser setting is. By default I think it's typically portrait in IE and FF. Complicating it further is the fact that only Chrome seems to do anything similar to a "preview" on a basic `window.print` invocation; other browsers have separate processes that aren't exposed to the javascript runtime. In those browsers, users would have no idea what the print output would be until it was actually printed.

ajrobbins:
I see. I think we should work that warning into the landscape/portait selector, if at all. People are used to that window, and having an additional one is confusing in my opinion.

mmcfarland:
I think we could cover that case with a little text update, something like this:

screenshot from 2016-12-19 12 28 28

jtarricone:
I pushed an update that replace the form with the generic message.

src/GeositeFramework/js/Export.js
```diff
@@ -66,7 +73,7 @@ require(['use!Geosite'],
                     'class': '.print-orientation-css',
                 }).appendTo('head');
                 _.delay(orientDeferred.resolve, 500);
-
+
```
kdeloach: Whitespace.

src/GeositeFramework/js/Export.js
```diff
@@ -34,14 +34,21 @@ require(['use!Geosite'],
...
         function pageCssLink(pageOrientation) {
             if (pageOrientation === 'Landscape') {
                 return 'css/print-landscape.css';
             } else {
                 return 'css/print-portrait.css';
             }
         }
-
+
```
kdeloach: Whitespace.

src/GeositeFramework/js/Export.js
```diff
@@ -105,14 +112,28 @@ require(['use!Geosite'],
...
+                        _.delay(previewDeferred.resolve, 250);
+                    }
+
+                    /* 
+                        Chrome's exposed 'window.print' method includes a preview and
+                        blocks, whereas non-Chrome browsers do neither; similarly,
+                        Chrome does not expose an 'onafterprint' event while the
+                        others do.
+                    */
+                    if (!isChrome()) {//ie, ff, safari?
```
kdeloach: The line comments here and below could be removed now that the code is more self-explanatory.

src/GeositeFramework/css/print-portrait.css
```diff
@@ -2,18 +2,13 @@
     @page {
         size: letter portrait;
     }
-
+
```
kdeloach: Whitespace.

src/GeositeFramework/css/print-landscape.css
```diff
@@ -2,18 +2,13 @@
     @page {
         size: letter landscape;
     }
-
+
```
kdeloach: Whitespace.

src/GeositeFramework/Views/Home/Index.cshtml
```diff
@@ -463,20 +464,6 @@
...
-                <div class="row">
-                    <div class="left-col">
-                        <div class="export-output-area">
-                            <a href="#" id="print-alert-okay" class="button button-primary radius">Okay</a>
-                        </div>
-                    </div>
-                    <div class="right-col">
-                    </div>
-                </div>
-            </div>
```
kdeloach: Consider squashing these commits since the second commit reverts things that were added in the first commit.

src/GeositeFramework/Views/Home/Index.cshtml
```diff
@@ -443,6 +443,7 @@
                         <h3 class="i18n" data-i18n="Orientation">Orientation</h3>
                         <label><input type="radio" name="export-orientation" value="Landscape" checked="checked"/><span class="i18n" data-i18n="Landscape">Landscape</label>
                         <label><input type="radio" name="export-orientation" value="Portrait"/><span class="i18n" data-i18n="Portrait">Portrait</label>
+                        <p class="font-italic small">* You may need to alter your browser print settings</p>
```
kdeloach: Don't forget to add i18n strings for this text.

Issue #749, opened 2016-11-22
Implement "view more" for plugin/scenario description in launchpad


caseypt:
Originally from #736 

> If the plugin description is greater than a given height it is truncated with a View more link to expand the rest of the text.

PR #772, opened 2016-12-13
Truncate Launchpad description text

jtarricone:
This implements a simple truncation on the description elements in the launch pad. When the text elements are over a certain height, the element is resized, a '...' is shown at the end, and a toggle button is added.

download 2
download 3

To test:
* clone this branch, bring up the environment, and open the app
* click on the sidebar item to open the launchpad
* check that the first item is truncated and has a "View more" button
* check that clicking the button toggle the text truncation
* check that the other items are not truncated

Connects #749


jtarricone:
Ignore all but the last commit. I'll rebase this as soon as #763 is merged.

kdeloach:
Mind removing these unrelated commits? Then you won't have to rebase when other pull requests are merged.

jtarricone:
I rebased on `prototype`, this PR is independent of my other ones.

mmcfarland:
I believe the referenced issue describes the text associated with a plugin, which are the elements under the"start exploring" header and not the main text body.  It may be useful to have it there as well, but it needs to be for the plugins primarily.

jtarricone:
I'll update this to remove the truncation of the main text, but as it stands the plugin descriptions are also truncated the same way, but none them are currently long enough trigger the adjustment. What is the height limit for those?

mmcfarland:
The original request wasn't expressed as a height limit but rather, to be "twitter-like", which is to say a few sentences.  I'm not sure how that translates to height, or if things like dpi would affect that across different computers.

src/GeositeFramework/plugins/launchpad/main.js
```diff
@@ -86,6 +86,33 @@ define([
...
+                                        this.textContent = 'View less';
+                                    } else {
+                                        this.textContent = 'View more';
+                                    }
+                                });
+                            }
+                        });
+                    } else {
+                        $(".truncate-toggle", this).css({ display:"none" });
+                    }
```
kdeloach: Couple of suggestions:

* This code does exactly what #749 suggests ("If the plugin description is greater than a given height ..."). However, I would recommend using text length instead of DOM height. Then we could move half of this logic to the template.

  Ex.

  ```
  <p><%= shortDescription %></p>
  <% if (description.length > 100) { %>
  <button class="button truncate-toggle">View more</button>
  <% } %>
  ```

* Consider moving the click events to the `bindEvents` function. We try to use live selectors whenever possible.
* For parity with the rest of the codebase, prefer to use jQuery functions to manipulate DOM attributes (ex. `$(this).text(...)` instead of `this.textContent`, and `$(this).toggle(...)` instead of `this.classList.toggle`)
* Defining hard-coded strings in code instead of templates presents some internationalization difficulties. Consider using an ellipsis instead to avoid this problem? Maybe even get rid of the option to "View Less" to keep things simple.

src/GeositeFramework/plugins/launchpad/style.css
```diff
@@ -26,3 +26,41 @@
     list-style-type: none;
     padding: 0;
 }
+
+.truncate-toggle {
+    padding: 2px !important;
+    font-size: 0.7em !important;
+    width: 35%;
+    float: none !important;
```
kdeloach: I think `!important` should be reserved for extreme cases. Like when there is third party CSS we physically cannot modify. Using it everywhere makes styles harder to override (for example, in print mode).

jtarricone:
The description in #749 mentioned height as the criterion, but regardless I think basing it on height is probably the most straightforward and possibly the most robust solution. I think `px` values, as far as CSS is concerned, are angular units as opposed to linear ones, so I think a height specified in `px` would be stable across viewers. There should maybe be some coupling between whatever the truncated height is and the font size within the container, once the desired effect is decided.

How long of text blocks are these intended to be?

src/GeositeFramework/plugins/launchpad/main.js
jtarricone: * Fair enough, but basing the truncation on element height rather than text length has the advantage of maintaining consistent behavior regardless of what the font size is or how it's defined (if that's the desired behavior). I don't think having the condition in the template is practically much simpler than evaluating it at runtime, respective of each criterion, but ultimately there should be some clarification of what the original request was
* Good point, I'll move the block to `bindEvents`
* Good point about using jQuery functions, I'll update those
* I used the "view more" label because that seemed to be implied in #749. I would think an ellipsis and/or up/down icons would be enough to get the intent across. Whether nixing the "view less" option is a good idea rather depends on how long the text blocks are expected to be.

jtarricone:
I just pushed an update that addresses the points that didn't require further clarification.

src/GeositeFramework/plugins/launchpad/style.css
jtarricone: The basic functionality is unchanged if the `button` class is removed, but all the accumulated styling goes with it, so it seems using `!important` is probably the easiest way to go.

src/GeositeFramework/plugins/launchpad/main.js
kdeloach: I agree this is more of a design issue. But I would argue that using text length is simpler and looks better. Making sure the "truncate container" height lines up with the text will be an extra chore.

See the difference between these two options:

1. Height
  image

2. Text length
  image

src/GeositeFramework/plugins/launchpad/style.css
kdeloach: In this case, you could increase the specificity of the rule to `.launchpad-plugin .plugins .button.truncate-toggle` to avoid using `!important`.

src/GeositeFramework/plugins/launchpad/main.js
```diff
@@ -54,6 +54,19 @@ define([
                         var saveCode = $(e.target).data('saveCode');
                         self.plugin.turnOff();
                         self.triggerEvent('launchpad:activate-scenario', saveCode);
+                    })
+                    .on('click', 'button.truncate-toggle', function (e) {
+                        var parentEl = $(e.target.parentElement);
+                        $('p', parentEl).each(function() {
+                            $(this).toggleClass('truncate-box');
+                        });
```
kdeloach: I think the jQuery way of doing this would be something like `$el.find('p').toggleClass(...)`.

src/GeositeFramework/plugins/launchpad/main.js
```diff
@@ -54,6 +54,19 @@ define([
...
+                        $('.truncate-toggle', parentEl).each(function() {
+                            if ($(this).text() === 'View more') {
+                                $(this).text('View less');
+                            } else {
+                                $(this).text('View more');
+                            }
```
kdeloach: We should handle i18n before merging.

src/GeositeFramework/plugins/launchpad/main.js
```diff
@@ -86,6 +99,18 @@ define([
...
                 $(this.container).empty().append($el);
+
+                // add truncate listeners to each toggle botton
+                $(".truncate-container").each(function () {
+                    if (parseFloat($('p', this).css("height")) > 30) {
+                        $('p', this).each(function() {
+                            $(this).toggleClass('truncate-box');
+                        });
+                    } else {
+                        $(".truncate-toggle", this).css({ display:"none" });
```
kdeloach: Consider using jQuery's `hide` and `show` functions instead.

src/GeositeFramework/plugins/launchpad/main.js
jtarricone: I'm not sure how literal the blurb in #749 was intended to be, but again I'd think and ellipsis and/or arrow icons would be sufficient.

ajrobbins:
Chiming in here:
- Original "height" request came from the client and should not be taken literally if there is a better way to implement the limitation across browsers.
- The idea of "Twitter-length" was suggested, so if we could limit to 140 characters (and that length looks good in this layout) that would be one option
- Needs to work with i18n
- I like the ellipses and "View More". Tagging @jfrankl who will likely follow with or have style recommendations

jtarricone:
After the clarifications, I updated this to truncate at 140 characters and refactored where necessary. I updated the `region.json` file to contain a description longer than 140 characters just for purposes of always having this feature on the launchpad.

jfrankl:
I would recommend implementing a 140 character maximum on plugin descriptions rather than use a view more text function like this. I think we should be encouraging plugin developers to write brief, easy to understand descriptions. Have people run into problems where they need to write longer descriptions, or is this predicting a future issue?

jfrankl:
I don't think the Launchpad is a good place for a "long" or "detailed" description. It should contain a straightforward introduction. If more of an explanation is needed, could it be in the application itself?

jtarricone:
I pushed another update that implements the i18n text for the control

ajrobbins:
@jfrankl I agree with you in principle, but based on specific requests from TNC I think we need to allow for the possibility that someone writes a description that's longer than 140 characters.

jtarricone:
What's the verdict on this?

mmcfarland:
I'd recommend text length restriction over height.

jtarricone:
@mmcfarland it's been updated to truncate at 140 characters. I meant verdict with respect to whether it needs to be merged, or if the feature is no longer required.

mmcfarland:
Still required.

src/GeositeFramework/plugins/launchpad/style.css
```diff
@@ -26,3 +26,41 @@
     list-style-type: none;
     padding: 0;
 }
+
+.launchpad-plugin .plugins .button.truncate-toggle, .scenarios .button.truncate-toggle {
+    padding: 2px !important;
+    font-size: 0.7em !important;
```
kdeloach: Consider removing `!important` from these lines. There didn't seem to be a visual difference when I removed them during testing.

src/GeositeFramework/plugins/launchpad/style.css
```diff
@@ -26,3 +26,41 @@
...
+    width: 35%;
+    float: none;
+}
+
+.truncate-box {
+    overflow: hidden;
+    position: relative;
+    line-height: 1.2em;
+    max-height: 2.6em;
```
kdeloach: I don't think we need to clip this container by height anymore, now that we're truncating the text to 140 characters. We may not need this container at all.

src/GeositeFramework/plugins/launchpad/style.css
```diff
@@ -26,3 +26,41 @@
...
+}
+
+.truncate-box:before {
+    content: '(...)';
+    position: absolute;
+    right: 5px;
+    bottom: 2px;
+}
+
+.truncate-box:after {
```
kdeloach: I removed this `.truncate-box:after` block during testing and it made no visual difference. Consider removing.

src/GeositeFramework/plugins/launchpad/style.css
```diff
@@ -26,3 +26,41 @@
...
+    position: relative;
+    line-height: 1.2em;
+    max-height: 2.6em;
+    text-align: justify;
+    margin-right:-1em;
+    padding-right: 2em;
+}
+
+.truncate-container > p {
+    width: 90%;
```
kdeloach: The text width is much narrower than the original design. I think some of this CSS and markup changes should be changed to preserve the original style.

Before:
image

After:
image

src/GeositeFramework/plugins/launchpad/main.js
```diff
@@ -54,6 +54,18 @@ define([
...
+                    .on('click', 'button.truncate-toggle', function (e) {
+                        var parentEl = $(e.target.parentElement);
+                        parentEl.find('p').toggleClass('truncate-box');
+                        parentEl.find('.truncate-toggle').each(function () {
+                            if ($(this).attr('data-i18n') === 'See More') {
+                                $(this).attr('data-i18n', 'See Less');
+                            } else {
+                                $(this).attr('data-i18n', 'See More');
+                            }
+                            $(this).localize();
```
kdeloach: Imagine how one might implement this feature using React. Clicking on the toggle button would update the component state, then we would re-render the component using the new state. 

That is the general design pattern that we are trying to follow with TNC plugins, and that's what I would recommend that we do here. The result should be a clean separation between the state and the template. No DOM manipulation should be necessary.

src/GeositeFramework/plugins/launchpad/templates.html
```diff
@@ -23,7 +28,12 @@
                 <li>
                     <div>
                         <h4><%= scenario.title %></h4>
-                        <p><%= scenario.description %></p>
+                        <div class="truncate-container">
+                            <p><%= scenario.description %></p>
+                            <% if (scenario.description.length > 140) { %>
+                                <button class="button truncate-toggle i18n" data-i18n="[title]See Less" title="See Less"></button>
```
kdeloach: Not sure if this was intentional, but the button markup is a little different for scenarios compared to plugins. I noticed that the tooltip doesn't always match the button text.

image

src/GeositeFramework/plugins/launchpad/main.js
```diff
@@ -86,6 +98,7 @@ define([
                 }));
 
                 $(this.container).empty().append($el);
+                $('.truncate-toggle').click();
```
jtarricone: It can be done in the template instead.

src/GeositeFramework/plugins/launchpad/style.css
jtarricone: The text itself isn't truncated, the box class is conditioned on the text length. The visible height is the mechanism of hiding the text.

src/GeositeFramework/locales/en.json
```diff
@@ -69,4 +69,6 @@
     "Length:": "Length:",
     "Double-click to end a line segment. Single-click the first node to end a polygon.": "Double-click to end a line segment. Single-click the first node to end a polygon.",
     "Click the map to start measuring": "Click the map to start measuring"
+    "See More": "See More",
```
kdeloach: We need trailing commas on line 71 to make this valid JSON.

src/GeositeFramework/plugins/launchpad/main.js
kdeloach: Ultimately, we shouldn't be making decisions based on i18n attributes. We can pair on this.

src/GeositeFramework/plugins/launchpad/style.css
```diff
@@ -26,3 +26,28 @@
     list-style-type: none;
     padding: 0;
 }
+
+.launchpad-plugin .plugins .button.truncate-toggle, .scenarios .button.truncate-toggle {
```
kdeloach: Minor, but consider adding a line break after the comma so it's easier to tell this style has 2 associated rules.

src/GeositeFramework/plugins/launchpad/style.css
kdeloach: Let's discuss in person.

src/GeositeFramework/plugins/launchpad/templates.html
```diff
@@ -23,7 +30,14 @@
                 <li>
                     <div>
                         <h4><%= scenario.title %></h4>
-                        <p><%= scenario.description %></p>
+                        <div class="truncate-container">
+                            <% if (scenario.description.length > 140) { %>
+                                <p class="truncate-box"><%= scenario.description %></p>
+                                <button class="button truncate-toggle i18n" data-i18n="[title]See More" title="See More"></button>
```
kdeloach: Not sure if this was intentional, but scenario buttons have a title attribute, but the plugin buttons do not.

src/GeositeFramework/plugins/launchpad/style.css
kdeloach: Might make more sense to address this in a future PR if it becomes an issue.

jtarricone:
After clarifying some points, I refactored this a bit to instead use a state object to avoid doing direct DOM manipulation.

Issue #782, opened 2016-12-15
Remove infographic and add mechanism for in-app help


mmcfarland:
The framework will no longer provide an infoGraphic option in the configuration, nor provide any UI elements to manage the _content_ of the help.  Instead, it will provide light mechanics to alert the plugin when it should activate its own internally developed help system.  Method/Config names below are merely suggestions to help explain purpose.

* Framework will provide a ? or similar on the plugin container (much like the current infoGraphic)
* Plugin config will include an optional `hasHelp: true|false (default)` setting which will determine if the ? is shown
* When the ? is clicked, the framework will execute a new registered function `showHelp` which the plugin can respond to as needed, internally
* The app object passed to the plugin will have a new method to communicate to the framework, `setShowHelpOnStartup(true|false)`.  This will allow the plugin to determine if the help should be shown when the plugin is activated.  This is persisted, per plugin, similar to how infoGraphic is handled. The plugin may set this values whenever it likes, but most likely would happen:
    * When a user clicks a "do not show" checkbox or button in the in-app help
    * After the first time the app is activated, so it only shows once per user
* The framework passes a new boolean value (from above), `showHelp`, to the activate method on the plugin when it's called, so the plugin knows to activate the help section.  If it never wants to show up on start, it would `setShowHelpOnStartup(false)` in the initialize method.
* The framework will take care of actually storing that value and passing along to the app
* Clicking the ? would call the method no matter the `localStorage` setting

PR #795, opened 2016-12-16
Remove inforgraphic in favor of in-app help

mmcfarland:
#### Overview
Plugins can register with the framework to acknowledge that there is
in-app help functionality available.  The framework handles the UI of
showing the help icon in the header and for managing the state of "show on
start" via local storage.  Its intended purpose is to allow plugins to
manage their own help functionality and move away from the exising
infographic approach.

The details of the approach should match the associated issue, which is descriptive

Connects #782 

#### Testing
* Load the Itendify Point plugin, verify that help is shown.
* Dismiss the in-app help, and click the `i` info button in the header, verify help is shown again (dismiss)
* Close plugin and re-open, verify that the help is not shown
* Refresh the page and do the same.

The goal is to allow the plugin to handle some in app instruction, with a light UI mechanism from the framework and the ability to manage whether the app shows help upon activation.  It's meant to be flexible, and the above is just one workflow that could accommodate it.

### DIFF
--- a/src/GeositeFramework/js/Plugin.js
+++ b/src/GeositeFramework/js/Plugin.js
@@ -46,7 +46,8 @@ require(['use!Geosite',
                         _unsafeMap: esriMap,
                         downloadAsCsv: requestCsvDownload,
                         downloadAsPlainText: requestTextDownload,
-                        dispatcher: N.app.dispatcher
+                        dispatcher: N.app.dispatcher,
+                        supressHelpOnStartup: _.partial(supressHelpOnStartup, model)
                     },
                     plugin: {
                         turnOff: model.turnOff.bind(model)
@@ -105,6 +106,12 @@ require(['use!Geosite',
             return model.get('pluginSrcFolder');
         }
 
+        // Allow the plugin to ask the framework to store and provide a
+        // flag for showing help when activating, across sessions
+        function supressHelpOnStartup(model, doSupress) {
+            model.setSupressHelpOnStartup(doSupress);
+        }
+
         /*
         Check that the plugin implements the minimal viable interface.
         Plugin code can just assume the plugin is valid if it has been loaded
@@ -127,9 +134,9 @@ require(['use!Geosite',
         N.models = N.models || {};
         N.models.Plugin = Backbone.Model.extend({
             defaults: {
+                startHelpKey: "-supress-help-start",
                 pluginObject: null,
                 active: false,
-                displayHelp: false,
                 pluginLayers: {},
                 pluginLayersVisible: true
             },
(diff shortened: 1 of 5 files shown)


jtarricone:
I noticed the help message might display in an unintended case. After switching to the branch, the app behaved as expected in the testing instructions. Upon page refresh and hard refresh, the help text was not displayed. Switching to another browser and loading it for the first time (in that browser), the help text was not displayed. I think this is the desired behavior.

However, selecting the "Identify" plugin (w/o help message) and then activating another plugin, then switching back brought the message back, but it didn't seem to be persisted in the sense that another browser didn't show the message when the plugin was activated.

jtarricone:
In IE11, clicking "Got It" to get rid of the help message resulted in a scrollbar showing on the right of the sidebar.

src/GeositeFramework/sample_plugins/identify_point/main.js
```diff
@@ -43,6 +53,28 @@ define(["dojo/_base/declare", "framework/PluginBase"],
...
+
+                self.$helpMsg = $("<div>Help: Click the map to see the point </div>");
+
+                $("<a>", {
+                    className: "button",
+                    href: '#',
+                    text: "Got it",
+                    click: function() {
+                        self.$helpMsg.remove();
+                        this.remove();
```
jtarricone: This throws when evaluated in IE11.

src/GeositeFramework/js/Plugin.js
```diff
@@ -46,7 +46,8 @@ require(['use!Geosite',
                         _unsafeMap: esriMap,
                         downloadAsCsv: requestCsvDownload,
                         downloadAsPlainText: requestTextDownload,
-                        dispatcher: N.app.dispatcher
+                        dispatcher: N.app.dispatcher,
+                        supressHelpOnStartup: _.partial(supressHelpOnStartup, model)
```
jtarricone: Not consequential, but "suppress" is misspelled.

jtarricone:
It looks like any toggling of the plugin is enough to reactivate the help message, whether it's just (de)-activating the same one or switching to another one and then back. Is the Identify plugin supposed to have the behavior described in #782 or is that one not intended to persist the setting?

mmcfarland:
Updated with fix for toggle bug and spelling errors.  The error in IE was vestigial code, so I removed it and I'm not concerned about the sidebar scroll since this is just a proof of concept example and probably will never be run in IE.

If you want to "unset" the suppress flag, you can do so in the devtools Application tab:
screenshot from 2016-12-20 14 17 41

Issue #790, opened 2016-12-15
Print fixups


mmcfarland:
* Scalebar obscures legend 
* Scalebar is not visible when above is fixed because the map is not resized to account for legend height.  If legend is added, it should resize the map to make sure the attribution and scale bar are visible.

PR #799, opened 2016-12-19
Print fixups

jtarricone:
This fixes a bug that resulted in the map scalebar and the attribution elements being obscured by the legend.

To test:
* clone this branch and open the app
* print a basic, unmodified map both with and without the legend, and check that the scalebar and the attribution are at the bottom of the page
* add layers that cause the legend to appear, then print with and without legend and check that the scalebar and attribution element move appropriately
* repeat last 2 steps for landscape/portrait layouts, in targeted browsers

Connects #790

### DIFF
--- a/src/GeositeFramework/css/app-print.css
+++ b/src/GeositeFramework/css/app-print.css
@@ -5,10 +5,6 @@
         text-align: center;
     }
 
-    .esriScalebar {
-        visibility: hidden;
-    }
-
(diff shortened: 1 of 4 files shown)


kdeloach:
Mind pushing this feature branch upstream and updating the branch name to reflect our [pull request guidelines](https://github.com/azavea/civic-apps/blob/master/guidelines/001_github_review_process.md)?

Issue #785, opened 2016-12-15
Add minimize button to plugin header


ajrobbins:
This functionality was available in the old framework version:

![screen shot 2016-12-15 at 11 27 36 am](https://cloud.githubusercontent.com/assets/12868342/21232472/b762287c-c2b9-11e6-9b92-4f59332c0829.png)

Add a minimize `_` button to plugin headers. Clicking on this should collapse the plugin sidebar while keeping the plugin's layers on.

PR #801, opened 2016-12-20
Add Plugin "Minimize" button

kellyi:
## Overview

This PR adds a minimize button to the sidebar plugin header. The feature had been available in the prior version, but until now we hadn't yet implemented it in the prototype. Here we add an underscore icon to the sidebar plugin header, and call the `.deselect()` method when it's clicked. This removes the plugin from the map viewport but leaves its layers active.

### Demo

screen shot 2016-12-20 at 10 43 02 am

& ->

screen shot 2016-12-20 at 10 43 12 am

### Notes

The code to add the minimize button includes an `<i class="icon-minimize">` element which is used primarily to size the button properly, in accordance with the other sidebar header controls.

The first and third commits here involve some cleanup work:
- in d0ed543, we remove the `map-utils` plugin references from the csproj file. The map-utils are now part of the core framework -- and not a plugin -- but the references hadn't been removed from Visual Studio, which meant the plugin dir had to be deleted manually to get the project to build
- in 4af0cb5 I cut a reference to `plugin-eye`, which was added in #702 then subsequently removed

## Testing Instructions

 * get and build the project and open it in the browser, along with the dev tools
 * open the Regional Planning plugin, pan to New York, and turn on the Hurricane Sandy layers
 * click the "_" button to minimize and verify that the sidebar plugin disappears but "Regional Planning" remains highlighted and the layers remain on the map

Connects #785


src/GeositeFramework/Views/Home/Index.cshtml
```diff
@@ -384,8 +384,12 @@
                             <i class="icon-info-circled-1"></i>
                         </button>
                     <% } %>
-
-                    <button class="button-link plugin-off i18n" data-i18n="[title]Turn off this plugin" title="Turn off this plugin"><i class="icon-cancel"></i></button>
+                    <button class="button-link plugin-minimize i18n" data-i18n="[title]Minimize this plugin" title="Minimize this plugin">
+                        <i class="icon-minimize"></i>_
+                    </button>
+                    <button class="button-link plugin-off i18n" data-i18n="[title]Turn off this plugin" title="Turn off this plugin">
```
mmcfarland: Since you've added the i18n stuff here, can you add the text and a google translation of spanish into:

https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/master/src/GeositeFramework/locales/en.json
https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/master/src/GeositeFramework/locales/es.json

kellyi: Sure!

kdeloach:
We may want to rethink the launchpad minimize/close behavior. It's inconsistent with other plugins. The minimize button on the launchpad closes the plugin instead of minimizing it.

Ref https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/prototype/src/GeositeFramework/plugins/launchpad/main.js#L87

kellyi:
Pushed a new commit which updates the Fontello icons to include `window-minimize`, which is now used in place of the underscore:

screen shot 2016-12-20 at 12 29 29 pm

kellyi:
Thanks for taking a look! I squashed the fixup & I'm going to merge.

Issue #788, opened 2016-12-15
Add setWidth to app object for plugins


mmcfarland:
Allow plugins to update their size while open.  A new method should take `'small', 'large'` or an integer (for `custom` and the framework will update the size.  It should return a deferred which the plugin will use to trigger a UI change.  Usage something like:

```js
app.setWidth('large').then(function() {

  // width has changed, do stuff

});
```

PR #803, opened 2016-12-20
Add resize method to plugin

jtarricone:
This adds a couple resize methods to the plugin object. As mentioned in the original issue, these return `Deferred` objects.

To test:
* clone this branch
* try invoking the resize methods on plugins, checking that the sidebar width changes as expected,  persists when the plugin is closed or asleep, and that other actions can be chained to the deferred object.

I did the following to test this locally -->
* add the following to `main.js` of the sample identifiy plugin starting after line 34:
``` javascript
var self = this;
var d = this.app.resize.setWidth("large");
d.then(function() { $(self.container).css({ color: "white", background: "black" }); });
d.resolve();
```
That should execute an action when you click on a point on the map to identify it, that will make the plugin sidebar wider and change its colors.

Connects #788

### DIFF
--- a/src/GeositeFramework/js/Plugin.js
+++ b/src/GeositeFramework/js/Plugin.js
@@ -32,7 +32,36 @@ require(['use!Geosite',
                 pluginName = model.get('pluginSrcFolder'),
                 $uiContainer = model.get('$uiContainer'),
                 $legendContainer = model.get('$legendContainer'),
-                logger = new Logger(pluginName);
+                logger = new Logger(pluginName),
+                resizers = {
+                    setWidth: function(val) {
+                        var d = $.Deferred(),
+                            setter = function(w) {
+                                return function() {
+                                    $uiContainer.css({ width: "" });
+                                    $uiContainer.removeClass(function() {
+                                        var classes = this.className.split(" ");
+                                        return _.filter(classes, function(c) {
+                                            return /sidebar-width-.*/.test(c);
+                                        }).join(" ");
+                                    });
+                                    $uiContainer.addClass("sidebar-width-" + w);
+                                };
+                            };
+                        _.delay(function() {
+                            if (typeof val === "string") {
+                                d.then(setter(val));
+                            } else if (typeof val === "number") {
+                                d.then(function() {
+                                    $uiContainer.css({ width: val });
+                                });
+                            }
+                            d.resolve();
+                            esriMap.resize(true);
+                        }, 100);
+                        return d.promise();
+                    }
+                };
 
             try {
                 pluginObject.initialize({
@@ -47,7 +76,8 @@ require(['use!Geosite',
                         downloadAsCsv: requestCsvDownload,
                         downloadAsPlainText: requestTextDownload,
                         dispatcher: N.app.dispatcher,
-                        suppressHelpOnStartup: _.partial(suppressHelpOnStartup, model)
+                        suppressHelpOnStartup: _.partial(suppressHelpOnStartup, model),
+                        resize: resizers
                     },
                     plugin: {
                         turnOff: model.turnOff.bind(model)
--- a/src/GeositeFramework/sample_plugins/identify_point/main.js
+++ b/src/GeositeFramework/sample_plugins/identify_point/main.js
@@ -16,6 +16,13 @@ define(["dojo/_base/declare", "framework/PluginBase"],
                 $(this.container).append(
                     '<h4 style="padding: 5px;">' + i18next.t('Click any point on the map to display Latitude and Longitude') + '</h4>');
 
+                // add buttons to bind resize events to
+                $(this.container).append('<p>The buttons below demonstrate how to use a plugins "resize" method</p>')
+                    .append('<button class="resize-btn" data-ui-key="resize-ctl-identify-set450">Set to 450</button>')
+                    .append('<button class="resize-btn" data-ui-key="resize-ctl-identify-small">Small</button>')
+                    .append('<button class="resize-btn" data-ui-key="resize-ctl-identify-large">Large</button>')
+                    .append('<p class="width-textbox"></p>');
+
                 // Hide the print button until the identify feature has been used.
                 $(this.printButton).hide();
             },
@@ -31,6 +38,28 @@ define(["dojo/_base/declare", "framework/PluginBase"],
                     // Don't show this help on startup anymore, after the first time 
                     this.app.suppressHelpOnStartup(true);
                 }
+
+                // attach resize events to sample controls, attach actions to jQuery deferred
+                var self = this;
+                _.each($(self.container).find("button.resize-btn"), function (el) {
+                    switch ($(el).data("ui-key")) {
+                        case "resize-ctl-identify-set450":
+                            $(el).on("click", function() {
+                                self.app.resize.setWidth(450).then($(".width-textbox").text("Width is now 450"));
+                            });
+                            break;
+                        case "resize-ctl-identify-large":
+                            $(el).on("click", function() {
+                                self.app.resize.setWidth("large").then($(".width-textbox").text("Width is now large"));
+                            });
+                            break;
+                        case "resize-ctl-identify-small":
+                            $(el).on("click", function() {
+                                self.app.resize.setWidth("small").then($(".width-textbox").text("Width is now small"));
+                            });
+                            break;
+                    }
+                });
             },
 
             identify: function(mapPoint, clickPoint, processResults) {


mmcfarland:
Can you add that snippet in the example to the actual sample plugin?  Then we can point to it when TNC devs want to use the feature.

src/GeositeFramework/js/Plugin.js
```diff
@@ -47,7 +47,43 @@ require(['use!Geosite',
                         downloadAsCsv: requestCsvDownload,
                         downloadAsPlainText: requestTextDownload,
                         dispatcher: N.app.dispatcher,
-                        suppressHelpOnStartup: _.partial(suppressHelpOnStartup, model)
+                        suppressHelpOnStartup: _.partial(suppressHelpOnStartup, model),
+                        resize: {
```
mmcfarland: The rest of the `app` object properties are references or function executions, can you do the same for `resize` so that it doesn't obscure the remaining arguments to `initialize`?

src/GeositeFramework/js/Plugin.js
```diff
@@ -47,7 +47,43 @@ require(['use!Geosite',
...
+                                                $uiContainer.addClass("sidebar-width-small");
+                                                $uiContainer.removeClass("sidebar-width-large");
+                                            });
+                                            break;
+                                        default:
+                                            break;
+                                    }
+                                } else if (typeof (val) === "number") {
+                                    d.then(function() {
+                                        $uiContainer.css({ width: val });
```
mmcfarland: If the plugin ever sets this, then subsequently settings `small` or `large` will never work since the `element.style` will always override a class.  The safest approach would be to remove the `width` css anytime a class is set.  That in turn might push the `case` for both `large` and `small` do be DRY'd up into a function.

src/GeositeFramework/js/Plugin.js
```diff
@@ -47,7 +47,43 @@ require(['use!Geosite',
                         downloadAsCsv: requestCsvDownload,
                         downloadAsPlainText: requestTextDownload,
                         dispatcher: N.app.dispatcher,
-                        suppressHelpOnStartup: _.partial(suppressHelpOnStartup, model)
+                        suppressHelpOnStartup: _.partial(suppressHelpOnStartup, model),
+                        resize: {
+                            adjustBy: function(val) {
+                                var currentWidth = parseFloat($uiContainer.css("width"));
```
mmcfarland: Not entirely consequential, but I don't think the `width` property can actually be a float, consider using `parseInt` instead.

jtarricone: I think the CSS lengths have type `<number>`, which is real-valued.

jtarricone:
I refactored a bit in light of the comments made, and I added an example on the sample plugin, to show the various invocations of the method and chaining additional actions to it.

src/GeositeFramework/js/Plugin.js
```diff
@@ -32,7 +32,40 @@ require(['use!Geosite',
                 pluginName = model.get('pluginSrcFolder'),
                 $uiContainer = model.get('$uiContainer'),
                 $legendContainer = model.get('$legendContainer'),
-                logger = new Logger(pluginName);
+                logger = new Logger(pluginName),
+                resizers = {
+                    adjustBy: function(val) {
+                        var currentWidth = parseFloat($uiContainer.css("width"));
+                        var d = $.Deferred();
```
kdeloach: Consider using the Dojo Deferred class. The docs also have some good example code to demonstrate how deferred objects should be used.

Ref: http://dojotoolkit.org/reference-guide/1.9/dojo/Deferred.html

src/GeositeFramework/sample_plugins/identify_point/main.js
```diff
@@ -16,6 +16,14 @@ define(["dojo/_base/declare", "framework/PluginBase"],
                 $(this.container).append(
                     '<h4 style="padding: 5px;">' + i18next.t('Click any point on the map to display Latitude and Longitude') + '</h4>');
 
+                // add buttons to bind resize events to
+                $(this.container).append('<p>The buttons below demonstrate how to use a plugins "resize" method</p>');
+                $(this.container).append('<button class="resize-btn" data-ui-key="resize-ctl-identify-incr">+</button>');
```
mmcfarland: These appends can all be chained together:
```
$(this.container)
    .append(...)
    .append(...)
    .append(...)
```

src/GeositeFramework/js/Plugin.js
```diff
@@ -32,7 +32,40 @@ require(['use!Geosite',
...
+                                    $uiContainer.css({ width: "" });
+                                    $uiContainer.removeClass(function() {
+                                        var classes = this.className.split(" ");
+                                        return _.reject(classes, function(c) {
+                                            return /sidebar-width-.*/.test(c);
+                                        });
+                                    });
+                                    $uiContainer.addClass("sidebar-width-" + w);
+                                };
+                            };
```
kdeloach: We can simplify this a bit since we're supporting a very limited number of sizes. Ex:

```
$uiContainer.css({ width: 'auto' })
    .removeClass('sidebar-width-small')
    .removeClass('sidebar-width-medium')
    .removeClass('sidebar-width-large')
    .addClass('sidebar-width-' + w);
```

src/GeositeFramework/js/Plugin.js
```diff
@@ -32,7 +32,40 @@ require(['use!Geosite',
                 pluginName = model.get('pluginSrcFolder'),
                 $uiContainer = model.get('$uiContainer'),
                 $legendContainer = model.get('$legendContainer'),
-                logger = new Logger(pluginName);
+                logger = new Logger(pluginName),
+                resizers = {
+                    adjustBy: function(val) {
```
kdeloach: I don't think we need to support this resize operation. Consider removing.

src/GeositeFramework/js/Plugin.js
```diff
@@ -32,7 +32,40 @@ require(['use!Geosite',
...
+                                        var classes = this.className.split(" ");
+                                        return _.reject(classes, function(c) {
+                                            return /sidebar-width-.*/.test(c);
+                                        });
+                                    });
+                                    $uiContainer.addClass("sidebar-width-" + w);
+                                };
+                            };
+                        if (typeof (val) === "string") {
+                            d.then(setter(val));
```
kdeloach: Consider whitelisting the text arguments which are acceptable to this function. I can imagine someone making a small typo such as `setWidth('Large')` and becoming confused as to why nothing happens.

src/GeositeFramework/js/Plugin.js
```diff
@@ -32,7 +32,40 @@ require(['use!Geosite',
...
+                                    $uiContainer.removeClass(function() {
+                                        var classes = this.className.split(" ");
+                                        return _.reject(classes, function(c) {
+                                            return /sidebar-width-.*/.test(c);
+                                        });
+                                    });
+                                    $uiContainer.addClass("sidebar-width-" + w);
+                                };
+                            };
+                        if (typeof (val) === "string") {
```
kdeloach: `typeof` is an operator so parens are not necessary.

src/GeositeFramework/sample_plugins/identify_point/main.js
```diff
@@ -31,6 +39,55 @@ define(["dojo/_base/declare", "framework/PluginBase"],
...
+                        $(self.container).css({ background: color });
+                    };
+                    var blink = function (color, blinks) {
+                        var originalColor = $(el).css("background"),
+                            n = (blinks % 2 === 0) ? blinks : ++blinks;
+
+                        return _.compose(_.map(_.range(0, n), function (i) {
+                            return _.delay(changeColor, i * 100, (i % 2 === 0) ? color : originalColor);
+                        }));
+                    }
```
kdeloach: Consider removing all of this color/blink code which might be confusing to plugin developers. 

One practical example that we could demonstrate is how to update a fixed width element, after a resize has completed. For example, we could add a table with sample data that is fixed at 100px, 200px, or 300px,  depending on whether the container is small, medium, or large.

src/GeositeFramework/js/Plugin.js
```diff
@@ -32,7 +32,40 @@ require(['use!Geosite',
...
+                            d.then(setter(val));
+                        } else if (typeof (val) === "number") {
+                            d.then(function() {
+                                $uiContainer.css({ width: val });
+                            });
+                        }
+                        return d.resolve();
```
kdeloach: Until we add animated transitions, you could simulate the transition delay by doing something like this:

```
var d = new Deferred();
setTimeout(function() {
    // Set width here
    d.resolve();
}, 1000);
return d.promise;
```

src/GeositeFramework/sample_plugins/identify_point/main.js
mmcfarland: I like this idea - I found the blink stuff a little complex.  We also don't need to show every option, but just the basic api usage - we'll have documentation in the wiki that will be more comprehensive.  I agree that it should be reduced and simplified.

src/GeositeFramework/sample_plugins/identify_point/main.js
```diff
@@ -31,6 +39,55 @@ define(["dojo/_base/declare", "framework/PluginBase"],
...
+                            });
+                            break;
+                        case "resize-ctl-identify-large":
+                            $(el).on("click", function() {
+                                self.app.resize.setWidth("large").then(blink("blue", 5));
+                            });
+                            break;
+                        case "resize-ctl-identify-small":
+                            $(el).on("click", function() {
+                                self.app.resize.setWidth("small").then(blink("red", 5));
```
jtarricone: `small` and `large`, at least, are already defined.

mmcfarland: Yeah, you're good with those two classes + the numeric.

jtarricone:
I made the example in the Identify plugin simpler and touched up the implementation a bit.

src/GeositeFramework/js/Plugin.js
```diff
@@ -32,7 +32,35 @@ require(['use!Geosite',
...
+                        _.delay(function() {
+                            if (typeof val === "string") {
+                                d.then(setter(val));
+                            } else if (typeof val === "number") {
+                                d.then(function() {
+                                    $uiContainer.css({ width: val });
+                                });
+                            }
+                            d.resolve();
+                        }, 1000);
```
mmcfarland: I understand the comment to add this, but I don't think we need to actually add an arbitrary delay of one second.  Let's cut this down to `100`, so that the class has been applied but not incur a noticeable lag.

kellyi:
I'm testing out this branch in the browser and the identify point resizing seems to work well. There is a slight but noticeable delay in the demo version between when the text updates to say "Width is now large" etc and when the plugin actually does resize -- it seems like maybe half a second between clicking and the resize happening? Wasn't sure whether that was intended behavior here. (Possibly seems related to the comment @mmcfarland just made above).

Also, may be worth considering having the change call `resize(immediate)` on the Esri map where `immediate = true` if we want to get rid of the map flicker on resizing as was done here: https://github.com/CoastalResilienceNetwork/GeositeFramework/pull/812 . The Esri map object has an internal/built-in delay on resizing which can be circumvented be calling `resize(true)`; otherwise it will lag behind the plugin resize.

src/GeositeFramework/sample_plugins/identify_point/main.js
```diff
@@ -16,6 +16,14 @@ define(["dojo/_base/declare", "framework/PluginBase"],
...
                     '<h4 style="padding: 5px;">' + i18next.t('Click any point on the map to display Latitude and Longitude') + '</h4>');
 
+                // add buttons to bind resize events to
+                $(this.container).append('<p>The buttons below demonstrate how to use a plugins "resize" method</p>')
+                    .append('<button class="resize-btn" data-ui-key="resize-ctl-identify-set450">Set to 450</button>')
+                    .append('<button class="resize-btn" data-ui-key="resize-ctl-identify-small">Small</button>')
+                    .append('<button class="resize-btn" data-ui-key="resize-ctl-identify-large">Large</button>')
+                    .append('<p class="width-textbox"></p>');
+
+
```
kellyi: Extra whitespace here.

jtarricone:
I reduced the time delay and added an `esriMap.resize(true)` invocation, which does really reduce the map flicker.

src/GeositeFramework/sample_plugins/identify_point/main.js
kdeloach: Cool, didn't realize those styles were already defined.

PR #805, opened 2016-12-21
Standardize icon sizes

mmcfarland:
TNC has created icons at 56x56 for all of their apps, these bring our
local icons to the same size so they are more consistent.

#### Testing
After a hard refresh, ensure that the icons are still visible for the launchpad and the sample identify.


jtarricone:
I don't think this branch is the cause, but I noticed that Firefox renders the buttons differently than Chrome or IE:

Firefox:
index

Chrome/IE:
download 2

Is there an in-progress issue that might address that or should a new one be added?

mmcfarland:
An issue for that was just recently created #809

Issue #784, opened 2016-12-15
Update unit display for measure


mmcfarland:
The measure tool offers display units in two spots, while the drawing is in progress:
![screenshot from 2016-12-15 11 23 18](https://cloud.githubusercontent.com/assets/1014341/21232209/e52f0596-c2b8-11e6-88c2-669743622419.png)

and after the measure is complete:
![screenshot from 2016-12-15 10 22 19](https://cloud.githubusercontent.com/assets/1014341/21231972/13d35c40-c2b8-11e6-81dc-cae72f62af90.png)

These should be updated to also include the appropriate metric unit as well

* miles/kilometers
* mi2/km2, 
* acres/hectares

kdeloach:
Just double checking-- Do we want to replace imperial units with metric? Or display both?

mmcfarland:
We'll display both.

PR #806, opened 2016-12-21
Add more UoM and associated values to measure tool

jtarricone:
Adds metric equivalent measurements to the output of the measuring tool, for length and area. Also adds acres/hectares to the area measurement.

To test:
* clone this branch and open the app
* activate the measuring tool
* when measuring, check that the metric equivalent value is displayed and is accurate
* check that the segment- and total-length measurements look correct, i.e. the total >= segment
* create a polygon and check that the area is now shown in square miles/kilometers and acres/hectares

Connects #784

### DIFF
--- a/src/GeositeFramework/Views/Home/Index.cshtml
+++ b/src/GeositeFramework/Views/Home/Index.cshtml
@@ -558,25 +558,44 @@
         <div class="measure-tooltip">
             <p><%= instructions %></p>
             <div>
-                <div class="measure-label"><%= segmentLabel %>:</div><%= segmentLength %> <%= units%>
+                <div class="measure-label"><%= segmentLabel %>:</div>
+                <% _.each(segmentLengths, function (item, index) { %>
+                    <div><%= item.length %> <%= item.uom %></div>
+                <% }); %>
             </div>
             <div>
-                <div class="measure-label"><%= lengthLabel %>:</div><%= totalLength %> <%= units%>
+                <div class="measure-label"><%= lengthLabel %>:</div>
+                <% _.each(totalLengths, function (item, index) { %>
+                    <div><%= item.length %> <%= item.uom %></div>
+                <% }); %>
             </div>
         </div>
     </script>
 
     <script type="text/template" id="template-measure-infobubble">
         <div class="measure-bubble">
-            <% if (area) { %>
+            <% if (areas) { %>
                 <p>
                     <span class="measure-label i18n" data-i18n="Area:">Area:</span>
-                    <span><%= area %> <%= areaUnits%><sup>2</sup></span>
+                    <div>
+                        <% _.each(areas, function (item, index) { %>
+                            <span>
+                                <%= item.area %> <%= item.uom %>
+                                <% if (item.uom === "mi" || item.uom === "km") { %>
+                                    <sup>2</sup>
+                                <% } %>
+                            </span>
+                        <% }); %>
+                    </div>
                 </p>
             <% } %>
             <p>
                 <span class="measure-label i18n" data-i18n="Length:">Length:</span>
-                <span><%= length %> <%= lengthUnits%></span>
+                <div>
+                    <% _.each(lengths, function (item, index) { %>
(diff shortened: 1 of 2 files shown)


src/GeositeFramework/js/widgets/map_utils/measure/AgsMeasure.js
```diff
@@ -210,14 +211,32 @@ define([
                     dojo.connect(_pointLayer, 'onDblClick', handleMarkerDblClick);
             },
 
+            getLength = function(line) {
+                return _.map(options.esriLengthUnits, function (u) {
```
mmcfarland: I presume `u` is representing a unit variable, it would be easier to read and comprehend with a full variable name like `unit`.

src/GeositeFramework/js/widgets/map_utils/measure/AgsMeasure.js
```diff
@@ -261,10 +284,21 @@ define([
 
                 // Calculate the length of the line, using a geographic coordindate system
                 geographicLine = webMercatorUtils.webMercatorToGeographic(line);
-                var geoLineLength = geodesicUtils.geodesicLengths([geographicLine], options.esriLengthUnits)[0];
+                var geoLineLength = getLength(geographicLine);
 
                 // Format the segment and line lengths
-                tipText = formatTooltip(geoLineLength, _renderedLength + geoLineLength);
+                var totalLength = _.map(
```
mmcfarland: I find this block pretty terse and hard to follow.  Breaking it down to component parts is about equivilent length but much better to read, in my opinion.  Along the lines of
```js
                var units = _.map(options.esriLengthUnits, function(unit) {
                        return _unitsLookup[unit];
                    }),
                    allSegments = _renderedLength.concat(geoLineLength),
                    totalLength = _.map(units, function(unit) {
                        return {
                            uom: unit,
                            length: sum(_.where(allSegments, { uom: unit }))
                        };
                    });
                function sum(arr) {
                    return _.reduce(arr, function (x, y) { return x + y.length; }, 0)
                }
```

src/GeositeFramework/Views/Home/Index.cshtml
```diff
@@ -554,25 +554,44 @@
...
 
     <script type="text/template" id="template-measure-infobubble">
         <div class="measure-bubble">
-            <% if (area) { %>
+            <% if (areas) { %>
                 <p>
                     <span class="measure-label i18n" data-i18n="Area:">Area:</span>
-                    <span><%= area %> <%= areaUnits%><sup>2</sup></span>
+                    <div>
+                        <% _.each(areas, function (item, index) { %>
```
mmcfarland: The results look correct, but they're hard to read in that popup and the delete button is obscured.  Perhaps we should display them stacked vertically and just drop the "length" for area, which I assume means perimeter (if they want to measure lenght, let them draw a line).  Then we could move the delete button up to the right side, or at least make the popup taller so that it's visible.

jtarricone: I left the styling alone since I didn't see any wireframes or other guidance for how the box should be laid out. I thought that  a "table" with two columns, one for imperial and one for metric, might be a decent layout?

jtarricone: I looked into updating the tooltip, and it looks like it involves modifying a couple CSS files, possibly in addition to a code change somewhere. The `height` rule on the tooltip itself isn't enough to fix the problem, since there are rules for the container `div` that also have to be updated. That `div` is within the dijit info window, and it's dimension appear to be specific to the element rather than derived from a specific stylesheet, so it seemed like resizing the box might involve enough untangling to be handled separately.

src/GeositeFramework/js/widgets/map_utils/measure/AgsMeasure.js
jtarricone: I refactored this function bit by splitting the functions and values into separate pieces to make the sequence a bit clearer.

Issue #780, opened 2016-12-15
Smooth transition when closing plugins


mmcfarland:
When the width of the map is adjusted to accommodate the space required or relinquished by toggling the state of a sidebar plugin, the map animates very abruptly, usually leaving a momentary black block along the right border.  The esri map controls also jump away from the fake map controls that aren't actually contained in the map div.

We should animate the width transition so this is a more visually appealing experience.

kellyi:
I did a bit of investigation for this one and it looks like it's possible to use a keyframe animation to set `width: 0px` and `visibility: hidden`  to the plugin sidebar over a 500ms transition as it's hiding. However, it seemed like the underlying map didn't update anymore quickly when I made that change, and so it remained noticeable.

PR #812, opened 2016-12-22
Resize Esri map immediately on hiding plugin

kellyi:
## Overview

This PR adds code to force the Esri basemap to resize right away on closing a sidebar plugin. It works by calling `.resize(true)` on the Esri map object, which causes it to resize immediately: https://developers.arcgis.com/javascript/3/jsapi/map-amd.html#resize

It looks like the Esri map has a built-in delay otherwise. Here's a note from the Esri docs for the `resize` method:

> By default, the actual resize logic is delayed internally in order to throttle spurious resize events dispatched by some browsers. In cases where you explicitly call this method in a one-and-done situation, you can use the boolean immediate argument to indicate that the resize logic should be applied immediately without any delay.

Here we code used to hide the plugin to get a reference to the Esri map object, then call `resize` after calling `hide` on the plugin. There are also some guards there to make sure the map object exists before trying to call `resize`.

## Notes

This seems to fix the transition in most cases. I toggled plugins on and off and noticed the lagging map for an instant maybe 1/5 or 1/10 of the time. However, when it does happen it's less prominent and most instantaneous than the map flicker on http://lr01.internal.azavea.com:81/HUDSON_TNC_NJ_Spanish_200/

## Testing

 * get and rebuild this branch in VS, then launch in the browser and open the browser console
 * toggle plugins on and off and verify that the map flicker has been noticeably reduced. Compare it with http://lr01.internal.azavea.com:81/HUDSON_TNC_NJ_Spanish_200/

Connects #780


kellyi:
Wrote this out then realized we probably should resize the map immediately on showing plugins, too, as the current delay there means that the zoom controls get pushed off the screen temporarily until it re-renders on opening a plugin. Will push a second commit to address that.

src/GeositeFramework/js/Plugin.js
```diff
@@ -383,14 +384,20 @@ require(['use!Geosite',
...
-            } else {
-                if (view.$uiContainer) {
+                } else {
                     view.$uiContainer.hide();
                 }
+                // Call the `resize` method on the Esri map, passing it
+                // true for its `immediate` arg so that it will resize
+                // immediately.
+                if (pluginObject.app) {
+                    if (pluginObject.app._unsafeMap) {
```
kellyi: Sure, I'll try that instead.

kellyi:
Thanks! I rebased on `prototype`, squashed the fixup commit, and I'm going to merge.

Issue #771, opened 2016-12-13
UI issues on tablet


ajrobbins:
#### Instructions
In the comments of this issue are a list of known problems with various tablets.  The first step would be to apply the output of the autoprefixer mentioned, to see what that resolves by itself.  The goal will be to check off as many of the items in this list as possible.

kdeloach:
Verify that legend can be minimized with Safari on iPad.

zferdana:
Legend can't be minimized on iPad in Safari confirmed

jfrankl:
After some testing, I have found that autoprefixer will solve that initial "Browserstack with Chrome on iPad Air" issue, Arianna.

The issue seems to be that some Browserstack devices use older versions of browsers or operating systems. When you get back to iOS 8, our site breaks. If we use autoprefixer to support the last 3 versions, it works.

Before autoprefixer:

![image](https://cloud.githubusercontent.com/assets/1809908/21274360/cd58a2c8-c395-11e6-9bfd-eb0dfe30ffac.png)

After autoprefixer:

![image](https://cloud.githubusercontent.com/assets/1809908/21274371/e02d930e-c395-11e6-9f7c-1eee485acb3c.png)

ajrobbins:
Good find, thanks @jfrankl !

kdeloach:
I will try to identify which styles are generated from the CSS autoprefixer and apply them manually.

kdeloach:
Jeff suggested using https://autoprefixer.github.io/ to add missing styles.

kdeloach:
Here are the browsers we are targeting for this iteration:

* iPad 2
* iPad 3
* iPad Air
* iPad Air 2
* Nexus tablet (Safari, Chrome)

ajrobbins:
Issues noted on iPad 3 running iOS 9.3.2, using Safari:
- [ ] Not seeing any layers in regional planning
- [ ] ... button icon is off-center. This may not matter with the new wrench icon, but I didn't see that on the ny-prototype.
![img_0091](https://cloud.githubusercontent.com/assets/12868342/21370985/b512e260-c6dc-11e6-890d-572029104451.jpg)

- [ ] Tooltip with app name on hover doesn't work with tablet; the tooltip hangs around, blocking the sidebar:
![img_0092](https://cloud.githubusercontent.com/assets/12868342/21370996/cf4ec0d6-c6dc-11e6-874e-6c067f11a16a.jpg)

[...]

PR #818, opened 2016-12-27
Update plugin list sidebar behavior on tablets

kellyi:
## Overview

On viewports narrower than 991px (which included iPads in portrait mode), the prototype had shown only plugin icons along with a not-straightforward-to-dismiss tooltip pointing to the icon when one of the buttons had been tapped. 

This PR removes the tooltips altogether and updates the plugin list sidebar behavior for viewports < 991px:

- when a plugin is open and displaying content in its sidebar container, we now show only the icons in the plugin list sidebar
- when no plugins are displaying content in a sidebar container, we now show the icons and plugin titles in the sidebar

This works by toggling a CSS class on the `nav-app` class, which houses the plugin list. When plugin content is visible, `nav-app` now gets a `nav-app-narrow` class -- which is removed when no plugin content's visible. That class has rules within a `@media (max-width: 991px) {` CSS rule which hides the text for the plugin buttons.

## Screenshots

img_0100

img_0106

img_0105

## Notes

To see selectable layers in regional-planning, make sure you've got the most recent commits from the plugin's `prototype` branch.

## Testing
- rebuild this branch in VS, then launch in the browser
- use the app as usual in your browser and verify that everything still works as before on a viewport > 990 px;
- use ngrok, iisexpress-proxy, or similar to make your version available to an iPad, then connect and check the site in both landscape and portrait mode
- in landscape mode, verify that everything still works as it did before and that the plugin buttons display the titles
- in portrait mode, verify that the plugin list sidebar now shows only icons when a plugin's content is visible, shows icons and titles when no plugin content is visible, and does not show any tooltips on taps
- try to get the plugin list sidebar out of sync with whether plugin content is visible. Do this by trying combinations of tapping a plugin to open it, selecting another one, selecting another one, closing/minimizing then reopening plugins in various sequences. Verify that the show/hide plugin titles behavior continues to work correctly.

Connects #771

### DIFF
--- a/src/GeositeFramework/js/Plugin.js
+++ b/src/GeositeFramework/js/Plugin.js
@@ -167,6 +167,7 @@ require(['use!Geosite',
                 startHelpKey: "-suppress-help-start",
                 pluginObject: null,
                 active: false,
+                visible: false,
                 pluginLayers: {},
                 pluginLayersVisible: true
             },
@@ -192,10 +193,18 @@ require(['use!Geosite',
             onSelectedChanged: function () {
                 if (this.selected) {
                     this.set('active', true);
+                    this.set('visible', true);
                     this.get('pluginObject').activate(!this.getSuppressHelpOnStartup());
                 } else {
                     this.get('pluginObject').deactivate();
+                    this.set('visible', false);
                     this.trigger('plugin:deselected');
+                    // Remove the `.nav-apps-narow` class if the
+                    // de-selecting the plugin has left no plugins
+                    // visible.
+                    if (this.collection.all({visible:false})) {
+                        $('.nav-apps').removeClass('nav-apps-narrow');
+                    }
                 }
             },
 
@@ -239,6 +248,7 @@ require(['use!Geosite',
                     // new status via the hibernate function
                     doDeactivate = function() {
                         self.set('active', false);
+                        self.set('visible', false);
                         self.get('pluginObject').hibernate();
                     }
 
@@ -409,7 +419,9 @@ require(['use!Geosite',
                     selected: model.selected || model.get('active'),
                     displayed: model.selected,
                     fullName: model.get('pluginObject').fullName
-                }));
+                })),
+                sidebarNavSizedForTablet = !!$('.nav-apps-narrow'),
+                pluginContentHidden = model.collection.all({visible:false});
 
             view.$el.empty().append(html);
             view.$el.addClass(model.getId() + '-' + view.paneNumber);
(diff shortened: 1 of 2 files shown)


kellyi:
Probably also worth ensuring that the plugin list sidebar updates correctly when the size of the viewport changes, as in when a tablet's turned from portrait to landscape mode. The expected behavior is that it should be sized correctly automatically on turning to landscape, as the `media(max-width)` rule will no longer apply.

src/GeositeFramework/js/Plugin.js
```diff
@@ -196,6 +196,14 @@ require(['use!Geosite',
                 } else {
                     this.get('pluginObject').deactivate();
                     this.trigger('plugin:deselected');
+                    // Remove the `.nav-apps-narow` class if the
+                    // de-selecting the plugin has left no plugins
+                    // active.
+                    if ($('.sidebar').filter(function() {
```
mmcfarland: I don't think we should be using the CSS properties for state in the app.  Instead, you can check if all plugins are  not `active`:
```js
if (this.collection.all({active:false})) {
    $('.nav-apps').removeClass('nav-apps-narrow');
}
````

File: src/GeositeFramework/js/Plugin.js [hunk 2]
```diff
@@ -409,17 +417,28 @@ require(['use!Geosite',
                     selected: model.selected || model.get('active'),
                     displayed: model.selected,
                     fullName: model.get('pluginObject').fullName
-                }));
+                })),
+                sidebarNavSizedForTablet = !!$('.nav-apps-narrow'),
+                pluginContentHidden = $('.sidebar').filter(function() {
```
mmcfarland: I _believe_ this is referring to all plugins, not just "this" plugin.  If so you, you can use the same model state from the previous comment, but you'll access the collection with `view.model.collection`.

src/GeositeFramework/js/Plugin.js
```diff
@@ -443,6 +462,17 @@ require(['use!Geosite',
             return view;
         }
 
+        function toggleTabletSidebarPluginText(visible) {
```
mmcfarland: `toggle` implies that you can invert the current setting.  If this really requires a bool arg, I'd rename it to `setSideBarPluginTextVisibility`, and drop the `Tablet` since that really hasn't got anything to do with the function implementation.

src/GeositeFramework/js/Plugin.js
kellyi: Much better -- I'll make that change here and in line 422 referred to in the next comment.

kellyi:
Pushed a fixup which addresses the requested changes. I updated the function name to make it more clear what it was going to do, and adjusted the checks to read model/collection state rather than reading from the UI. 

However, it seemed like `active` wasn't quite the thing I needed to check on, as it seems like plugins which have been minimized yet are still on remain `active`. I think what it should be checking is whether any plugins are `visible`, and so I added that as part of the model state, then set and updated it accordingly. This may be already available in another way and, if so, it'd supplant needing to add a new key to keep track of in the state.

kellyi:
Ah, that makes sense. I checked out picky and it looks like the repo's README says it's deprecated and that people should consider using Backbone Select instead. I'll leave the `visible` property as part of the state here.

Thanks for your feedback on this! Going to squash & merge.

kellyi:
Squashed and rebased on `prototype`.

PR #819, opened 2016-12-28
Center map-utils wrench icon on iOS Safari

kellyi:
## Overview

This PR adds a CSS rule to center the wrench icon used on the button for the map-utils tool. Previously, the icon had appeared centered on desktop browsers, but it was off center in iOS Safari. The changes here add rules to position it 50% of the way horizontally across its parent div.

## Screenshots

*IE 11 on Windows:*
screen shot 2016-12-28 at 10 54 22 am

*Chrome on OS X:*
screen shot 2016-12-28 at 10 54 34 am

*Safari on iPad:*
img_0107

*Safari on OS X:*
screen shot 2016-12-28 at 10 58 54 am

## Notes

In addition to the browsers with screenshots above, I also checked this out in Firefox on Windows 7 -- and it worked.

## Testing
 * rebuild this branch in VS, then serve and view it in a few desktop browsers to ensure the icon's centered
 * use ngrok or iisexpress-proxy or similar to make the app available to an iPad, then visit the site and confirm that the icon's centered there, too

Connects #771

### DIFF
--- a/src/GeositeFramework/css/main.css
+++ b/src/GeositeFramework/css/main.css
@@ -2341,3 +2341,7 @@ body.x-body {
 .sidebar-content > h2  {
   margin-top: 1rem;
 }
+#map-utils-dropdown-button > i:before {
+  margin-left: 50%;
+  transform: translate(-50%, 0%);
+}


src/GeositeFramework/css/main.css
```diff
@@ -2335,3 +2335,7 @@ body.x-body {
 .launchpad-plugin li p {
   font-size: 13px;
 }
+.icon-wrench:before {
```
jfrankl: I'd make this selector more specific, perhaps referencing the parent button. That way, if someone uses this icon elsewhere in the application, it won't be impacted by this style that is made for centering on a button.

kellyi: Sounds good. I pushed a fixup which adds the id and classname from the parent button to ensure we're only applying the rule here.

src/GeositeFramework/css/main.css
```diff
@@ -2335,3 +2335,7 @@ body.x-body {
 .launchpad-plugin li p {
   font-size: 13px;
 }
+.btn-default.dropdown-toggle#map-utils-dropdown-button > .icon-wrench:before {
```
rajadain: Consider using the selector:

```css
#map-utils-dropdown-button > i:before
```

Since ids are unique, they need no further qualification. By targeting `i` rather than `.icon-wrench`, we ensure that the placement will stick even if the specific icon is changed down the line.

kellyi: Sounds good to me! I just rebased this on `prototype` then pushed a fixup which uses that selector to get the icon.

kellyi:
Thanks! I just squashed the two fixup commits and force pushed the updated branch. Going to merge.

Issue #809, opened 2016-12-22
Sidebar text wraps under icons in Firefox


ajrobbins:
Observed with Firefox 50 on Mac (Sierra) at full screen width (13inch)
![screen shot 2016-12-22 at 10 37 00 am](https://cloud.githubusercontent.com/assets/12868342/21430900/2d4ed26e-c833-11e6-9615-df8f27e21c0f.png)

PR #821, opened 2016-12-28
Add `float: left` to plugin list navbar button icons

kellyi:
## Overview

This PR adds a `float: left` rule to `.nav-apps-button>.button-icon` so that the span will render to the left of the plugin title text on the same line in Firefox.

## Screenshot

*Firefox on OS X*
screen shot 2016-12-28 at 2 02 17 pm

## Notes

Checked this in OS X Safari, Windows IE11, and OS X Chrome -- where the icon and text were already horizontally aligned -- and they plugin icons and text remain horizontally aligned.

## Testing
 * rebuild this branch in VS, open the project in the browser and verify that the plugin icon and text are horizontally aligned
 * check in Firefox, Chrome, Safari, and IE11

Connects #809

### DIFF
--- a/src/GeositeFramework/css/main.css
+++ b/src/GeositeFramework/css/main.css
@@ -986,6 +986,7 @@ nav {
     font-size: 18px;
     position: relative;
     top: 1px;
+    float: left;
 }
 .nav-apps-button > .button-icon img {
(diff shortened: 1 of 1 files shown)


kellyi:

Issue #808, opened 2016-12-22
Can't scroll Launchpad in IE 11 or Firefox


ajrobbins:
Observed in IE 11 and Firefox 50 on Windows 10

kellyi:
This is also happening on Firefox for OS X but not Safari or Chrome for OS X, which probably means webkit vs non-webkit browsers.

kellyi:
I believe this occurs because the launchpad plugin is missing a `height` property -- it seems like we do set that in regional-planning, which enables it to scroll on Firefox and IE11. I tried out `height: 80vh` so far and it worked to force the scroll but didn't look quite right, so I'm going to test out some other values.

kellyi:
`height: 100%` turns out to be too large, as it looks like the value's 100% of the viewport height and then the launchpad pushes below the bottom margin without being scrollable. However, if you hit the space bar in FF with the height set that way, it will scroll the plugin content bar and the entire map down to the bottom such that the bottom of the plugin will be visible.

mmcfarland:
Re: height, is something like this appropriate?  https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/prototype/src/GeositeFramework/css/main.css#L938

kellyi:
Yeah! Looks like `height: calc(90vh - 60px);` seems to work.

ajrobbins:
I just observed this again on Firefox and IE 11, and TNC reported it in IE11 as well.

kellyi:
I believe @mmcfarland just made a fix to the regional-planning `prototype` branch's scroll behavior that might potentially work here, too.

lewfish:
This also seems to happen on Chrome in Windows.

mmcfarland:
Don't mean to poke my nose in ongoing work, but this turns out to be the same problem I had with regional planning.  Update this particular style should resolve it:

```css
.launchpad-plugin {
    padding: 10px;
    height: 100%;
    overflow: auto;
}
```

lewfish:
This fixes it on Chrome, but not IE11. It looks like the PR I did to make the layer info box visible is causing further issues than can be fixed with this patch.

lewfish:
I think I know how to fix it now.

PR #822, opened 2016-12-28
Set height for launchpad plugin to enable scrolling in IE and Firefox

kellyi:
## Overview

This PR adds a computed `height` rule to the launchpad plugin -- and makes it so that it's scrollable in Firefox and IE11.

Previously the plugin's content div was missing a `height` rule, and without it the plugin seemed to expand to fit the height of the content, which meant that the bottom of the map was also pushed off the screen while users were still unable to scroll. (Tapping the spacebar would scroll the plugin content and the entire map down, and on initially loading you wouldn't be able to see the map attributions.)

## Notes

Adding `height: calc(90vh - 60px);` to the `launchpad-plugin` class reduces the div's size such that the existing `overflow: auto` rule takes effect and compels it to scroll within the div. `60px` accounts for the `40px` margin set on the `sidebar-content` div along with the `20px` total of padding above and beneath the `launchpad-plugin`. `90vh` was picked based on what looked best in a variety of browsers.

## Testing
 * check Chrome, IE, and Firefox at a variety of screen heights to verify that the plugin's scrollable when it should be and renders fully when the viewport's tall enough to accommodate it
 * also check that it remains scrollable on closing it, then reopening it. On initially loading, the `launchpad-plugin`'s `sidebar` div does not get the `style="display: block;"` rule used when toggling plugins visible and not, so a test should ensure that it scrolls with and without that

Connects #808

### DIFF
--- a/src/GeositeFramework/plugins/launchpad/style.css
+++ b/src/GeositeFramework/plugins/launchpad/style.css
@@ -1,5 +1,6 @@
 .launchpad-plugin {
     padding: 10px;
+    height: calc(90vh - 60px);
 }
 .launchpad-plugin .plugins div,
 .launchpad-plugin .scenarios div {


kellyi:
Thanks!

Issue #823, opened 2016-12-29
Integrate styleguide


mmcfarland:
@jfrankl has developed some material to serve as a visual styleguide, with additional styling assets intended to be incorporated into the framework so that app developers can easily use the new styles.

PR #824, opened 2016-12-29
Add styleguide

mmcfarland:
#### Overview
The styleguide is a mock framework index page with a mock plugin demonstrating various markup and style components.  Its source is sass but it has compiled css assets that are included directly into the framework.

The styleguide can be viewed via a webserver at `/styleguide`, but I also created a gh-pages branch with the assets so it can be viewed at http://coastalresiliencenetwork.github.io/GeositeFramework/.  I've added that link to the wiki documentation, as well.

Connects #823 

#### Testing
Check the framework for style-collisions or areas where new styles should be applied (though not specifically in plugins).  I fixed up "Create Map" to use the new form elements and I'm going to take a pass at the Regional Planning repo to see if I can update that a bit, too.


kellyi:
Taking a look!

kellyi:
It won't let me comment on the specific file because it's a binary, but consider adding `.DS_Store` to `.gitignore` to skip adding this: 

https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/b226e0037b680314621fe46c4fe777735ef63e41/styleguide/assets/fonts/fontello/.DS_Store

kellyi:
Sample plugin and GitHub pages page both look good!

styleguide/assets/sass/main.scss
```diff
@@ -0,0 +1,45 @@
+/* * * *
+ * Utilities
+ * No css should output from these
+ * * * */
+@import
+  'utils/functions',
+	'utils/variables',
+	'utils/mixins',
```
kellyi: Extra whitespace.

styleguide/assets/sass/main.scss
```diff
@@ -0,0 +1,45 @@
...
+/* * * *
+ * Vendors imports
+ * * * */
+// @import Placeholder
+
+/* * * *
+ * Base boilerplate styling
+ * * * */
+@import
+	'base/base';
```
kellyi: Extra whitespace.

mmcfarland:
Thanks, made those small fixups but left the scrolling issue.  Going to consult design about best overall approach.

Issue #828, opened 2017-01-03
Create experimental JSAPI 4.0 branch


mmcfarland:
We'll be doing some work that doesn't have an explicit path to be merged into master, based off of the prototype branch.  This will involve upgrading the ESRI API to the most recent 4.x series and getting it working "basically", where "basically" will mean major features will work (map, adding layers), but it does not need to be universally complete.  In fact, the API is not even feature complete with the 3.x series, so it may not be possible to.

kellyi:
Fork is here: https://github.com/CoastalResilienceNetwork/GeositeFramework/tree/js-api-4x-pilot

kellyi:
It appears that the JS Web Optimizer is not available for building bundles using the 4.x API, so I'm going to start out by including the JS & CSS from a CDN.

kellyi:
Upgraded to 4.2 using the CDN and there were about 15 immediate issues related to breaking API changes. I was able to resolve the following issues by renaming imports or adjusting directory paths:

- `esri/dijit/Popup` is now imported from `esri/widgets/Popup`
- `ArcGISTilerMapServiceLayer` is now `TileLayer`
- `esri/map` is now `esri/Map`
- `esri/SpatialReference` is now imported from `esri/geometry/SpatialReference`
- `esri/InfoTemplate` is now `esri/PopupTemplate`
- `esri/graphic` is now `esri/Graphic`
- `esri/dijit/Search` is now imported from `esri/widgets/Search`
- `ArcGISDynamicMapServiceLayer` is now `MapImageLayer`
- `esri/tasks/IdentifyParamter` is now imported from `esri/tasks/support/IdentifyParameter`
- `esri/geometry/webMercatorUtils` is now imported from `esri/geometry/support/webMercatorUtils`

The following errors are going to take slightly more work to resolve as they're related to features not yet implemented in 4.x or features which aren't going to be implemented in 4.x per this reference: https://developers.arcgis.com/javascript/latest/guide/functionality-matrix/index.html

- `esri/dijit/Scalebar` is "Coming soon"
- `esri/geometry/geodesicUtils` is "Coming soon"
- `esri/geometry/mathUtils` is "Not planned"
- `esri/units` is "Not planned", although the API docs suggest an alternate way to handle it
- `estri/dijit/InfoWindow` is "Not planned"; the API docs recommend using `View.popup` instead

Going to pull in the regional planning plugin and see if it has any additional errors.

[...]

PR #831, opened 2017-01-09
Create banch to upgrade to Esri ArcGIS JS API v4.2

kellyi:
## Overview

This PR begins the work of creating a new branch to test out upgrading the Esri ArcGIS JavaScript API from v3.x to v4.2. I've already created a new branch off `prototype` at `js-api-4x-pilot`, and this branch starts the process of upgrading the framework to accommodate the new API.

I tried to have each commit here address only one or two changes for the upgrade. Some of the adjustments include:

- updating names or import paths to correspond to the new names in the v4.2 API
- commenting out (and writing searchable text) for features which aren't or aren't yet supported
- setting up a new basemap, along with the related CSS rules to mount it and adjust how the basemap layer selector works
- splitting the `Map` and `MapView` in some cases to match the new API. The new API uses the 'Map' as a store for data, and the 'MapView' as the mechanism for handling the UI. Here I've done some preliminary work to split those, but we'll probably need up update the model more thoroughly to handle it.
- adjusting the `MapWrapper`'s calls to toggle layers to use the API's new method names

## Screenshots

The app currently looks like this, and includes a working basemap selector along with partially-working layers in the `regional-planning` plugin (from a PR here: https://github.com/CoastalResilienceNetwork/regional-planning/pull/75)

screen shot 2017-01-11 at 5 48 39 pm

screen shot 2017-01-11 at 5 48 51 pm

screen shot 2017-01-11 at 5 49 36 pm

screen shot 2017-01-11 at 5 49 51 pm

## Notes

Only some of the layers from `regional-planning` are working thus far, including the Inundation Depth tile layer and some of the Hurricane Sandy Layers. The other layers are making requests to the map services when added, but they're currently grabbing the incorrect sublayers. Failed layers seem to try for index `-1`, and the Hurricane Sandy County impact seems to be grabbing sublayers from indices `[0, 2]`. 

I'm inclined to make another card to handle this, as I'm not certain yet whether it needs to be handled at the level of the plugin or at the level of the framework.

I've catalogued a bunch of additional notes here:

https://github.com/CoastalResilienceNetwork/GeositeFramework/issues/828

and there's a PR which needs to be reviewed in tandem with this one here:

https://github.com/CoastalResilienceNetwork/regional-planning/pull/75

## Testing
- get this branch along with the similar branch from https://github.com/CoastalResilienceNetwork/regional-planning/pull/75
- rebuild in Visual Studio and open in the browser
- verify that you see a map and that the plugins load without error
- select different basemaps from the dropdown and verify that they render
- toggle `regional-planning` on and select the Inundation Layers and the Hurricane Sandy Layers and verify that they render on the map
- toggle the other layers and verify that you see the requests being made successfully -- even if the images are currently empty

Connects #828


src/GeositeFramework/js/widgets/map_utils/main.js
```diff
@@ -1,7 +1,14 @@
 ﻿define([
     'use!Geosite',
     'dojo/_base/declare',
+    /*
+    Disable measure tool as it includes some features
+    not yet implented in Esri JS API 4.2, specifically:
```
mmcfarland: implented -> implemented

src/GeositeFramework/js/Map.js
```diff
@@ -33,14 +36,10 @@ require(['use!Geosite',
...
-        // Return an ESRI layer object for the currently-selected basemap spec
+        // Add the selected basemap to the Esri map
         var basemap = getSelectedBasemap(model);
-        if (basemap.layer === undefined) {
-            // This basemap has no layer yet, so make one and cache it
-            basemap.layer = new TileLayer(basemap.url);
-            esriMap.addLayer(basemap.layer);
-        }
-        return basemap.layer;
+        basemap.layer = new TileLayer(basemap.url);
```
mmcfarland: Is there a reason to dispose of the mechanism to cache the layer creation, it seems like the only substantial change here is to change `esriMap.addLayer` to `esriMap.add`, or is there more going on?

src/GeositeFramework/js/Map.js
```diff
@@ -101,20 +100,24 @@ require(['use!Geosite',
...
+            if (!_.isEqual(currentExtent, view.esriMapView.extent)) {
                 loadExtent(view);
             }
         });
 
+        /*
         // Configure the esri proxy, for (at least) 2 cases:
         // 1) For WMS "GetCapabilities" requests
         // 2) When it needs to make an HTTP GET with a URL longer than 2000 chars
         esri.config.defaults.io.proxyUrl = "proxy.ashx";
```
mmcfarland: This got effectively commented out, but it's hard to see because of mixed comment style.  If it was unintentional, you may need to restore the proxy setting, or if it was meant to be commented out, can you convert to just one style so that it's easy to catch and put in the searchable comment text about the support in 4.2 so we can track it?

src/GeositeFramework/js/Map.js
```diff
@@ -101,20 +100,24 @@ require(['use!Geosite',
...
-
+        */
         createMap(view);
     }
 
     function createMap(view) {
+        /*
+        // Reworking this because the Esri JS API 4.2
```
mmcfarland: Nest comment styles here, too.

src/GeositeFramework/js/Map.js
```diff
@@ -203,11 +213,13 @@ require(['use!Geosite',
         // the app entirely.  The map does, in fact, load and its loaded property is
         // set.  I put in this hack to check up on the event a little while after
         // it was created and manually raise the event if the library didn't do it.
+        /*
```
mmcfarland: This reads like an IE8 specific hack.  Could you add a TODO comment to delete once we confirm that it's not useful anymore?

src/GeositeFramework/js/Map.js
```diff
@@ -130,8 +133,15 @@ require(['use!Geosite',
                     }
             }, 300),
             loadEventFired = false;
+        */
```
mmcfarland: Does the resize happen transparently now?

src/GeositeFramework/js/Map.js
```diff
@@ -216,22 +228,15 @@ require(['use!Geosite',
...
     function selectBasemap(view) {
-        // Hide the current basemap layer
-        if (view.currentBasemapLayer !== undefined) {
-            view.currentBasemapLayer.hide();
-        }
-        // Show the new basemap layer (at index 0)
-        view.currentBasemapLayer = view.model.getSelectedBasemapLayer(view.esriMap);
-        view.currentBasemapLayer.show();
-        view.esriMap.reorderLayer(view.currentBasemapLayer, 0);
+        view.model.getSelectedBasemapLayer(view.esriMap);
```
mmcfarland: This change is a little opaque as well.  It looks like it's taking advantage of some side effects in `getSelectedBasemapLayer` which is kind of confusing.  Was the rest of the code superfluous?  Consider renaming the method if it is actually more like `setBasemapLayer` or something more accurate.

src/GeositeFramework/js/Plugin.js
```diff
@@ -57,7 +57,7 @@ require(['use!Geosite',
                                 });
                             }
                             d.resolve();
-                            esriMap.resize(true);
+                            //esriMap.resize(true);
```
mmcfarland: Can you add a comment to help find and resolve this in the future?

src/GeositeFramework/js/Map.js
kellyi: I think it's still calling a `resize` internally but doesn't seem like there's an exposed method similar to what we were using with 3.19. It seems like it would get called on the MapView/SceneView, but in the docs for MapView, at least, `resize` is only listed as an event:

https://developers.arcgis.com/javascript/latest/api-reference/esri-views-MapView.html

(For comparison, it's listed as a method in 3.19: https://developers.arcgis.com/javascript/3/jsapi/map-amd.html#resize)

src/GeositeFramework/js/Plugin.js
kellyi: Sure. I'm adding a `TODO` comment plus a short explanation of why it's commented out & a link to the docs for reference.

mmcfarland:
One last note, this branch currently rebases against prototype cleanly- can you do so to try and keep the pilot branch tracking prototype for now?

kellyi:
Sure! I'll rebase `...pilot` on `prototype`, then push it remote again.

src/GeositeFramework/js/Map.js
kellyi: Sounds good. I removed the nesting altogether from these comments. For commenting out single lines, I'll drop the `/*` & `*/` stuff and just prefix the line with `//`. For longer pieces, I'll keep the multiline comment format but move the `/*`s immediately before the disabled code.

src/GeositeFramework/js/Map.js
```diff
@@ -122,16 +133,26 @@ require(['use!Geosite',
                     }
             }, 300),
             loadEventFired = false;
+        */
+        var esriMap = new Map({ basemap: 'topo' });
```
mmcfarland: The framework previously allowed setting the initial basemap via configuration in the region.json (uses the first in the list of basemap layers) - I don't think we _need_ to retain that in this pilot branch, but it'd be good to at least leave a comment that there's some expediency being taken here.

src/GeositeFramework/js/Map.js
kellyi: Looked through the docs some more and it looks like I may be able to un-comment this and adjust how it's set instead:

https://developers.arcgis.com/javascript/latest/api-reference/esri-config.html#portalUrl

mmcfarland: I don' think so, actually.  A `portal` is a specific esri technology for self hosted clould like environment.  The proxy setting here is really just to point to a server executed redirect.

kellyi: Update worked, so uncommenting this.

kellyi: Oops -- I linked the wrong thing above. I think the actual upgrade would be:

```
esriConfig.request.proxyUrl = "proxy.ashx";
```

src/GeositeFramework/js/Map.js
kellyi: Sounds good. Adding a `TODO` comment to consider re-enabling and explaining that it's hard-coded currently to use the topographic basemap.

src/GeositeFramework/js/Map.js
kellyi: Investigated this a little bit more and it looks like we might want to adjust how we're creating the basemap layers in accordance with the new API. I think right now this PR takes advantage of just setting each layer individually and directly, but the better way to handle this would be to create an initial `Basemap` object using the `baseLayers` specified in `region.json` -- https://developers.arcgis.com/javascript/latest/api-reference/esri-Basemap.html#baseLayers -- and then adjust everything to toggle those layers on changing the dropdown as it was working before. 

I think what may be happening on this branch is that they're being added as "operational layers" rather than basemap layers, and the new API seems to want to split those into two categories: 

https://developers.arcgis.com/javascript/latest/guide/migrating/index.html#mapandlayer

For now I'm going to restore the code I cut, but comment it out with a note about updating how it's handled and a link to the docs. I'll make a new issue for setting them up differently, as I think that would also be able to solve setting the basemaps from `region.json` as was the case before.

src/GeositeFramework/js/Map.js
kellyi: Similar to the comment on `getSelectedBasemapLayer` above, I'm going to restore the cut code but comment it out with a note explaining that we'll need to update how we configure the basemap layers initially and how we toggle them. It's correct, however, that this is making use of a side-effect to work. 

I think we'd probably prefer that it *not* work like that, so I'll leave the method name the same for now

kellyi:
Pushed a few commits to address the comments. Many of the new changes involved reformatting comments to make them more clear, but I also re-enabled setting the `proxyUrl` and restored (yet commented out) the logic used for setting basemaps and saving them to a cache. 

I think these pieces will require more adjustments for the new API, and can be folded into re-enabling setting the initial basemap from `region.json` again. I left the commented-out code so it will be more apparent what logic we're intending to replicate.

kellyi:
Thanks for your help with this! Going to merge this PR, and I'll make some new cards for this and regional-planning.

Issue #836, opened 2017-01-11
Retarget Styleguide components


mmcfarland:
Some form components are hiding raw HTML form components instead of targeting the styleguide markup version.

PR #837, opened 2017-01-11
Namespace styleguide form components

mmcfarland:
The styles on form components like checkboxes and radio buttons were too
broad and ended up hiding raw html elements.  These are changed to require
classes on the surrounding markup to ensure we're targeting only form
components which are trying to make use of styleguide styles.

Connects #836 

Testing
* If it doesn't already exist, load the [Benefits Explorer](https://github.com/CoastalResilienceNetwork/benefits_explorer_prototype) plugin into your environment
* Verify the raw checkboxes are visible on the "View HydroBASINS" section

screenshot from 2017-01-11 11 49 46

* Ensure the styleguide form components are still stylized (Create Map | Save & Share)

screenshot from 2017-01-11 11 50 28


mmcfarland:
Also updated the styleguide gh-pages branch:
http://coastalresiliencenetwork.github.io/GeositeFramework/

Fixes visible there are:
* Slider component on Component tab should resemble that from the "Sample" tab (previously was just raw jQueryUI slider)
* Form tab checks/radios should still be styled.

caseypt:
The styleguide changes also look good.

mmcfarland:
Right - I forget that save 'n share has no form elements besides text input.

PR #838, opened 2017-01-11
Make layer info box visible

lewfish:
This PR makes the layer info box visible.

untitled
#### Testing
* In the app, select the Region Planning plugin, and then expand elements in the layer selector until it causes a vertical scrollbar to appear. Check that the vertical scrolling behavior is the same as before, and that no horizontal scrollbar appears.
* Click the info icon on a layer and check that the info box displays and can be closed by clicking the X.

Connects #63


kellyi:
About to take a look!

Issue #833, opened 2017-01-10
Overflow menu changes


mmcfarland:
As proposed by TNC (and moved from #829 for better tracking)

- [ ] Move the hamburger menu to the left of the project title so the hamburger icon lines up with the app icons.
- [ ] Allow the project title to expand over the top bar space, not being constrained within the vertical black column width

PR #841, opened 2017-01-12
Remove styleguide rules from main.css

caseypt:
These rules should only exist in one place. Encountered while working on #833.

**Testing**
- Verify that the rules removed here exist in `styleguide.css` and that the site generally looks the same.

Connects to #833


kellyi:
Out of curiosity, could the first nine lines of `main.css` get removed, too? They're also in `styleguide.css`, but it may be that we need the `import` statement at least:

```
@import url("../fonts/fontello/css/mantle.css");
.clearfix:after,
.clearfix:before {
    content: " ";
    display: table
}
.clearfix:after {
    clear: both
}
```

caseypt:
Good call. I'll take a look.

kellyi:
Looks good otherwise! Double checked the `index` file to verify that the `styleguide.css` file was still being imported and it is.

caseypt:
Made the change. Thanks.

PR #842, opened 2017-01-12
Add back overflow styling to sidebar

lewfish:
A previous PR https://github.com/CoastalResilienceNetwork/GeositeFramework/pull/838 changed the overflow styling to make the Regional Planning info box visible. Unfortunately, it broke the scrolling on the Launchpad, so we undo that PR in this commit. 

#### Testing
* Check that you can scroll in the Launchpad and Regional Planning plugins when there is overflow, on Chrome, Firefox, and IE11.

Connects #808

### DIFF
--- a/src/GeositeFramework/Views/Home/Index.cshtml
+++ b/src/GeositeFramework/Views/Home/Index.cshtml
@@ -372,7 +372,7 @@
     </script>
 
     <script type="text/template" id="template-plugin-container">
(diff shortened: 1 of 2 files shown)


kellyi:
Launchpad scrolling works in all three browsers, but it seems like scrolling doesn't work in regional-planning.

mmcfarland:
Since this is supporting the release of Protecting Water which doesn't use Regional Planning - let's merge this in so we can go ahead with that launch, and track an issue specific to Regional Planning.

kellyi:
In that case this is good to go. +1

Issue #844, opened 2017-01-12
Adjust basemap initialization and toggling


kellyi:
To use the Esri ArcGIS JS Api v4.2, we'll need to update how basemaps are initialized and toggled by the framework. The new API splits basemap layers and operational layers and handles them differently, so we'll need to adjust for that, update the method calls we use to show and hide basemaps, and possibly update how the cache works.

The `js-api-4x-pilot` branch has comments in `Map.js` pointing to what looked like the most relevant API docs.

PR #849, opened 2017-01-20
Adjust basemap initialization & toggling

kellyi:
## Overview

This PR updates how we set the basemaps for the new pilot branch. 

Previously the layers specified in `region.json` had been added to the map directly as `TileLayers` on this branch, above an already-set `topo` basemap. Now we initialize the `Map` object without a preset basemap, then create and cache Basemap objects from the `region.json` layers.

## Screenshots

screen shot 2017-01-20 at 4 25 24 pm

screen shot 2017-01-20 at 4 25 48 pm

## Notes

Most of this restores prior functionality, but one of the more significant changes is in the `selectBasemap` method. It seems like it had previously done some shuffling through a set of tile layers for basemaps, showing the selected one and hiding all the others. Here I created one Basemap object for each layer specified in `region.json`, and then explicitly set `view.esriMap.basemap` to be the selected basemap (which excludes the others). On balance this seems to be more in tune with how the new API wants to have things set:

https://developers.arcgis.com/javascript/latest/guide/migrating/index.html#properties

(I also removed or updated some `TODO:` comments added when initially creating the pilot branch.)

## Testing
 * get and rebuild this branch, along with the `js-api-4x-pilot` branch of regional-planning, then open it in the browser
 * add this log statement to line 42 of `Map.js`:

```
window.console.log('Creating basemap', basemap.name);
```

 * refresh the site in the browser and verify that the topographic basemap's visible initially and that the console logs `"Creating basemap Topographic"`
 * use the basemap toggle to select "National Geograhic" and verify that it's visible and that you see `"Creating basemap National Geographic"` in the console
 * switch back to the Topographic basemap and verify that it becomes visible but that you don't see a `"Creating..."` log statement, indicating that the basemap has been cached
 * turn on the "Hurricane Sandy - County Impact Analysis" layer in the Regional Planning plugin and verify that it shows up
 * toggle through the basemap layers and verify that the Hurricane Sandy layers remain visible

Connects #844

### DIFF
--- a/src/GeositeFramework/js/Map.js
+++ b/src/GeositeFramework/js/Map.js
@@ -8,6 +8,7 @@ require(['use!Geosite',
          'esri/config',
          'esri/Map',
          'esri/views/MapView',
+         'esri/Basemap',
          // Scalebar is not yet implemented in Esri JS API v4.2:
          // https://developers.arcgis.com/javascript/latest/guide/functionality-matrix/index.html#widgets
          // 'esri/dijit/Scalebar',
@@ -24,6 +25,7 @@ require(['use!Geosite',
              esriConfig,
              Map,
              MapView,
+             Basemap,
              // Scalebar is not yet implemented in Esri JS API v4.2:
              // https://developers.arcgis.com/javascript/latest/guide/functionality-matrix/index.html#widgets
              // ScaleBar,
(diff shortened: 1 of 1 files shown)


kellyi:
Thanks! Going to merge.

Issue #856, opened 2017-01-24
Change behavior of deactivated plugin


mmcfarland:
Currently, when any plugin is deactivated other than by the `X` button, it will keep data layers on and hibernate - the same behavior as when the user clicks the minimize `_` button.  Modify the behavior such that switching between plugins has the same behavior as `X` and fully closes down the plugin.  This is essentially the same behavior that has forced in for "Get Started"/Launchpad - so we may be able to take that one-off and codify as regular behavior.  The minimize button should remain and operate as normal, but that is the only path to activate that behavior.

kellyi:
Should closing a plugin sidebar by clicking on the plugin icon in the list on the left be handled as a "minimize" event or a "turn off" event? For example, if regional-planning is open...

<img width="1032" alt="screen shot 2017-01-25 at 10 29 26 am" src="https://cloud.githubusercontent.com/assets/4165523/22296836/687d52b8-e2e9-11e6-8c49-1d1330a74819.png">

... and I click on the "Regional Planning" in the plugin list, should it minimize and keep the layers and close altogether?

<img width="1031" alt="screen shot 2017-01-25 at 10 29 40 am" src="https://cloud.githubusercontent.com/assets/4165523/22296885/8e13fa7c-e2e9-11e6-923e-52693a12a95a.png">

ajrobbins:
@kellyi I think it should minimize.

kellyi:
Thanks @ajrobbins. One more q: If a plugin is minimized, and then I select another plugin, should the minimized plugin close? It seems like selecting a new plugin from the list when one is already open and has its sidebar visible/isn't minimized should close the open one; should it work the same way if a plugin is open and minimized?

ajrobbins:
Yes, given the behavior we're implementing when the sidebar is open, I think selecting another plugin should close the open & minimized one.

ajrobbins:
Closing the minimized plugin would make sense, but then there would be no way to have multiple apps open (and data layers on) at once at all. So I think it should stay open.

ajrobbins:
@kellyi Just got word from Zach that we can keep the existing behavior of deactivated plugins. Sorry for the trouble - thanks for working on this, we should close and not merge.

PR #860, opened 2017-01-25
Turn off an active plugin on selecting a new plugin

kellyi:
## Overview

This PR updates the plugin behavior such that we close open, non-minimized plugins and turn off their layers in on selecting a different plugin. For plugins which have been minimized, we keep them/their layers on. We also keep the minimize behavior when the user clicks the name of a plugin which is currently open.

Connects #856 

## Notes

* I'm not entirely sure if this is the desired behavior based on the discussion in #856, but I'm putting this up so we can evaluate it before merging. It may be that we want to close the plugins in all cases except when the minimize button has been clicked, but that would prevent users from having layers from two plugins active & visible simultaneously.

 * The first commit here removes an event listener no longer used in the framework.

## Testing

 * rebuild this branch with the regional planning plugin included
 * verify the following paths in the app:
     * clicking the minimize button for a plugin keeps its layers on and its name highlighted in the list
     * clicking the name of an open plugin minimizes it
     * clicking a new plugin name from the list when another one's active closes the previously active plugin and removes its layers
     * minimizing a plugin, then clicking a new plugin name from the list keeps the first plugin's layers active
     * clicking close on an open plugin closes it & removes its layers

### DIFF
--- a/src/GeositeFramework/js/Plugin.js
+++ b/src/GeositeFramework/js/Plugin.js
@@ -502,10 +502,6 @@ require(['use!Geosite',
                 .find('.plugin-off').on('click', function () {
                     model.turnOff();
                 }).end()
-                // Unselect the plugin, but keep active
-                .find('.plugin-close').on('click', function () {
-                    model.deselect();
-                }).end()
                 .find('.plugin-help').on('click', function () {
                     pluginObject.showHelp();
                 }).end()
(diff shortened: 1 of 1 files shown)


kellyi:
Leaving this open until we verify that this is how plugin toggling should work.

kellyi:
Closed this PR per update on #856

Issue #858, opened 2017-01-25
Add titleDetail to header


ajrobbins:
With #853, we'll be losing the region name from the header for Coastal Resilience sites. Add a configurable second header to the header that looks the same as the old header did, but isn't a drop down:

![screen shot 2017-01-25 at 9 33 51 am](https://cloud.githubusercontent.com/assets/12868342/22294506/88c97cf2-e2e1-11e6-9fb1-310798125dca.png)

mmcfarland:
For clarification, this isn't a second header element, but rather just a label showing a region name in the existing header.  I believe we already have a region.json config element, `titleDetail` which is being used as the display value portion of the existing dropdown, we'll probably just want to nix the dropdown control component.

caseycesari:
The behavior of the region menu was that if no items were provided, the menu wasn't rendered but the subtitle still was. I'm not sure if that logic (and the DOM elements) are still in the code, though.

Issue #853, opened 2017-01-24
Move regionLinks dropdown instide of a overflow drop down


mmcfarland:
Initial work has been done on this here: https://github.com/CoastalResilienceNetwork/GeositeFramework/commit/47b592205c502861395265144d736cdb8e472814

This should be a sub-dropdown menu within the main dropdown menu.

To figure out: @ajrobbins in the Coastal Resilience program, the region link area doubled as a secondary header for the region name.  This isn't the case in Protecting Water, can you find out if we should keep/introduce a secondary header for this (ie Coastal Resilience - New York) now that this will be gone?

ajrobbins:
I'll ask.

PR #861, opened 2017-01-25
Replace region dropdown with region subtitle in header

kellyi:
## Overview

This PR removes the region links dropdown from the header and replaces it with a subtitle from `region.json`'s `titleDetail.text` property. In addition to removing the loop which configured the links, I replaced the nested divs with a single `nav-region-subtitle` div and some rules to style it.

Connects #858

## Screenshot

screen shot 2017-01-25 at 2 58 48 pm

## Notes

I picked up #853 which is related to this.

## Testing

* rebuild this branch, then view in the browser
* verify that you see the "Sample Region" subtitle and that it's no longer a button

### DIFF
--- a/src/GeositeFramework/Views/Home/Index.cshtml
+++ b/src/GeositeFramework/Views/Home/Index.cshtml
@@ -636,21 +636,8 @@
                 @RenderLinks(Model.HeaderLinks)
             </ul>
         </div>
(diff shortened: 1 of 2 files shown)


kellyi:
Just checked and it looks like the only place elements with those classes appear is in the prototype and the styleguide. I'll remove them from `main.css` file.

PR #862, opened 2017-01-25
Move region links into main dropdown

kellyi:
## Overview

This PR moves the `regionLinks` out of a previously used separate dropdown menu into the main hamburger menu dropdown. Along with updating the index file and the CSS, the changes here also adjust the structure of `region.json` to nest the `regionLinks` in the `headerLinks`. 

(If we keep the `region.json` change, we'll need to update the Wiki instructions and any regions which build).

Connects #853 

## Screenshots

screen shot 2017-01-25 at 5 23 49 pm

## Notes

The first commit here is from #861 

## Testing

 * rebuild this branch in VS, then view in the browser
 * open the menu and mouseover the "Other Regions" item
 * verify that a submenu of outbound links appears and that the links work

### DIFF
--- a/src/GeositeFramework/region.json
+++ b/src/GeositeFramework/region.json
@@ -11,22 +11,20 @@
     },
     "headerLinks": [
         { "text": "Azavea", "url": "http://www.azavea.com/" },
-        { "text": "GIS", "url": "http://en.wikipedia.org/wiki/Geographic_information_system" },
-        { "text": "Other Regions", "items":
-            [
-                { "text": "California", "url": "http://maps.coastalresilience.org/california/" },
-                { "text": "Connecticut", "url": "http://maps.coastalresilience.org/connecticut" },
-                { "text": "Global", "url": "http://maps.coastalresilience.org/global/" },
-                { "text": "Grenada, St. Vincent & the Grenadines", "url": "http://maps.coastalresilience.org/gsvg/" },
-                { "text": "MesoAmerican Reef", "url": "http://maps.coastalresilience.org/mar/" },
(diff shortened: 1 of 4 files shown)


kellyi:
Thanks! I'll hold off on merging this until #861 has been reviewed, then merge them in order.

src/GeositeFramework/region.json
```diff
@@ -11,20 +11,22 @@
     },
     "headerLinks": [
         { "text": "Azavea", "url": "http://www.azavea.com/" },
-        { "text": "GIS", "url": "http://en.wikipedia.org/wiki/Geographic_information_system" }
-    ],
-    "regionLinks": [
```
mmcfarland: Is this key no longer supported in the config?  If so, we'll want to update the JSON schema and also the wiki documentation for the region.json file - noting that there is a special `items` key that can nest additional links.

kellyi: Yeah -- one change in this update is that it moves what was previously in `regionLinks` into the `headerLinks`.

I'll update the schema & wiki and also update the New York & New Jersey develop region builds to match.

kellyi: Seems like it may be less disruptive to keep the `regionLinks` key, so I'll update the framework to get its links from there & display them as pictured above.

kellyi:
I updated this to restore the `regionLinks` key. The PR keeps the prior changes in which we now loop over `link.Items` to create the submenu, but updates `Geosite.cs` to add a method used to nest the `regionLinks` in the `headerLinks` before the view renders.

caseycesari:
Changes look good. I think it makes sense to not change the schema again. Good call.

kellyi:
Thanks! Going to merge this.

Issue #852, opened 2017-01-24
Initially select contents of Save & Share text box


mmcfarland:
* Rename Help -> Tour, including translation files.
![screenshot 2017-01-24 at 2 00 54 pm](https://cloud.githubusercontent.com/assets/1014341/22262094/9218d6bc-e23d-11e6-8bef-cea34cb690d5.png)

* Initially select contents of Save & Share text box.  I'm fairly sure that this was prior behavior that recently broke.
![screenshot 2017-01-24 at 2 02 36 pm](https://cloud.githubusercontent.com/assets/1014341/22262158/c9f5aa2e-e23d-11e6-8132-38106a4f1756.png)

kellyi:
Renamed "Help" -> "Tour" in #863; still working on the Save & Share box.

kellyi:
Renamed this issue to highlight the second item.

PR #866, opened 2017-01-30
Restore highlight & select for Save & Share

kellyi:
## Overview

This PR fixes a bug which had prevented the shortlink from being selected when the Save & Share popup opened. 

The problem seems to be a typo on a CSS classname for that input which made its way into the app when we created the prototype: previously the input had a class named `permalink-textbox`, but during the transition to the prototype, the class was updated to `permalink-tex` without a corresponding change in the `Permalink.js` file.

Here's the previous code:

https://github.com/CoastalResilienceNetwork/GeositeFramework/blame/590c09222009537457159e82270c4174e9789562/src/GeositeFramework/Views/Home/Index.cshtml

To fix this I changed the class name back to `permalink-textbox`.

Connects #852 

## Screenshot

screen shot 2017-01-30 at 1 53 36 pm

## Testing

* rebuild this branch in VS, then open it in the browser
* open the Save & Share and verify that the Permalink input is focused and selected. Copy the value to the clipboard, open another browser tab, then paste the text and it should show the permalink and lead back to the site.

### DIFF
--- a/src/GeositeFramework/Views/Home/Index.cshtml
+++ b/src/GeositeFramework/Views/Home/Index.cshtml
@@ -502,7 +502,7 @@
             <div class="popover-content">
                 <div class="row" id="permalink-dialog">
                   <label class="popover-section"><div class="h3 i18n" data-i18n="Permalink">Permalink</div>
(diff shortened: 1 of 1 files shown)


kellyi:
Thanks for taking a look!

PR #869, opened 2017-02-08
Enable toggle of 3d & 2d mode

mmcfarland:
### Overview
Expose methods to plugins which toggle between SceneViews and MapViews. This is currently done naively by having two views available but one hidden from the DOM.  When hidden, it appears a view does not attempt to render, which makes this a seemingly good approach.

Connects #868 

### Testing
* Clone https://github.com/CoastalResilienceNetwork/3d-sample/pulls into your plugin list on this branch
* Test that 3d view is activated when opening "Pilot Demo" plugin.  Basemaps, etc work as usual.
* 2d is activated when closing
* If demo plugin is hibernated, it stays in 3d and other plugins (regional planning) can add layers.  These layers persist through view changes.
* Zoom level and extent are persisted through view changes

screenshot from 2017-02-08 14 50 27

### DIFF
--- a/src/GeositeFramework/Views/Home/Index.cshtml
+++ b/src/GeositeFramework/Views/Home/Index.cshtml
@@ -205,6 +205,7 @@
             <div class="flex-container">
                 <div id="map-0" class="map">
                     <div id="esri-mapview-mount"></div>
+                    <div id="esri-sceneview-mount"></div>
                     <div class="control-container">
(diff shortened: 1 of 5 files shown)


kdeloach:
+1

mmcfarland:
Yeah, most of the changes were just threading references to the two view objects down to the plugin.

Issue #873, opened 2017-02-13
Vector basemap support


mmcfarland:
Add vector tile basemap options to a region to demonstrate support for this new type of layer.

PR #874, opened 2017-02-13
Allow short code basemap URLs

mmcfarland:
### Overview
Vector tile rendering support is native to the 4.x API, but the official
basemap options are made available as string codes and the URL doesn't
have to be specified. This commit allows that type of basemap entry which
will allow vector tile demonstration.

Connects #873 

### Demo
screenshot from 2017-02-13 15 08 10

screenshot from 2017-02-13 15 10 26

### Testing
* Clone https://github.com/CoastalResilienceNetwork/3d-sample/ if you don't already have it
* Rebuild the framework and verify you can use the new vector map options as provided in the sample `region.json` configuration.  With one selected, open the 3D plugin and verify the same tile set is used on the globe.  
* Note that on my windows VM, rendering performance is not excellent.

### DIFF
--- a/src/GeositeFramework/js/Map.js
+++ b/src/GeositeFramework/js/Map.js
@@ -41,11 +41,16 @@ require(['use!Geosite',
         // Return an ESRI Basemap object for the currently-selected basemap spec
         var basemap = getSelectedBasemap(model);
         if (basemap.layer === undefined) {
-            basemap.layer = new Basemap({
-                baseLayers: [new TileLayer(basemap.url)],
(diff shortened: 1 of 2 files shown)


kellyi:
Started testing this and found out that my dev env for the newest changes is in some weird bad state, so I'm going to remove the repo and start over from scratch to test this.

kellyi:
The issues I had getting this to run were related to WebGL 1.x not working in Chrome in my local's Windows dev env. Tested this in Firefox and it worked well!

Issue #871, opened 2017-02-09
Hide Save and Share save code


mmcfarland:
A regression has allowed this to be on by default.  Should be hidden until "Get it here" is clicked.

kellyi:
Shares a point with #877

kellyi:
Visibility issue was due to a typo in the div id plus a missing CSS rule to hide it to start. I'm not sure how this worked in the prior version of the framework, but should the "Get it here" text & link hide when after it's clicked and the save code has been displayed?

Before clicking:

![screen shot 2017-02-16 at 9 59 00 am](https://cloud.githubusercontent.com/assets/4165523/23026361/066bea42-f42f-11e6-9111-a53a600af47b.png)

After clicking:

![screen shot 2017-02-16 at 9 59 11 am](https://cloud.githubusercontent.com/assets/4165523/23026369/0b1ebd80-f42f-11e6-98a7-f5c25e13f127.png)

PR #880, opened 2017-02-16
Initially hide long Save Code, then toggle visible on clicking "Get it here"

kellyi:
## Overview

This PR fixes a bug whereby the long Save Code had been visible by default in the new framework. There seemed to be two errors here: first, the `id` for the permalink had gotten truncated from `long-permalink-textbox` to `long-permalink-tex` in the updated framework, which meant that jQuery couldn't find the right element to toggle. Second, the permalink element was not set to hide initially, so clicking "Get it here" for the first time would then hide it.

The second commit here fixes the CSS id typo; the third one sets it to hide until the "Get it here" link has been clicked. First one's here to reformat the elements before making any changes.

Connects #871

## Screenshots

Before clicking "Get it here" ->

screen shot 2017-02-16 at 9 59 00 am

After clicking "Get it here" ->

screen shot 2017-02-16 at 9 59 11 am

### DIFF
--- a/src/GeositeFramework/Views/Home/Index.cshtml
+++ b/src/GeositeFramework/Views/Home/Index.cshtml
@@ -532,7 +532,7 @@
                       <a href="javascript:void(0);" class="show-long-permalink i18n" data-i18n="Get it here.">
                           Get it here.
                       </a>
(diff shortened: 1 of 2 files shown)


kellyi:

Issue #883, opened 2017-02-17
Update styleguide


mmcfarland:
The style guide is a sub-project within this repo with its own build tools and dependencies.

https://github.com/CoastalResilienceNetwork/GeositeFramework/tree/develop/styleguide

We've been provided with an additional page that takes the style output from above and uses it to make a tool to help copy the needed markup and JS syntax if required.

* Port new components to the original style guide, both markup in the display and also the style/classes to the final css
* Add links to and from both style guide pages
* Add jquery UI script to framework to allow usage of sliders
* Update framework version of styleguide css after the above are completed
* Update the gh-pages branch to have the updated styleguide apps

When we put a PR up for this, we can have @cschneebeck test out a build based on the branch to see if he's able to remove his custom styleguide in favor of the updated framework version.

[styleGuide.zip](https://github.com/CoastalResilienceNetwork/GeositeFramework/files/784146/styleGuide.zip)

mmcfarland:
An instance of the "get code" page can be found here: http://cschneebeck.com/styleguide/

Issue #875, opened 2017-02-14
Change colors


ajrobbins:
Request from TNC:
- header background from #0f1C27 to #5394B6 (main.css line 857)
- nav-apps background from #27343E to #333333 (main.css line 787)
- font color defined for body from #73787D to #5d6165 (main.css line 441) - provides more contrast against a white background.
- selected radio, pills, and checkboxes background from #4b9CAF to #0096D6 (styleguide.css)
- primary button and slider handle background from #4b9CAF to #0096D6 (styleguide.css)

caseycesari:
Shares a point with #872

caseycesari:
Shares a point with #872

Issue #859, opened 2017-01-25
Use defaults for colors.active, colors.primary, etc if not specified in `region.json`


kellyi:
Currently if a build specifies only some of these colors in its `region.json` file, the build will fail at runtime.

We should update the framework to have it use default values for any unspecified colors.

caseycesari:
Implemented in #884

PR #884, opened 2017-02-17
Update Framework colors

caseycesari:
Updates various framework colors per request from TNC

image

**Testing**
- Test that the framework colors match those requested in #875, with and without the color overrides in `region.json`.

**Note**
Styleguide changes listed in #875 will be addressed in #883 

Connects to #875 
Connects to #859


caseycesari:
Sure. Will do.

caseycesari:
Added a commit to make the colors optional. Best way to test it is to only provide some of the colors in the config, and to remove the config all together.

caseycesari:
Thanks

Issue #882, opened 2017-02-17
Plugin close button is visible for active, minimized apps when sidebar is minimized


caseycesari:
![image](https://cloud.githubusercontent.com/assets/1042475/23079716/0d1e61a6-f51c-11e6-8002-bb956dd5ad59.png)

I believe this was introduced in https://github.com/CoastalResilienceNetwork/GeositeFramework/pull/879, now that the close button is shown without needing to hover. We should hide it when the sidebar is minimized.

kellyi:
Shares a point with #877

kellyi:
This seemed to be happening because the `:hover` postfix was still attached to this line:

https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/develop/src/GeositeFramework/css/main.css#L872

Issue #877, opened 2017-02-15
Show plugin sidebar X buttons without hover


ajrobbins:
With https://github.com/CoastalResilienceNetwork/GeositeFramework/pull/870, we are showing an X button for active yet minimized apps in the sidebar menu on hover. TNC wants us to have this button display at all times for active yet minimized apps, not just on hover. This will make the feature more easily discoverable by users.

kellyi:
Should the close icon button also display for active, non-minimized plugins?

![screen shot 2017-02-15 at 6 12 03 pm](https://cloud.githubusercontent.com/assets/4165523/22999606/563f5674-f3aa-11e6-9bd1-0c101473eea6.png)

mmcfarland:
Nope, maintain the existing behavior for that.  Still just minimized.

kellyi:
Shares a point with #871

kellyi:
Shares a point with #882 too

PR #885, opened 2017-02-17
Hide plugin close button on narrow viewports

kellyi:
## Overview

This PR fixes an error introduced in fixing #877 whereby the close button for minimized plugins was also visible on narrow viewports when the sidebar was narrower. The fix involves nesting the `display: inline-block` rule for the minimized plugins inside a check that the viewport's > 991px wide, which matches the value we use to set the narrow sidebar

Connects #882 

## Screenshots

*Desktop-sized screen*:
screen shot 2017-02-17 at 5 07 16 pm

*Tablet width, small sidebar*
screen shot 2017-02-17 at 5 07 26 pm

*Tablet width, expanded sidebar*
screen shot 2017-02-20 at 10 11 17 am

## Testing
 * rebuild this branch in VS then view it in the browser
 * at standard desktop width, check that the close plugin button appears properly for minimized plugins
 * at tablet width, check that the close plugin button does not appear for minimized plugins when the sidebar displays only icons, and check that the close button does appear when all open plugins are minimized
 * verify that it still doesn't show up for active, unminimized plugins in either case

### DIFF
--- a/src/GeositeFramework/css/main.css
+++ b/src/GeositeFramework/css/main.css
@@ -865,13 +865,13 @@ nav {
 .plugin-launcher.active + .button-link.button-sidebar-plugin-close {
   display: inline-block;
 }
+@media (max-width: 991px) {
+  .nav-apps-narrow .button-sidebar-plugin-close {
+    display: none !important;
+  }
+}
 .plugin-launcher.app-displayed.active + .button-link.button-sidebar-plugin-close {
   display: none;
-}
-@media (max-width: 991px) {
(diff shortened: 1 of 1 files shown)


kellyi:
Just realized this fixes half of it; but that the close button should still show up on tablets when the sidebar's un-minimized. Pushing a fixup to address.

kellyi:
Pushed an amended commit which should fix it in both instances.

kellyi:
Thanks for taking a look -- and for discovering the original bug!

Issue #876, opened 2017-02-14
Adjust icon sizing


ajrobbins:
- Decrease the size of the white binocular icon relative to the transparency
- Increase the framework icon image size to make the icons look bigger

caseycesari:
Shares a point with #872

PR #890, opened 2017-02-21
Update launchpad icon and adjust plugin icon sizes

caseycesari:
The binoculars in the launchpad icon looked bigger than other icons because the shape of the binoculars allows the image to take up more space in the icon. Also, the icon padding was adjusted to make all of the plugin icons larger.

Before:
screen shot 2017-02-21 at 12 38 29 pm

After:
screen shot 2017-02-21 at 12 47 02 pm

**Testing**
- Load the site and verify that the binoculars do not appear as large and that all of the plugin icons are larger.

Connects to #876


caseycesari:
@mmcfarland Made some adjustments so that the menu icon lines up better with the plugin icons.

Before:

image

After:

image

caseycesari:
Thanks

Issue #888, opened 2017-02-20
Add example of using image in modal


mmcfarland:
In a sample app, show how to use TINY to make an image viewer for a full size image from a thumbnail (css resized, can be the same actual image).   

Try using an infographic from the old app as it's likely what they'll want to use this on.
http://maps.coastalresilience.org/gulfmex/plugins/future_habitat/FutureHabitat_c.jpg

 Add a snippet of this code to the wiki, as well.

kellyi:
Updated the wiki's Plugin Settings page with instructions for creating a modal here:

https://github.com/CoastalResilienceNetwork/GeositeFramework/wiki/Plugin-Settings

PR #891, opened 2017-02-21
Demonstrate using modal image in sample plugin

kellyi:
## Overview

This PR updates the sample `identify_point` plugin to include a demonstration of how to create an image modal popup triggered on clicking a thumbnail. I tried to keep the footprint of the changes as small as possible; the updates here include:

- create a clickable div for the thumbnail in the plugin
- add a click event listener to the div which calls a new `renderSampleImageModal` method
- use TINY.box to render a modal from the image source path passed in as an argument to `renderSampleImageModal`

Connects #888 

## Screenshots

*Thumbnail*
screen shot 2017-02-21 at 1 14 51 pm

*Modal*
screen shot 2017-02-21 at 1 14 58 pm

## Notes

I tried to keep the changes here small but legible, and I'm happy to push it more in one direction or the other. Once this is ready to merge, I'll update the Wiki either to add a new page or to update the "Plugin Settings" page: https://github.com/CoastalResilienceNetwork/GeositeFramework/wiki/Plugin-Settings

## Testing

 * open the "Identify Point" plugin and click on the thumbnail
 * verify that you see the modal appear on the screen
 * click the background to close the modal and verify that it disappears


src/GeositeFramework/sample_plugins/identify_point/main.js
```diff
@@ -103,6 +120,19 @@ define(["dojo/_base/declare", "framework/PluginBase"],
...
+
+            // Use TINY to display a modal image from a path to an image
+            renderSampleImageModal: function(imageSourcePath) {
+                createHTMLForModalImage = function(src) {
+                    return '<img src="' + src + '"/>';
+                }
+
+                TINY.box.show({
+                    animate: false,
+                    html: createHTMLForModalImage(imageSourcePath),
```
kdeloach: Consider inlining this function.

src/GeositeFramework/sample_plugins/identify_point/main.js
```diff
@@ -60,6 +70,13 @@ define(["dojo/_base/declare", "framework/PluginBase"],
                             break;
                     }
                 });
+
+                // Call `renderSampleImageModal` method on clicking thumbnail
+                $('.identify-point-plugin-thumbnail-image').on('click', function() {
+                    $thumbnailImage = $('#sample-plugin-thumbnail');
```
kdeloach: Missing `var`.

src/GeositeFramework/sample_plugins/identify_point/main.js
```diff
@@ -25,6 +25,16 @@ define(["dojo/_base/declare", "framework/PluginBase"],
 
                 // Hide the print button until the identify feature has been used.
                 $(this.printButton).hide();
+
+                // Display a clickable image thumbnail
+                $(this.container)
+                    .append('<p>Click the thumbnail to see a sample modal image popup</p>')
+                    .append('<div class="identify-point-plugin-thumbnail-image"' +
```
kdeloach: The plugin class names don't need to be globally unique if we use `$(this.container).find(...)`. Consider shortening this (remove `-image`?)

kdeloach: Maybe we can drop the wrapper div completely to keep it simple.

src/GeositeFramework/sample_plugins/identify_point/main.js
```diff
@@ -25,6 +25,16 @@ define(["dojo/_base/declare", "framework/PluginBase"],
...
+                            'style="cursor: pointer"></div>');
+
+                $('.identify-point-plugin-thumbnail-image')
+                    .append('<img src="sample_plugins/identify_point/FutureHabitat_c.jpg"' +
+                            'id="sample-plugin-thumbnail" width="300" height="195" />');
```
kdeloach: Consider using the `identify-point-plugin-thumbnail-image` class name on the `img` element instead of using an ID.

src/GeositeFramework/sample_plugins/identify_point/main.js
kdeloach: Consider putting the event handler directly on the `img` element. We could use `$(this)` to refer to the target element.

kdeloach:
Looks good but I did notice some z-index issues with TINY. Not sure if an issue exists for this already.

image

kellyi:
Thanks for taking a look! I'll address this feedback before merging, and I'll investigate the TINY z-index issue, too.

src/GeositeFramework/sample_plugins/identify_point/main.js
kellyi: Sounds good. I'll drop the `div` and move the cursor style and listener to the `img`.

kellyi:
Looks like the tiny box does have a z-index, but it's overridden by something else in the app:

screen shot 2017-02-21 at 4 42 47 pm

Going to make another issue for that, as I don't know whether changing that will entail adjusting other element z-indices.

src/GeositeFramework/sample_plugins/identify_point/main.js
kellyi: I'm updating this to:

```js
// Call `renderSampleImageModal` method on clicking thumbnail
$(this.container).find('#sample-plugin-thumbnail').on('click', function() {
    // Use the thumbnail source as the modal image source
    var sampleImageSource = this.src
    self.renderSampleImageModal(sampleImageSource);
});
```

src/GeositeFramework/sample_plugins/identify_point/main.js
kellyi: I dropped the class name altogether and am going to use the `sample-plugin-thumbnail` id attribute instead.

src/GeositeFramework/sample_plugins/identify_point/main.js
kellyi: Also going to replace the `this.src` assignment with

```js
var modalImageSource = 'sample_plugins/identify_point/FutureHabitat_c.jpg';
```

because it's slightly less opaque.

Issue #892, opened 2017-02-21
Tiny Box modal z-index is overridden


kellyi:
Currently the z-index attribute for the `.tbox` on `TINY.box` elements is overridden, which means modals can slide behind other elements on specific viewport sizes:

![screen shot 2017-02-21 at 4 42 47 pm](https://cloud.githubusercontent.com/assets/4165523/23186427/2076858a-f855-11e6-87a0-e14fe8e2f9a0.png)

PR #897, opened 2017-02-23
Increase tinybox element z-indexes

caseycesari:
Ensures tinybox modals are displayed on top of every other element in the page.

image

**Testing**
- Open the modal in the sample app and verify that it displays on top of every other element in the app.

Connects to #892


kellyi:
Works for the image popup:

screen shot 2017-02-23 at 3 23 20 pm

... Save & Share:

screen shot 2017-02-23 at 3 23 55 pm

... & Print modals:

screen shot 2017-02-23 at 3 24 22 pm

Issue #887, opened 2017-02-20
App sidebar collapse


ajrobbins:
TNC has requested that we add a sidebar collapse button, something like `<` that collapses the sidebar to the icon-only skinny sidebar.

PR #899, opened 2017-03-03
Add tool to expand and shrink the map

caseycesari:
## Overview

Add button to the map that allows the user to expand and shrink the size of the map by collapsing the sidebar.

This is also done automatically at small screen sizes when plugins are activated, so the tool listens for changes to the map size so that the icon can be appropriately updated (or else it would become out of sync).

Since the CSS rules for the collapsed sidebar were in a media query, they are copied and placed outside of the media query as well so that they can be used at all screen sizes. This is not ideal.

Connects #887 

### Demo

tnc887

## Testing Instructions

- Without any active plugins, use the button to expand and shrink the map. Verify that the button does indeed do this, and that the icon of the button stays in sync with the changes.
- Reduce the size of your browser window to less than 990 pixels.
- Activate and minimize plugins, and verify that the map expand tool button icon stays in sync with the changes.


src/GeositeFramework/css/main.css
```diff
@@ -824,6 +824,27 @@ nav {
     width: 28px;
 }
 
+.nav-apps-narrow {
```
caseycesari: I'd rather not have to duplicate all of these rules, but I couldn't think of a way to have inside and outside of the media query below.

kellyi:
This seems to be working well! I noticed one quirk -- or possibly an expected behavior: when the map's narrow and has the sidebar resize behavior already happening & there's a plugin open, clicking on the resize button will keep the plugin open but show the full sidebar, too:

screen shot 2017-03-06 at 11 51 08 am

Is that the expected behavior for narrow viewports? Everything else seems to be working well.

caseycesari:
> Is that the expected behavior for narrow viewports? 

I believe so. Since it can now be toggled manually, users can get the full-sized sidebar on narrow screens.

@ajrobbins can confirm.

kellyi:
Going to check this out one more time now that we've confirmed the expand map button should be visible & functioning at all screen sizes.

kellyi:
This is working well, but I just noticed one quirk: if plugins are minimized and the user clicks the expand button, the "x" close icon shows up again. I think there's an additional rule which stops that for narrow viewports which we generalize to all sizes.

screen shot 2017-03-07 at 3 14 04 pm

caseycesari:
Good find. I'll take a look.

caseycesari:
@kdeloach Amended the last commit to fix the issue you found.

caseycesari:
Thanks

Issue #898, opened 2017-03-02
App close X obscures long titles


ajrobbins:
See:

![screen shot 2017-03-02 at 11 56 31 am](https://cloud.githubusercontent.com/assets/12868342/23517783/60dcad7a-ff3f-11e6-8b5b-c39ac8fdd4ea.png)

PR #901, opened 2017-03-03
Limit the width of the plugin name element

caseycesari:
## Overview

Limiting the width of the plugin name element prevents the close button
from obscuring long names.

Connects to #898

### Demo

image

## Testing Instructions

- Activate one of the plugins, then minimize it.
- Using the developer tools, edit the name of the plugin so that it needs to wrap.
- Hover over the plugin name, and verify that the close button does not overlap with the name.

### DIFF
--- a/src/GeositeFramework/css/main.css
+++ b/src/GeositeFramework/css/main.css
@@ -823,6 +823,10 @@ nav {
     height: 28px;
     width: 28px;
 }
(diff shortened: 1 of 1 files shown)


mmcfarland:
screenshot 2017-03-07 at 12 31 15 pm

caseycesari:
Thanks!

Issue #900, opened 2017-03-03
Replace Get Started and Regional Planning icons


ajrobbins:
TNC has requested that all icons be bigger in the app. For our apps, Get Started and Regional Planning, @jfrankl and I created enlarged icons - they're actually the same size, but the content is now edge-to-edge.

Icons attached: They're white.
![icon_sm - az](https://cloud.githubusercontent.com/assets/12868342/23562852/23379f24-0012-11e7-8cc1-ed3979285002.png)
![binoculars-whoa-bold](https://cloud.githubusercontent.com/assets/12868342/23562867/310813a4-0012-11e7-9446-ed2910584241.png)

PR #902, opened 2017-03-03
Update launchpad icon

caseycesari:
The size of the image has stayed the same, but the content of the image
was enlarged to use more of the image canvas.

Connects to #900

Before:
image

After:
image


ajrobbins:
Looks good from the screenshot!

Issue #903, opened 2017-03-06
Make sure Chosen, JQueryUI are available to plugins


mmcfarland:
Two JS libraries (Chose & JQueryUI) were added to the framework to support styleguide features but may not be globally available.  Add two components (slider & chosen select) to the sample app and make sure the framework supports their being called either globally or within the `use` statement.

kellyi:
Was looking at how the i18-next library's set up, and it looks like it gets initialized in App.js by adding it to jQuery. Tried out a similar `addChosen` method and it looks like we can add it there and make it available to plugins:

https://github.com/CoastalResilienceNetwork/GeositeFramework/commit/ae1cf0f5f9f67a5f29054bcbc74127fcad4e3257

->

![screen shot 2017-03-07 at 5 27 29 pm](https://cloud.githubusercontent.com/assets/4165523/23681038/5d02dce0-035b-11e7-8258-b53878dcccdf.png)

Prior to that, however, $ doesn't have `.chosen` attached. Going to see about manually attaching it either here or somewhere else in the hierarchy.

PR #906, opened 2017-03-08
Make chosen and jquery-ui available globally

kellyi:
### Overview

This PR makes the chosen and jquery-ui libraries available to plugins by changing them into use module imports rather than cdn-loaded scripts, them importing them into the BasePlugin.js file after which they'll be available to plugins inheriting from it. To accommodate loading them via `use`, this supplants using script tags with importing the minified jquery-ui and chosen files.

This enables both libraries to be used without any additional imports.

To demonstrate using these, I also added a range slider, accordion menu, and multiselect box to the Identify Point plugin -- mostly by just copying the code from the style guide: http://coastalresiliencenetwork.github.io/GeositeFramework/snippet_index.html

Connects #903 

### Screenshot

screen shot 2017-03-08 at 4 01 56 pm

### Notes

~~Still two todos here, one which is to make the "pips" library available to the slider to show interval ticks as in the "Slider" example here: http://coastalresiliencenetwork.github.io/GeositeFramework/snippet_index.html~~

~~The other's to update the wiki with example code snippets.~~

Pips is now available via some updates to the imported file; the examples used in the sample plugin were pulled directly from the style guide, and the Wiki already mentions using snippets from the styleguide here:

https://github.com/CoastalResilienceNetwork/GeositeFramework/wiki/Styleguide

### Testing

 * rebuild this branch and bring it up in the browser
 * open identify point and verify that you see the new example elements and that they function as they should; compare them with what's in the style guide linked above
 * verify that you don't see any errors in the console


kellyi:
Still trying to get the pips stuff to work, but figured I'd assign it for a look at the current work.

kellyi:
I pushed a set of three new commits which:

1. adds the base jquery-ui-slider-pips library to the `lib` dir
2. adjusts jquery-ui-slider-pips to export a callable function rather than calling itself immediately
3. imports the updated pips library, then calls it on $ to add the "pips" and "floats" options
4. updates the example slider implementation to use the version with pips & floats

screen shot 2017-03-09 at 11 14 40 am

src/GeositeFramework/js/lib/jquery-ui-slider-pips.js
```diff
@@ -1,583 +1,566 @@
 /*! jQuery-ui-Slider-Pips - v1.11.4 - 2016-10-02
 * Copyright (c) 2016 Simon Goellner <simey.me@gmail.com>; Licensed MIT */
 
-(function( factory ) {
+var pips = {};
```
kellyi: Apologies for this diff; the high level summary of this is that:

1. it drops the IIFE factory function thing which pips would otherwise use to attach itself to `$.slider`
2. it exports the internal function it uses to extend $.slider as a callable function, which will then get called in `BasePlugin`'s constructor

Nothing internal to the function changed though, and almost all of the diff here just updates the function's lines to have the proper indentation.

src/GeositeFramework/js/PluginBase.js
```diff
@@ -79,6 +80,7 @@ define(["dojo/_base/declare",
             constructor: function(args) {
                 isBlacklisted = _.partial(_.contains, Geosite.app.data.region.identifyBlacklist);
                 declare.safeMixin(this,args);
+                if ($().slider) { pips.addjQueryUISliderPips($); }
```
caseycesari: Is this in a conditional because there's still a chance it might fail?

src/GeositeFramework/js/lib/jquery-ui-slider-pips.js
caseycesari: Consider splitting out the whitespace commit into a separate commit.

src/GeositeFramework/js/PluginBase.js
kellyi: I think this is a leftover from getting the imports set up. Let me drop it in a fixup commit and see if it ever does fail; it shouldn't now.

kellyi: Yeah -- this still works without the conditional.

kellyi:
Rebased this set of commits to squash the fixup and split the pips whitespace changes off from the pips module update.

kellyi:
Thanks for your help with this! Going to merge.

src/GeositeFramework/js/PluginBase.js
```diff
@@ -9,15 +9,20 @@ define(["dojo/_base/declare",
...
     function (declare,
                 xhr,
                 on,
                 dIdentifyTask,
                 IdentifyParameters,
                 dDeferredList,
-                Deferred
+                Deferred,
+                jqueryui,
+                chosen
```
kdeloach: This is missing a "pips" argument.

kellyi: Yeah -- this got lost in a rebase & fixed in #916

kdeloach: Great.

Issue #912, opened 2017-03-09
Plugins sidebar shouldn't become narrow on desktop automatically


caseycesari:
This is happening because the class is duplicated, and used within and outside of the context of a media query.

We should have a separate class with the same rules that can be used by the shrink/expand button.

PR #915, opened 2017-03-10
Fix sidebar toggle

caseycesari:
The initial implementation of this tool duplicated the CSS rules that
were contained in a media query to make the sidebar narrow. Doing this
meant that the sidebar became narrow at any screensize once a plugin was
activated. To fix this, a different class name is now used for the tool.

Also, the logic to check whether or not the sidebar is narrow was improved
here to take the size of the viewport into account.

To better keep the tool button icon in sync with the state of the sidebar,
changes to the viewport size are listened to and the icon is updated if
need be.

One side-effect of having a tool to manually toggle the state of the
sidebar is that users may have to expand the sidebar using the button
on small screens if the sidebar was initially toggled manually and any
active plugins are turned off.

Connects to #912

**Testing**
- With a plugin active, resize the viewport and verify that the plugin icon is update/kept in sync with the viewport changes.
- Verify that on large screens, when the app loads and the launchpad starts, the sidebar is not collapsed.
- Do the above on small screens, but verify that the sidebar is collapsed.
- Manually toggle the state of the sidebar using the plugin. Verify the it produces the expected behavior at different screen sizes and with plugins on and off.

### DIFF
--- a/src/GeositeFramework/css/main.css
+++ b/src/GeositeFramework/css/main.css
@@ -828,13 +828,13 @@ nav {
     width: 75%;
 }
 
-.nav-apps-narrow {
+.nav-apps-narrow-manual {
     width: 46px;
 }
-.nav-apps-narrow .button-text {
+.nav-apps-narrow-manual .button-text {
     display:none;
 }
-.nav-apps-button:hover .button-text:after {
+.nav-apps-narrow-manual .nav-apps-button:hover .button-text:after {
     content: "";
     top: 16px;
     left: -12px;
--- a/src/GeositeFramework/css/main.css
+++ b/src/GeositeFramework/css/main.css
@@ -1940,13 +1940,9 @@ body.x-body {
 #sidebar-help-area {
     position: fixed;
     bottom: 0;
-    padding: 4px 10px;
+    padding: 4px 5px;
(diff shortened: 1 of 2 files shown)


mmcfarland:
Added some commits to update variable names and fix a regression with the Tour link padding.

Reassigning to @kellyi so i'm not reviewing myself.

kellyi:
Changes generally look good! I tested this in Chrome, Firefox, and Safari on Desktop viewports, along with an iPad.

I did notice one bug/quirk which came back, which is that in some conditions the X close button now shows on minimized plugins when the sidebar's been narrowed. You can replicate this by:

1. loading the app and letting "Get Started" open
2. click the map control to narrow the sidebar
3. click on another plugin to open it
4. minimize that plugin so that it remains open but its div is hidden

The result is:

screen shot 2017-03-15 at 9 54 51 am

...and it seems to happen across browsers, on both tablets and desktop-sized viewports.

mmcfarland:
Rebased on `develop` and added a commit to handle nav-bar clear button when the app bar is manually narrowed.  Good catch.

kellyi:
Tested this again and it worked on both desktop and tablet-sized viewports.

Issue #910, opened 2017-03-09
'chosen' is not a function when app only has Identify Point and Get Started plugin


kellyi:
#903 tried to make `chosen` and jQueryUI available to plugins, but `chosen` returns an error if the regional-planning plugin isn't installed:

```js
Uncaught TypeError: $(...).chosen is not a function
    at Object.activate (main.js:128)
    at n.onSelectedChanged (Plugin.js:197)
    at n.<anonymous> (Plugin.js:338)
    at _ (backbone.js:368)
    at m (backbone.js:356)
    at f (backbone.js:155)
    at n.l.trigger (backbone.js:346)
    at n.select (backbone.picky.js:125)
    at n.toggleSelected (backbone.picky.js:157)
    at n.handleLaunch (Plugin.js:390)
```

I think this is probably a timing issue, as regional-planning makes some initial API calls which seems to delay the app loading.

One was to try fixing this is to update the `chosen.jquery.js` file the way #903 updated the `pips` file.

PR #916, opened 2017-03-10
Adjust jQuery library imports for Plugins

kellyi:
### Overview

This PR updates the jQuery library imports for `chosen` and `jquery-ui-slider-pips` to declare their dependencies explicitly in `Index.cshtml`. This in turn forces them to load in the correct order when imported into PluginBase.

As a consequence of these changes, we can revert the adjustments made to the `pips` module in #906 and the explicit function call, and also replace the patched version of `pips` with a minified version.

The second commit here fixes a typo in one of the `pips` options.

Connects #910 

### Screenshots

screen shot 2017-03-10 at 11 33 36 am

### Testing
 * remove all plugins which are not part of the base framework on your local
 * rebuild this branch in VS & view it in the browser with the console open
 * open the sample plugin and verify that the new UI elements display & work correctly: the slider should have pips, the accordion menu should work, and the select box should have chosen's style
 * add the regional planning and benefits explorer plugins, then rebuild this and open the sample plugin again. Verify that the sample plugin still has the correct appearance and behavior


kellyi:
Thanks!

Issue #917, opened 2017-03-10
Fix "Object doesn't support this action" error in IE


caseycesari:
The framework fails to load in IE11 on the latest `develop`.

![image](https://cloud.githubusercontent.com/assets/1042475/23804041/8e4eee2e-0586-11e7-98a4-f645fb7fdab2.png)

The [line of code](https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/master/src/GeositeFramework/js/Pane.js#L21) it is referencing hasn't been updated in a year, and the app worked in IE within the last couple months.

caseycesari:
A similar error happens in Chrome if you checkout an earlier commit in `develop`, like 39c20c3

![image](https://cloud.githubusercontent.com/assets/1042475/23804259/4a15029c-0587-11e7-87c3-2bac2c923bf9.png)

caseycesari:
Doing a `git bisect` points to ebb7e6045e64e3ee0061804000b22f51c45e288f as the breaking commit.

caseycesari:
If you inspect `N.models.Map` in `pane.js`, it is undefined. Not requiring and initializing the map utils plugin in `Map.js` fixes the app in Chrome while ebb7e60 is checked out. Requiring these plugins in `Map.js` may push the initialization of the `Map` module until some other parts of the framework are ready (because of some dependency of one of the plugins), which means it's not defined when `Pane.js` tries to create a `map`.

This also fixes the app in IE on `develop`. 

I'm next going to check how the app became fixed in Chrome even though the map util plugins are still being required.

caseycesari:
It also may be worth noting that I don't have the regional planning plugin installed in my local framework. This may have somehow been preventing this error, like it was preventing https://github.com/CoastalResilienceNetwork/GeositeFramework/issues/910.

caseycesari:
It appears that this error was fixed in Chrome by 8fc55b1282d0a9eb2ddb1c4fb375bada27e1f543.

caseycesari:
I tried delaying the initialization of the map in `Pane.js` with a `setTimeout`, but this causes other problems.

kellyi:
Checked this in Safari and Firefox without any plugins and both of them work (as does Chrome). Investigating IE.

kellyi:
In IE11 `N.models` doesn't have a Map key when this line gets executed:

[...]

PR #919, opened 2017-03-14
Drop explicit jQuery import in AgsMeasure.js

kellyi:
## Overview

This PR remedies a timing issue in IE whereby `N.models.Map` would not be available when `Pane.js` was being initialized, causing the error described in #917. That issue has a trail of more detailed comments, but the problem appears to be that the map utils, and the Measure tool specifically, take too long an interval to initialize in IE.

Drilling down through the imports list, it turned out that the `jquery` import in `AgsMeasure` caused things to hang up a bit, so this change drops the explicit import in favor of using the globally available `$`. I believe the explicit import might be a vestige of this being set up as a plugin rather than as part of the framework.

With these changes, the site will now load in IE. I also tested this in Chrome, Safari, and Firefox without any additional plugins installed and it loaded in each browser and the measure tool still works in each browser

Connects #917 

## Screenshots

Here's Chrome using the measure tool:

screen shot 2017-03-14 at 2 18 51 pm

Here's IE11 using the measure tool:

screen shot 2017-03-14 at 2 19 54 pm

## Notes

The "Geosite Framework Sample" text is a little bunched up on IE, so I'll make another card to fix that. (See #920)

## Testing

 * get this branch, `rm -rf` all plugins which are not part of the base framework, then load the site in IE11, Firefox, Chrome, and Safari and verify that
     * the site loads in each
     * the sample/identify point plugin has the proper styling
     * the maputils -- and the measure tool specifically -- still work
     * the rest of the app works as it should
 * add the regional-planning plugin, and repeat the testing steps above to verify that the app still works across browsers.

### DIFF
--- a/src/GeositeFramework/js/widgets/map_utils/measure/AgsMeasure.js
+++ b/src/GeositeFramework/js/widgets/map_utils/measure/AgsMeasure.js
@@ -1,6 +1,5 @@
 define([
         "use!Geosite",
-        "jquery",
         "esri/geometry/Polyline",
         "esri/geometry/Polygon",
         "esri/geometry/geodesicUtils",
@@ -17,7 +16,6 @@ define([
         "esri/layers/GraphicsLayer"
     ],
     function(N,
-             $,
              Polyline,
              Polygon,
              geodesicUtils,


kellyi:
Thanks! About to merge.

Issue #920, opened 2017-03-14
Header text bunched up in IE11


kellyi:
![screen shot 2017-03-14 at 2 29 01 pm](https://cloud.githubusercontent.com/assets/4165523/23916447/9e9bdffa-08c2-11e7-9d77-88992780df46.png)

PR #922, opened 2017-03-15
Drop flex & -ms-flex attributes on .nav-main-title

kellyi:
## Overview

This PR fixes a CSS issue in IE11 whereby the title header text would be stacked vertically & thereby overflow its div:

screen shot 2017-03-14 at 2 29 01 pm

This was happening because the `-ms-flex` attribute on `.nav-main-title` was being overriden by `flex: 1`. However, after checking into it it turns out that the text displays properly without either attribute, so I just dropped them both for the Framework & the Style Guide.

Connects #920

## Screenshots

Chrome ->

screen shot 2017-03-15 at 10 32 53 am

IE 11 ->

screen shot 2017-03-15 at 10 33 21 am

Safari ->

screen shot 2017-03-15 at 10 33 45 am

Firefox ->

screen shot 2017-03-15 at 10 34 04 am

Style guide ->

screen shot 2017-03-15 at 10 34 40 am

## Testing

 * rebuild this branch in VS then serve it to the browser
 * check the site in IE11, Chrome, and Firefox to verify that the header still works
 * also might be worth upping the length of the title in the base `region.json` file to verify that longer titles still render correctly

### DIFF
--- a/src/GeositeFramework/css/main.css
+++ b/src/GeositeFramework/css/main.css
@@ -1010,10 +1010,6 @@ header .nav-main {
     top: inherit;
 }
 .nav-main-title {
-    -ms-flex: 1;
-    flex: 1
-}
-.nav-main-title {
     font-size: 18px;
     font-weight: 900;
     color: #fff;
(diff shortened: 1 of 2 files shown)


kellyi:
Thanks!

Issue #924, opened 2017-03-16
Plugin nav bar close button shows inconsistently


mmcfarland:
The `X` to close a nav-bar plugin doesn't appear until all plugins are minimized, when it should be visible when a particular plugin is minimized.

To recreate:
1. http://lr01.internal.azavea.com:81/HUDSON_TNC_NY_Prototype_279/
2. Open Regional Planning
3. Click Risk Explorer, which opens that app and minimizes Regional Planning
4. Observe that Regional Planning does not display the `X`
5. Minimize Risk Explorer
6. Observe that _no_ plugins have the `X` where both Risk and Regional should
7. Toggle the "Expand Map" tool to minimize the sidebar, then again to expand it
8. Observe that all plugins show the `X` as desired.
9. Open Regional Planning
10. Observe that all `X` disappear, where it should only disappear from Regional Planning while it's open.

PR #926, opened 2017-03-16
Display plugin close buttons if sidebar isn't narrow

kellyi:
## Overview

This PR remedies a bug noted in #924 whereby the minimized-plugin-close-button would not appear in the sidebar unless all of the plugins were minimized. The bug was due to the `display: none` rule on `.nav-apps-narrow .button-sidebar-plugin-close` not being conditionally applied based on the viewport size.

The `.nav-apps-narrow` class is applied in the sidebar whenever any of the plugins are open; this was done to make it so that rules on the class could take effect immediately on tablets if a person rotated the tablet from landscape to portrait mode.

To fix it, I split the prior `display: none` rule for `button-sidebar-plugin-close` to be two rules: 

- first, we always hide the close buttons when the expand map/collapse sidebar button has been clicked to collapse the sidebar
- second, we hide the buttons if the viewport's narrower than 991px and any plugins are open, visible, and not minimized

Connects #924 

## Screenshots

Tablet, portrait mode, plugin minimized and another open ->

screen shot 2017-03-16 at 3 57 17 pm

Tablet, portrait mode, all plugins minimized ->

screen shot 2017-03-16 at 3 57 25 pm

Desktop, standard viewport, single plugin minimized ->

screen shot 2017-03-16 at 3 57 58 pm

Desktop, standard viewport, single plugin minimized & expanded map mode activated ->

screen shot 2017-03-16 at 3 58 05 pm

## Testing
 * rebuild this branch in VS with regional planning installed, then view it in the browser
 * verify that the bug described in #924 has been fixed and can't be reproduced
 * try a few combinations of toggling the expand button, opening, closing, and minimizing plugins to verify that the close button is hidden for minimized plugins when the sidebar's narrow and that it is visible and clickable for minimized plugins when the sidebar's standard-sized
 * using the device simulator to check the behavior for an iPad-sized device in both portrait and landscape mode, verifying that...
    * in landscape mode, the app behaves the same as it does on desktops
    * in portrait mode, the sidebar narrows automatically on opening a plugin -- and plugins which are minimized while another plugin is open do not have the close button if the sidebar's narrowed
    * note that if you manually expand the sidebar in portrait mode by clicking on the expand button, the minimized plugins should display a close button

### DIFF
--- a/src/GeositeFramework/css/main.css
+++ b/src/GeositeFramework/css/main.css
@@ -890,9 +890,13 @@ nav {
 .plugin-launcher.active + .button-link.button-sidebar-plugin-close {
   display: inline-block;
 }
-.nav-apps-narrow-manual .button-sidebar-plugin-close,
-.nav-apps-narrow .button-sidebar-plugin-close {
+.nav-apps-narrow-manual .button-sidebar-plugin-close {
   display: none !important;
+}
+@media (max-width: 991px) {
+    .nav-apps-narrow .button-sidebar-plugin-close {
+        display: none !important;
+    }
 }
 .plugin-launcher.app-displayed.active + .button-link.button-sidebar-plugin-close {
   display: none;


kellyi:
This PR's going to need to be updated before a review.

kellyi:
I updated this with a new commit and revised the initial PR comment.

kellyi:
Great! Thanks for taking a look.

Issue #911, opened 2017-03-09
Make a wiki page for any required js-pilot changes to a region


mmcfarland:
* Vector tile basemap shortcodes
* Enable dynamic layers for regional planning
* Build instructions

mmcfarland:
Shares a point with #909

kellyi:
Page is here:

https://github.com/CoastalResilienceNetwork/GeositeFramework/wiki/Updating-a-Region-for-the-ArcGIS-JavaScript-API-4.x-Pilot

Let me know if it needs and adjustments or additions.

mmcfarland:
+1, Page looks good

kellyi:
Thanks!

PR #927, opened 2017-03-16
Include additional vector basemap options

kellyi:
## Overview

This PR updates the basemaps list in `region.json` to include the other vector basemap options for demo purposes.

Connects #911 

## Testing
 * rebuild this branch in VS, then load it in the browser
 * click through the basemaps and verify that they all work


kellyi:
Thanks!

PR #928, opened 2017-03-17
Adjust 3d control CSS & disable search placeholder update

kellyi:
## Overview

This PR fixes two small bugs encountered after rebasing the `js-api-4x-pilot` branch on `develop`.

First, it adjusts the CSS rules for the 2-d/3-d toggle button to drop `font-size` and `line-height`. Some CSS updates made on `develop` on behalf of the expand control work to center the globe icon in the control button, and the extra rules in the toggle button's css file now push it off-center.

Second, it disables a routine which seems to handle internationalizing text for the search box. This code was showing an error in the console because `search.get('sources')` was returning an undefined object. I'm going to make another card for fixing it; for now I just commented it out to circumvent the error.

Connects #911 

## Testing
 * rebuild this branch in VS then open it in the browser and verify that everything still works as it should
 * it's probably worth using the 3-d Sample plugin alongside these changes to verify that that still works as well.

### DIFF
--- a/src/GeositeFramework/js/Map.js
+++ b/src/GeositeFramework/js/Map.js
@@ -252,6 +252,7 @@ require(['use!Geosite',
 
             // The translation lookup isn't ready when this is initialized.
             // A slight delay is needed.
+            /* Commenting this out for the js-api-4x-pilot branch
             window.setTimeout(function() {
                 // Required to set the placeholder text.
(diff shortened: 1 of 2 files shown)


kellyi:
Cool -- thanks for taking a look! I'll quote that comment on #929 so we can keep it around.

Issue #935, opened 2017-03-30
Change basemap selector to open on click, not hover


ajrobbins:
Currently, the basemap selector dropdown opens when the mouse hovers over it:

![screen shot 2017-03-30 at 10 50 37 am](https://cloud.githubusercontent.com/assets/12868342/24510290/cceab3ba-1536-11e7-979b-21d9af5ec57a.png)

This can lead users to accidentally open it pretty easily when they are just moving their mouse around. Since the menu covers the map controls, this is inconvenient.

The dropdown should open when the basemap title is clicked.

Note: this change should ultimately be applied to the `master` branch only, and not to the `v1_archive` branch.

PR #940, opened 2017-03-31
Adjust basemap selector behavior

caseycesari:
## Overview

The basemap list is now shown by clicking the basemap dropdown instead of hovering over it.

Per request from TNC.

The previous behavior was handled largely by CSS. To implement the requested behavior, additional event handling code needed to be added.

Connects #935

### Demo

tnc5

## Testing Instructions

- Verify that the basemap selector no longer opens when you hover over it.
- Click the basemap selector and verify it opens.
- Select a basemap.
- The basemap should update, and the basemap selector should close.


caseycesari:
Thanks

Issue #936, opened 2017-03-30
Slim down nav-apps sidebar


ajrobbins:
TNC is requesting that we reduce the fixed width of the app menu sidebar. It's currently at 240px; I think we could take it down to 220px:

![screen shot 2017-03-30 at 11 00 41 am](https://cloud.githubusercontent.com/assets/12868342/24510809/32a50eac-1538-11e7-8573-de276eba2b57.png)

cc @jfrankl

jfrankl:
+1

PR #941, opened 2017-03-31
Reduce the width of the sidebar

caseycesari:
## Overview

Per request from TNC.

Connects to #936

### Demo

Before:

image

After:

image

## Testing Instructions

- Verify that the sidebar is indeed reduced by 20px from it's previous width.
- Activate a plugin, then minimize it. Verify that the plugin text still wraps correctly so that the plugin close button does not cover up the plugin name.
- Open the header dropdown menu, and verify that it is the same width as the sidebar.
- Toggle the sidebar size manually with the expand map control. Verify that it is restored to the correct width.

### DIFF
--- a/src/GeositeFramework/css/main.css
+++ b/src/GeositeFramework/css/main.css
@@ -786,7 +786,7 @@ nav {
 }
 .nav-apps {
     text-align: left;
-    width: 240px;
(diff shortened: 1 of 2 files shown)


caseycesari:
Thanks

Issue #945, opened 2017-04-10
Sidebar collapsed by default


ajrobbins:
From TNC:
>Jeff and Arianna- thanks for the comments. I have discussed this with our core tech team and this is the desired, default behavior that we want: have sidebar collapsed at all times. So this means on default only app icons show, and Get Started is open. Then when a user hits a Go link it opens an app, but the sidebar is still closed. Then the user has the ability to open the sidebar with the new placement of the toggle button (which is great, BTW). We would like this to be implemented now for testing with project teams and stakeholders. We assume this is an easy behavior to change (let us know if not!); Get Started will be the initial screen that introduces the apps, with short explanation, so no need to have sidebar open.

PR #946, opened 2017-04-10
Collapse sidebar by default

kellyi:
## Overview

This PR sets the plugins sidebar to be collapsed by default unless explicitly expanded by clicking on the map control.

Connects #945 

### Notes

This works by adding the specified classes to the `nav` element initially; later toggles look to check whether those classes exist to do their toggling.

Also updated the style guide CSS to match.

## Testing Instructions

 * get this branch and check IE 11, Firefox, Safari, Chrome, and iPad Safari to verify that the following behavior holds:
    * on loading the app the sidebar is collapsed
    * on clicking to open, close, or minimize a plugin, or to switch plugins, the sidebar is collapsed
    * clicking the expand map control expands the sidebar and keeps it expanded through future plugin opens, closes, minimizes, and un-minimizes

### DIFF
--- a/src/GeositeFramework/Views/Home/Index.cshtml
+++ b/src/GeositeFramework/Views/Home/Index.cshtml
@@ -206,7 +206,7 @@
     </form>
 
     <script type="text/template" id="template-pane">
-        <nav class="nav-apps plugins nav-apps-narrow nav-apps-narrow-manual">
+        <nav class="nav-apps plugins nav-apps-narrow-manual">
             <div id="sidebar-help-area">
                 <a id="help-overlay-start" href="#" data-i18n="Tour">Tour</a>
             </div>
--- a/src/GeositeFramework/Views/Home/Index.cshtml
+++ b/src/GeositeFramework/Views/Home/Index.cshtml
@@ -206,7 +206,7 @@
     </form>
 
     <script type="text/template" id="template-pane">
-        <nav class="nav-apps plugins">
+        <nav class="nav-apps plugins nav-apps-narrow nav-apps-narrow-manual">
             <div id="sidebar-help-area">
                 <a id="help-overlay-start" href="#" data-i18n="Tour">Tour</a>
             </div>
--- a/src/GeositeFramework/Views/Home/Index.cshtml
+++ b/src/GeositeFramework/Views/Home/Index.cshtml
@@ -206,7 +206,7 @@
     </form>
 
     <script type="text/template" id="template-pane">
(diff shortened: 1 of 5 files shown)


kellyi:
One additional expected behavior for iPad/Safari in portrait mode:

- if the user has toggled the map to be collapsed/the sidebar to be expanded, opening a plugin currently minimizes the sidebar -- and minimizing/closing all plugins will expand the sidebar. This is leftover behavior from a tablet/portrait-mode change we made in December, but we could change that here too to take on same behavior as desktop browsers.

ajrobbins:
Based on TNC's comments that "the sidebar should be collapsed at all times", I think that minimizing/closing all plugins should *not* expand the sidebar. the sidebar should only be expanded with the toggle button.

caseycesari:
I can take another look when @ajrobbins's comment is addressed.

kellyi:
Just pushed a second commit here which drops everything related to the `nav-apps-narrow` class which was used to automatically set the sidebar width on tablets with viewports < 991px. 

Since the sidebar's now narrow by default, we no longer need to manually add and remove these classes.

Expected behavior now is that: for all viewport sizes, the sidebar is narrow unless the user has specifically toggled the map collapse button to make it enlarged. Toggling plugins should not impact whether the sidebar is expanded or collapsed in any case.

caseycesari:
Taking another look.

src/GeositeFramework/plugins/map_expand/main.js
```diff
@@ -64,13 +61,7 @@ define(
                 var docWidth = $(document).width(),
                     narrowWidthThreshold = 991;
```
caseycesari: I think these variables can be deleted.

src/GeositeFramework/Views/Home/Index.cshtml
```diff
@@ -206,7 +206,7 @@
     </form>
 
     <script type="text/template" id="template-pane">
-        <nav class="nav-apps plugins">
+        <nav class="nav-apps plugins nav-apps-narrow-manual">
```
caseycesari: Consider dropping `-manual` from the class name since there is only one class now.

kellyi:
Pushed a fixup to drop those unused variables and another commit to rename `nav-apps-narrow-manual` to `nav-apps-narrow`. Mind taking another look at this to ensure that it's still working? I checked it on my local and everything seemed to.

caseycesari:
Will do.

kellyi:
Thanks! Btw, I had to refresh my browser cache to get the CSS classname changes to appear.

PR #947, opened 2017-04-11
Collapse basemap list onclickoutside or on title

kellyi:
## Overview

This PR updates the basemap selector component to enable toggling the basemap list open and closed on clicking the title and to close it on clicking outside the div.

To make this happen, I added a click listener to the document from the basemap selector, then called the `hideBasemapList` method on any click which was not the list or the title.

Separately, the component now keeps track of the state of whether the list is hidden or visible, then use that state to toggle it open and closed on clicking the title.

Connects #944 

### Demo

tnc-close-basemap-selector-on-clicks

## Testing Instructions

 * verify that you can open and close the basemap selector list by clicking on the title
 * verify that selecting a basemap option closes the list
 * verify that clicking anywhere else in the app closes the list if it's open

### DIFF
--- a/src/GeositeFramework/js/BasemapSelector.js
+++ b/src/GeositeFramework/js/BasemapSelector.js
@@ -9,6 +9,17 @@
         render(view);
         // When the map's selected basemap changes, update the title element in the DOM
         view.model.on('change:selectedBasemapIndex', function () { renderSelectedBasemapName(view); });
+        view.model.set('basemapListVisible', false);
+        $(document).on('click', function(e) {
+           if (!_.contains(['basemap-selector-list', 'basemap-selector'], e.target.className)) {
+               hideBasemapList(view);
+           }
+        });
+        // Listen on the "Tour" link element to ensure the
+        // basemap selector closes before the tour overlay's applied
+        $('#help-overlay-start').on('click', function() {
+            hideBasemapList(view);
+        });
     }
 
     function render(view) {
@@ -33,16 +44,24 @@
         // DOM element's 'data-index' attribute tells us which item was clicked
         var index = $(e.currentTarget).data("index");
         view.model.set('selectedBasemapIndex', index);
-        hideBasemapList(view, e);
-    }
-
-    function showBasemapList(view) {
-        view.$el.find('.basemap-selector-list').show();
-    }
-
-    function hideBasemapList(view, e) {
-        view.$el.find('.basemap-selector-list').hide();
         e.stopPropagation();
+        hideBasemapList(view);
+    }
+
+    function toggleBasemapList(view, e) {
+        e.stopPropagation();
+        if (view.model.get('basemapListVisible')) {
+            view.model.set('basemapListVisible', false);
+            view.$el.find('.basemap-selector-list').hide();
+        } else {
+            view.model.set('basemapListVisible', true);
+            view.$el.find('.basemap-selector-list').show();
+        }
+    }
+
+    function hideBasemapList(view) {
+        view.model.set('basemapListVisible', false);
+        view.$el.find('.basemap-selector-list').hide();
     }
 
     N.views = N.views || {};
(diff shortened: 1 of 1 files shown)


kellyi:
Basemap list wasn't automatically closing on clicking the "Tour" link, so the second commit here adds a listener specifically for that.

src/GeositeFramework/js/BasemapSelector.js
```diff
@@ -15,6 +15,9 @@
                hideBasemapList(view, e);
            }
         });
+        $('#help-overlay-start').on('click', function(e) {
+            hideBasemapList(view, e);
```
caseycesari: I'm curious why we need to create a listener just for this element? I thought it would be captured by the listener you created in the previous commit that is listening to clicks on `document`. My thought is that if that listener isn't capturing all the clicks, we may have to add more of these special case listeners.

kellyi: Wasn't sure either -- I think something about how that thing in particular is set up may be intercepting click events on `document`. I'd guess that would happen because it's essentially placing an overlay on top of the whole page with elements which may stop clicks from propagating through?

src/GeositeFramework/js/BasemapSelector.js
```diff
@@ -36,11 +45,18 @@
...
+        } else {
+            view.model.set('basemap-list-visible', true);
+            view.$el.find('.basemap-selector-list').show();
+        }
     }
 
     function hideBasemapList(view, e) {
+        view.model.set('basemap-list-visible', false);
         view.$el.find('.basemap-selector-list').hide();
         e.stopPropagation();
```
kellyi: Just realize that this line will intercept map clicks from every element now, so going to push a fixup to adjust how it gets called.

src/GeositeFramework/js/BasemapSelector.js
```diff
@@ -9,6 +9,15 @@
         render(view);
         // When the map's selected basemap changes, update the title element in the DOM
         view.model.on('change:selectedBasemapIndex', function () { renderSelectedBasemapName(view); });
+        view.model.set('basemap-list-visible', false);
+        $(document).on('click', function(e) {
+           if (!_.contains(['basemap-selector-list', 'basemap-selector'], e.target.className)) {
+               hideBasemapList(view);
+           }
+        });
+        $('#help-overlay-start').on('click', function() {
```
caseycesari: Consider adding a comment as to why this is required.

src/GeositeFramework/js/BasemapSelector.js
```diff
@@ -9,6 +9,15 @@
         render(view);
         // When the map's selected basemap changes, update the title element in the DOM
         view.model.on('change:selectedBasemapIndex', function () { renderSelectedBasemapName(view); });
+        view.model.set('basemap-list-visible', false);
```
caseycesari: Model attributes should be in camelCase. See line 44 for an example.

src/GeositeFramework/js/BasemapSelector.js
```diff
@@ -33,24 +42,32 @@
...
-        hideBasemapList(view, e);
+        e.stopPropagation();
+        hideBasemapList(view);
     }
 
-    function showBasemapList(view) {
-        view.$el.find('.basemap-selector-list').show();
+    function toggleBasemapList(view, e) {
+        e.stopPropagation();
+        if (view.model.get('basemap-list-visible')) {
```
caseycesari: Same here, regarding camelCase.

kellyi:
Pushed a fixup to add a comment about the tour link event listener along with another to rename `basemap-list-selector` -> `basemapListSelector` as the model key.

kellyi:
Thanks! Going to squash then merge.

kellyi:
Squashed this down to one commit with an itemized list of changes.

PR #951, opened 2017-04-13
Add WMS Server needed for future_habitat_v2

lesserj:
We need to use the WMS Server provided by mapserver.maine.gov as requested for the future_habitat_v2 app.  They provide LIDAR data in our region of interest.

## Overview

Adds a rule for the mapserver.maine.gov WMS server

Connects #XXX

## Testing Instructions

Build the future_habitat_v2 app with this change and turn on the LIDAR layer.  LIDAR layer should be proxied through the CoastalResilience proxy and displayed on the map.

### DIFF
--- a/src/GeositeFramework/proxy.config
+++ b/src/GeositeFramework/proxy.config
@@ -22,6 +22,7 @@
     <serverUrl matchAll="true" url="http://lr13:6080/"></serverUrl>
     <serverUrl matchAll="true" url="http://sampleserver1.arcgisonline.com"></serverUrl>
     <serverUrl matchAll="true" url="http://preview.grid.unep.ch:8080"></serverUrl>
+    <serverUrl matchAll="true" url="http://mapserver.maine.gov"></serverUrl>
   </serverUrls>
 
 </ProxyConfig>


mmcfarland:
Thanks for setting up a PR and using the template @lesserj.  I think there may be a more localized way you can get this proxy source into the app.  The build tool will copy over specific files from the region repo into the installed application, overwriting any that are the default in this repository (for a list of files and directories that are copied, see https://github.com/CoastalResilienceNetwork/geosite-framework-build/blob/master/build.py#L225-L230)

You should be able to take a `proxy.config` base, remove the servers that are not relevant and add the new ones that are, and place it in the root of your region repo.  On the next build, that `proxy.config` should be available (this is identical to how the `region.json` file works).  The benefits are that the framework can stay agnostic of the dependencies of the various regions, and the regions won't need to wait on a release of the framework for new configuration (although it looks like the default has developed a fair amount of cruft over the years).

Let me know if you think that solution will work for you.

lesserj:
Thanks @mmcfarland I wasn't sure if some of the resources in the proxy were used by any of the base functionality in the framework.  I didn't want to overwrite the proxy.config if some of those servers were needed globally.  If not, I'm happy to add this in the region repo.

mmcfarland:
That's a valid concern.  To the best of my knowledge, apart from the `*.arcgisonline.com` entries, the default values are experimental, in the sense that they helped us debug services added to the layer selector over the years in various regions.  To play it conservatively, you can keep an intact version of the default file and just append your server, as you've done for this PR.  My sense is that for the majority of regions, the defaults are superfluous, but I am interested in your experience if you do make changes locally to your region.

lesserj:
Great.  I'll let you know if we run into any trouble.  Meanwhile, will close this request.  Thanks!

PR #955, opened 2017-04-18
Fix app-print cleanup

lesserj:
Currently the app-print.css file, with class 'app-print-css' is not getting removed from the app when closing the print dialog because the class is created with a "." in it.  The class shouldn't be prepended with a "." in the class attribute.  Removing fixes the problem and the css file is correctly removed on cleanup. c.c. @zferdana

## Overview

Fixes the naming of the app-print-css class so the app can correctly reference it

## Testing Instructions

Open a plugin with print functionality.
Open your browsers dev tools and expand the head section in elements.
Click the print button in the app.
See css/app-print.css file with class=".app-print-css" added to the head.
Click close on the print dialog.
The css/app-print.css file is not removed from the document head.  Opening the print dialog again with add another copy of the css/app-print.css file.

### DIFF
--- a/src/GeositeFramework/js/Export.js
+++ b/src/GeositeFramework/js/Export.js
@@ -10,7 +10,7 @@ require(['use!Geosite'],
             $('<link>', {
                 rel: 'stylesheet',
                 href: 'css/app-print.css',
-                'class': '.app-print-css' 
+                'class': 'app-print-css' 
             }).appendTo('head');
             var html = N.app.templates['template-export-window']();
             view.$el.empty().append(html);


mmcfarland:
Thanks for submitting a PR, @lesserj.  This looks like the right fix, but the PR is targeting the `master` branch.  Could you retarget to `develop`?  You may need to cherry-pick your commit to your local `develop` branch since `master` has additional commits (the added and reverted) which aren't in `develop`.

Let me know if that makes sense.

lesserj:
Thanks @mmcfarland That mismatch between develop and master was my fault.  My apologies again for screwing that up.  This should be pointing to the develop branch now.

caseycesari:
This fix looks good as well, though this PR also has extra commits. If you could also remove the commits that aren't relevant from this PR, that would be great. I can also do it if that works better for you.

caseycesari:
Whatever process you followed on the other PR to clean-up the commits should work here as well.

lesserj:
@caseycesari Should be good to go.  Thanks!

lesserj:
@caseycesari Ok.  Now it really is good to go.  Thanks.

caseycesari:
Thanks!

PR #956, opened 2017-04-20
CSS & JS version mismatch for Jquery-UI

lesserj:
The version of jquery-ui that is loaded is 1.12, however, the css version was 1.11.3.  This broke some jquery-ui elements, particularly tooltips, which only showed with 0.3 opacity.  Ensuring the versions match fixes the jquery-ui rendering issues.  c.c. @zferdana @cschneebeck @iKimbot

## Overview
Fixes broken jquery-ui tooltips, and likely other jquery-ui issues

## Testing Instructions

Add a jquery-ui tooltip element following the Styleguide.
Hover over the tooltip.
The tooltip renders, but is mostly transparent and the content is unreadable


lesserj:
Repointed the pull request to the develop branch.

caseycesari:
@lesserj Before merging would you mind cherry-picking d44e0f7 on top of the current `develop` and then force pushing this branch up again? That way the merge commit and the proxy change and revert commits are dropped from this PR.

I can walk you through that if you need help.

lesserj:
@caseycesari I think I did it right, but please confirm.

caseycesari:
Thanks! Good to go. Feel free to merge whenever you are ready.

caseycesari:
Actually, I'm just going to go ahead in merge it in case you don't have permissions to. Thanks again for the fix!

PR #962, opened 2017-05-24
dropped height from sidebar-content

cschneebeck:
The height property (even without calc) hides elements in some apps. Removing it makes them show up. It appears to have no effect on other apps.

### DIFF
--- a/src/GeositeFramework/css/main.css
+++ b/src/GeositeFramework/css/main.css
@@ -769,7 +769,6 @@ nav {
 }
 .sidebar-content {
     position: relative;
-    height: calc(100% - 40px);
     overflow-y: auto;
     margin-top: 40px;
     width: 100%;


ajrobbins:
Hi @cschneebeck - thanks for making this fix. Can you provide some steps to replicate the original issue, and to demonstrate how the reviewer can check that this fix was made? What devices/OS to test on, etc. Thanks!

cschneebeck:
1. Go to http://dev.maps.coastalresilience.org/northcarolina/# using Safari (desktop to show problem, iOS to make sure it worked afterwards). 
2. Open CRS Explorer. 
3. Observe the nothingness of the app window.
4. Show web inspector and choose the blank div where the app should be.
5. In the Elements section of developers tools, click on the child div of id="plugins/community-rating-system-0" with class sidebar-content.
6. In the Styles portion of developers tool, find the styles for sidebar-content.
7. Uncheck the height property. 
8. Watch the nothingness in the app window turn into something.
9. Turn on other apps and toggle the height property on and off. This will test if the height property has an impact on any apps. If you like to be thorough, do the same thing in other regions.
10. If you decide the height property has no use in any of the apps, get rid of it in the main.css file of the framework (line 770).
11. Rebuild and test on desktop and iOS versions of Safari. If you can see something when you open CRS Explorer then things are good.

mmcfarland:
I'm not sure that this is the appropriate fix.  It certainly does make the content show up for CRS in Safari, but so does removing the `overflow` attribute and leaving the `height`, for example. However, I don't think the `height` or `overflow` properties are being misused here. Safari supports `calc` as far back as version 6 and doesn't seem to be the culprit, other apps are loading fine under the same conditions.  

The purpose of having that height attribute is so that the content is "full height" within the container.  The drawback to removing it is that the height of the `sidebar-content` element is then sized to the content, instead of the container.  In practical terms, this means you can't place things lower than the content, and would have to write javascript to resize your container height manually on every window resize.

Additionally, removing the height doesn't fix related issues like the icon showing up in the header or applying the appropriate padding:
**Chrome**
screenshot from 2017-05-26 13 46 58
**Safari (no height)**
screenshot from 2017-05-26 13 47 27

I would suggest evaluating the layout markup for this app compared to other apps that are working and trying to identify salient differences between them.

cschneebeck:
Sounds reasonable to me. I’ll take a look the week after next.

Cheers,
Casey Schneebeck
Web Developer | The Nature Conservancy
North America Science | Boulder, CO
415-722-9768

cschneebeck:
From my reading Safari doesn't consistently support calc in CSS from version 6 on. Even adding a webkit prefix to calc does not do the trick. Since we don't want to change that in the framework, I looked into a way to work around it in my apps.

When I build an app I create a new ContentPane and add it to the parent container of the app (a div with class sidebar-content). Then I push all my HTML into the ContentPane. The parent container's height is determined by sidebar-content using calc. Since calc doesn't play well in Safari 6+, the height of the div with sidebar-content class has a height of zero. Combine that with an overflow-y of auto (also part of the sidebar-content class) and you don't see most elements in my technique of building apps.

screen shot 2017-06-08 at 10 26 32 am

screen shot 2017-06-08 at 10 37 21 am

The solution I've found is to add a min-height to the ConentPane that houses my HTML. When min-height is added, the div with class sidebar-content magically expands to that height. Once a selection is made and new elements are displayed, the sidebar-content div expands beyond the min-height with no issue.

screen shot 2017-06-08 at 10 45 53 am

I did this on dev and prod for the community-rating-system app and everything is working now. To my understanding this was never a tablet issue, it is a Safari version 6+ issue. When you combine the framework's use of calc, my ContentPane workaround (adding HTML to an AMD environment), and Safari 6+ it doesn't work. I don't think it would be hard to update to the framework to fix this issue (figure out how to get full-height without calc), but it doesn't seem like a good use of money when all Matt S and I have to do is add a min-height property to our apps.

mapossum:
Casey.  I used a significant amount of your CRS code to start an app and I am having a similar issue.  What is the CSS code you are changing to get it to work?  I don't have anything safari to test on I've tried a few things but I must not be doing it right.  Thanks

cschneebeck:
I just figured out a fix yesterday. Check out where I create the content pane in main.js. Adding a large min height to it seemed to help it show up. I'm heading to a concert so can't get on a computer but hopefully that helps.

Casey

On Jun 9, 2017, at 12:12 PM, George Raber <notifications@github.com<mailto:notifications@github.com>> wrote:


Casey. I used a significant amount of your CRS code to start an app and I am having a similar issue. What is the CSS code you are changing to get it to work? I don't have anything safari to test on I've tried a few things but I must not be doing it right. Thanks

-

Issue #958, opened 2017-04-24
Drop shadow covers small sidebar


lesserj:
When opening up an app with the minimized version of the toolbar, the toolbar turns white and the apps are not visible.

![image](https://cloud.githubusercontent.com/assets/2304500/25351227/de84c912-28dc-11e7-95c7-595f52286c69.png)

The offending css appears to be line 743 of main.css

.sidebar {
  box-shadow: 0px 0px 26px 0px rgba(51, 51, 51, 0.42);
  z-index: 2000;
}

Disabling the box-shadow css corrects the issue.  That does get rid of the box shadow in the app however.  The css needs to be tweaked to preserve the box shadow without covering the toolbar.

c.c. @zferdana @iKimbot

lesserj:
Here's a video of this occurring.  The rendering issue seems to be dependent on screensize.  It is not always consistent.  I think the safest thing for now is to disable the box-shadow until we can figure out what's going on exactly.

![sidebar-box-shadow](https://cloud.githubusercontent.com/assets/2304500/25359323/9e72854a-28f9-11e7-93bf-3c41f0ca369e.gif)

I've noticed this on both Protecting Water and Ocean Wealth.  I have not been able to reproduce in my local environment.

lesserj:
If I change the z-index value of the sidebar class from 2000 to 1, the dropshadow is preserved, but the issue is no longer present.  I haven't been able to figure out why (yet).

lesserj:
Seems to be some sort of overflow issue with the sidebar-nav header.  Very strange.

![header-overflow](https://cloud.githubusercontent.com/assets/2304500/25362525/4b2a8de6-2909-11e7-8df1-ad5a31e8823e.gif)

kdeloach:
We are unable to reproduce this error. Could you provide any more details on how to reproduce this? There are some other style errors in your screenshot which we don't see (ex. the "expand sidebar" button is misplaced). We should rule out anything specific to your system which could be causing this error.

lesserj:
[...]

PR #968, opened 2017-08-15
Reduce sidebar z-index to fix dropshadow bug

caseycesari:
## Overview

Changing the z-index from 2000 to 1 fixes the dropshadow white-out bug,
but does not impact the visual style of the app.

Fix provided by @lesserj.

Connects to #958

## Testing Instructions

I haven't been able to reproduce this locally, but I was able to test the fix by doing the following

- Visit http://maps.coastalresilience.org/hawaii/
- Activate the "Ecosystem Effects of Sea Level Change".
- Note the white sidebar
- Apply the change from this PR to the stylesheet in chrome.
- The white sidebar should go away.

### DIFF
--- a/src/GeositeFramework/css/main.css
+++ b/src/GeositeFramework/css/main.css
@@ -742,7 +742,7 @@ nav {
 }
 .sidebar {
   box-shadow: 0px 0px 26px 0px rgba(51, 51, 51, 0.42);
-  z-index: 2000;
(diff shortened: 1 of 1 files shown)


caseycesari:
All credit for the fix goes to @lesserj. I didn't spend much time figuring out why it worked, but that it didn't cause any additional problems; which seems to be the case.

Issue #966, opened 2017-08-10
Launch super web app mode


caseycesari:
Based on the configuration created in #965, launch super web app mode. This includes:

- Launching the appropriate app
- Hiding the sidebar
- TBD

Issue #965, opened 2017-08-10
Devise configuration for super web app mode


caseycesari:
This may only be a single block of configuration which specifies which plugin is active.

PR #970, opened 2017-08-17
Implement single plugin mode

caseycesari:
## Overview

Takes a first pass at implementing single plugin mode. There is some follow-up work to do, but the basics are implemented here. See commit messages for details.

Connects #966 
Connects #965 

### Demo

image

### Notes

Follow-up work:

- Decide on launchpad implementation
- Disable subregions in single plugin mode
- Remove plugin minimize/close buttons.

## Testing Instructions

* Launch the site and verify it is in single plugin mode. The sidebar should be hidden, the layer picker should be active, and the app title should mention single plugin mode.
* Update the region config to disable single plugin mode. Reload the site, and confirm the app loads and works as normal.

### DIFF
--- a/src/GeositeFramework/Models/Geosite.cs
+++ b/src/GeositeFramework/Models/Geosite.cs
@@ -4,6 +4,7 @@ using System.Drawing;
 using System.IO;
 using System.Linq;
 using System.Text.RegularExpressions;
+using System.Web.Helpers;
 using GeositeFramework.Helpers;
 using Newtonsoft.Json.Linq;
 
@@ -60,6 +61,7 @@ namespace GeositeFramework.Models
         public String TertiaryColor { get; private set; }
         public String ActiveAppColor { get; private set; }
         public String PrintHeaderLogo { get; private set; }
+        public Boolean SinglePluginMode { get; private set; }
 
         /// <summary>
         /// Create a Geosite object by loading the "region.json" file and enumerating plug-ins, using the specified paths.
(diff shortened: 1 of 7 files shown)


caseycesari:
>Unsure if we need the mode config to be as complex as it is, and not entirely sure of the scope of the attached cards. Main issue I noticed is that the min/close buttons are still available.

Yeah, I just wanted to get the basics working, and not get too far ahead of the wireframes.

src/GeositeFramework/Models/Geosite.cs
```diff
@@ -142,8 +142,23 @@ private static List<JsonData> GetPluginConfigurationData(IEnumerable<string> plu
             jsonObj.Add("pluginFolderNames", new JArray(PluginFolderNames.ToArray()));
 
             // Set public properties needed for View rendering
-            TitleMain = ExtractLinkFromJson(jsonObj["titleMain"]);
-            TitleDetail = ExtractLinkFromJson(jsonObj["titleDetail"]);
+            if (SinglePluginMode)
+            {
+                TitleMain = new Link
+                {
+                    Text = (string)jsonObj["singlePluginMode"]["title"],
```
caseycesari: You're right. This probably isn't needed. I was originally thinking they would want to switch it on and off depending on the deployment, but that probably isn't the case. I'll revise.

caseycesari: Okay. Ditched this commit.

Issue #931, opened 2017-03-18
Update tests


mmcfarland:
We haven't been running the unit test suite via CI on develop and merging to master has exposed some failures, likely due to region.json updates that aren't reflected in the tests:

http://jenkins01.internal.azavea.com/view/TNC/job/GeositeFramework/599/console

PR #975, opened 2017-08-23
Fix unit tests

caseycesari:
## Overview

Over time, `region.json` has changed as new features were added to the app. The unit tests weren't always updated in tandem and became stale. Also, `regionSchema.json` was no longer valid JSON, and due to some errors with brackets, some of the properties of the schema weren't actually included in the schema. In this PR, the tests are updated and `regionSchema.json` is repaired.

Connects #931 

### Demo

screen shot 2017-08-23 at 3 43 45 pm

### Notes

I tried to break up the changes to `regionSchema.json` so that it would be a little easier to see what changed. But, since a lot of changes are related to whitespace, it's still hard to see.

## Testing Instructions

- Run the unit tests and verify that they pass.

### DIFF
--- a/src/GeositeFramework/App_Data/regionSchema.json
+++ b/src/GeositeFramework/App_Data/regionSchema.json
@@ -51,7 +51,7 @@
                 "type": "object",
                 "properties": {
                     "text": {"type": "string", "required": true},
-                    "url": {"type": "string", "required": true},
+                    "url": {"type": "string", "required": true}
                 },
                 "additionalProperties": false
             }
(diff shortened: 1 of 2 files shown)


mmcfarland:
Tests are passing but the checked in version of the `region.json` doesn't seem to be in compliance with the updated schema:

```
Property 'singlePluginMode' has not been defined and the schema does not allow additional properties. Line 4, position 24.
Property 'legend' has not been defined and the schema does not allow additional properties. Line 152, position 14.
```

caseycesari:
Good catch. I'll check it out.

caseycesari:
Pushed a fixup commit. Should be good

caseycesari:
Thanks

PR #976, opened 2017-08-24
Remove single plugin mode title

caseycesari:
## Overview

We decided this was extraneous in #970.

Essentially reverts bf5451a, which should have been reverted in #970.

Connects #966

### Notes

This should have been removed in #970, but I must have not force pushed the update branch.

## Testing Instructions

- Load up the app, verify it is in single plugin mode, and that the app has the standard title.


caseycesari:
Thanks

Issue #967, opened 2017-08-10
Demo passing of configuration to plugin in super web app


caseycesari:
If it's not already happening, we should be passing the region configuration to each plugin. Then, using the configuration, a plugin could see if it super web app mode was active and show additional text, etc.

PR #977, opened 2017-08-29
Demo single plugin mode

caseycesari:
## Overview

Demos single plugin mode by showing a custom message in the identify plugin. Also, some basic guards were added when parsing the config for single plugin mode.

The identify plugin is patched up so that it can function as a stand alone plugin. It previously relied on other plugins bringing in jQuery. 

Connects #967 

### Demo

image

## Testing Instructions

- Load the app, which should be in single plugin mode. Verify that the identify point plugin loads and displays some text about single plugin mode. 
- Turn off single plugin mode, and reload the app. Open the identify plugin, and verify that text about single plugin mode is not displayed.
- In the config, enter an invalid plugin name for single plugin mode. Verify that the app does not load.
- In the config, remove the plugin name for single plugin mode. Verify that the app does not load.

### DIFF
--- a/src/GeositeFramework/sample_plugins/identify_point/main.js
+++ b/src/GeositeFramework/sample_plugins/identify_point/main.js
@@ -94,7 +94,7 @@ define(["dojo/_base/declare", "framework/PluginBase"],
                 if (showHelpOnStart) {
                     this.showHelp();
 
-                    // Don't show this help on startup anymore, after the first time 
+                    // Don't show this help on startup anymore, after the first time
                     this.app.suppressHelpOnStartup(true);
                 }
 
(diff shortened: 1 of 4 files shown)


src/GeositeFramework/sample_plugins/identify_point/main.js
```diff
@@ -22,6 +22,11 @@ define(["use!Geosite", "dojo/_base/declare", "framework/PluginBase"],
 
             initialize: function(args) {
                 declare.safeMixin(this, args);
+
+                if (N.app.singlePluginMode) {
```
mmcfarland: I don't think we've previously advertised the internal api to the apps.  Perhaps this setting should be transferred to the `app` object that comes in `initialize(args)` so we can maintain some independence with the internal implementation.

caseycesari: Good call. Pushed some fixup commits.

caseycesari:
Thanks.

Issue #979, opened 2017-09-07
Hide "Create map" option in map tools menu


caseycesari:
All the other options should be persisted.

![image](https://user-images.githubusercontent.com/1042475/30179236-ab898b02-93d9-11e7-92fa-cc42ccb91806.png)

PR #980, opened 2017-09-08
Hide create map tool in single plugin mode

caseycesari:
## Overview

Hides the "create map" (aka print) tool when in single plugin mode. Also cleans up some somewhat related legacy code.

Connects #979 

### Demo

image

## Testing Instructions

- Activate single plugin mode, and verify that the "create map" tool is hidden in the map tools menu.
- Turn off single plugin mode, and verify that the "create map" tool is displayed.


caseycesari:
Yeah, good call. I believe TNC requested this because they want users to print via the "plugin print" feature. I'll update the commit message.

Issue #973, opened 2017-08-22
Implement app minimize 


caseycesari:
![image](https://user-images.githubusercontent.com/1042475/30178761-1f537a54-93d8-11e7-86de-17e11ea8bd55.png)

![image](https://user-images.githubusercontent.com/1042475/30178769-2631aab2-93d8-11e7-8ca4-e66628a2ae29.png)

PR #981, opened 2017-09-11
Implement "expand sidebar" button

caseycesari:
## Overview

Implements the "toggle sidebar" button which hides/restores the plugin in single plugin mode. Also removes some of the plugin UI in single plugin mode to match the wireframes.

Connects #973

### Demo

tnc1

### Notes

~~As you will note in the demo video above, the CSS for the plugin minimize button is not final. It's tricky to have an element that lives outside the plugin content (so that it is outside the scope of the plugin) but does not interfere with the plugin contents. I'll need to consult with @jfrankl.~~

## Testing Instructions

- Launch the app with single plugin mode enabled.
- Minimize the activate plugin using the "Hide sidebar" plugin.
- Use the "Expand sidebar" button to restore the plugin.
- Verify that these buttons are not displayed in regular mode.

Also looking for feedback on:
- Use of razor templating within underscore templates for single plugin mode elements and style. It's a little confusing, but I think results in cleaner templates (instead of lots of if/else in JS templates).
- Location of "toggle sidebar" button code. I wasn't sure of the best place for this.

### DIFF
--- a/src/GeositeFramework/Views/Home/Index.cshtml
+++ b/src/GeositeFramework/Views/Home/Index.cshtml
@@ -175,6 +175,27 @@
         .dropdown-submenu > a:not(:hover) {
             background-color: @Model.SecondaryColor;
         }
+
+        @if (Model.SinglePluginMode) {
+        <text>
+        .sidebar-content {
+            height: 100% !important;
+            margin-top: 0 !important;
+        }
+
+        .sidebar-nav {
+            display:none;
+        }
+
+        #toggle-plugin-container button {
+            background-color: @Model.PrimaryColor;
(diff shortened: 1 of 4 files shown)


mmcfarland:
Implementation looks good, but I'd like to revisit the visual design of it.

caseycesari:
Updated with the new design from @jfrankl. Demo below:

tnc2

ajrobbins:
I can't quite tell but the gif makes it look like the drop shadow on the sidebar is on top of the collapse button - is that the case? Otherwise, looking good to me!

caseycesari:
Good catch. Fixed.

image

caseycesari:
Thanks

Issue #974, opened 2017-08-22
Implement help screen


caseycesari:
- Add link in header
- Open up panel over plugin
- Add place for content. Maybe a template?

![image](https://user-images.githubusercontent.com/1042475/30178729-ffe7936c-93d7-11e7-9800-bf8c5be98e53.png)

PR #983, opened 2017-09-14
Implement single plugin mode help pane

caseycesari:
## Overview 

Add mechanism for developers to provide custom help content in single
plugin mode.

Where possible, code is only included on the client when single plugin
mode is active.

Connects #974

### Demo

tnc1

## Testing Instructions

- Launch single plugin mode, and verify the app subtitle is replaced with the word "About".
- Click "About", and verify that the help panel opens.
- Launch the app in regular mode, and verify that the subtitle is back.

### DIFF
--- a/src/GeositeFramework/js/SinglePluginModeHelp.js
+++ b/src/GeositeFramework/js/SinglePluginModeHelp.js
@@ -1,0 +1,48 @@
+(function (N) {
+    'use strict';
+
+    var $container;
+    var $pluginToggle;
+    var singlePlugin;
+
+    function initialize(view, $el) {
+        $container = $el;
+        $pluginToggle = $('#toggle-plugin-container');
+
+        // Get the content from the developer-provided HTML partial
+        var content = $('#single-plugin-mode-help-content').html();
+
+        // Get the template for the help pane
+        var tmpl = _.template($('#single-plugin-mode-help-template').html());
+
+        // Insert the template, with the custom contents, to the help pane.
+        $container.html(tmpl({ content: content }));
+
+        // It's a bit odd to have this here, but there isn't a more discoverable
+        // place for it to be.
+        $('#show-single-plugin-mode-help').click(function() {
+            $container.show();
+            $pluginToggle.hide();
+            singlePlugin.get('$uiContainer').hide();
+        });
+    }
+
+    function close(view) {
+        $container.hide();
+        $pluginToggle.show();
+        if (singlePlugin.selected) {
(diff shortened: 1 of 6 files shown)


mmcfarland:
I'm not sure if this is desired behavior.  If the plugin isn't exactly the same width as the help, they both show up.  We may want to toggle the visibility of the plugin when the help content is toggled.

screenshot from 2017-09-20 14 37 16

caseycesari:
Rebased on `develop`.

mmcfarland:
> What about making the help pane as wide as the plugin?

I guess that assumes the content of the help will look good at that width.  Do they have the ability to override the help width like a plugin has?

caseycesari:
> I guess that assumes the content of the help will look good at that width. Do they have the ability to override the help width like a plugin has?

No, it's static at the moment.

I'll try and toggle the plugin visibility.

caseycesari:
Updated to hide the plugin when help is activated. Also included some fixes related to merging #981

src/GeositeFramework/js/SinglePluginModeHelp.js
```diff
@@ -2,9 +2,13 @@
     'use strict';
 
     var $container;
+    var $activePlugin;
+    var $pluginToggle;
 
     function initialize(view, $el) {
         $container = $el;
+        $activePlugin = $('.sidebar:visible');
```
mmcfarland: This isn't actually selecting the plugin on `initialize`, doesn't look like it's in the DOM yet.  This plugin isn't actually hidden as a result.

src/GeositeFramework/js/SinglePluginModeHelp.js
```diff
@@ -19,11 +23,15 @@
         // place for it to be.
         $('#show-single-plugin-mode-help').click(function() {
             $container.show();
```
mmcfarland: If you move the `$activePlugin = $('.sidebar:visible');` assignment to the click handler, but above the `$container.show`, it finds it and hides it appropriately.  After `$container` is visible, that selector would find two items (help & plugin).

src/GeositeFramework/js/SinglePluginModeHelp.js
caseycesari: Weird. I must have screwed something up before checking it in. It was hiding for me initially. Sorry. I'll take another look.

caseycesari:
Ready for another view. Reworked this a bit.

caseycesari:
Thanks

Issue #984, opened 2017-09-18
Make framework print css overrideable


caseycesari:
Some of the rules in the framework [print.css](https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/master/src/GeositeFramework/css/print.css) use `!important`, which make them tough to override in the case where plugins want to implement custom print.

We should find a way to implement these rules without using `!important`. We wouldn't have used that flag if it wasn't needed, so this may require reworking some of the base framework css.

caseycesari:
@lesserj From your work with the Maine region, do you recall which style rules gave you the most trouble, or that you wanted to override? I see that you provided a [custom stylesheet](https://github.com/CoastalResilienceNetwork/maine-region/blob/master/css/print-landscape.css) to replace the Framework's `print-landscape`. Are there any other specific rules that got in your way?

I'm going through the existing sheets to see which `!important` flags can be removed, and it looks like a targeted approach may be more fruitful.

caseycesari:
I put up a WIP PR that removes`!important` from the rules in the portrait and landscape stylesheets. It appears we were a bit egregious in our use of that flag in these two files. You can view that here: https://github.com/CoastalResilienceNetwork/GeositeFramework/pull/988. I tried to remove some of the `!important` flags from `print.css` and `app-print.css`, but that quickly lead to problems in the "Create Map" print workflow. 

Also, I wasn't able to switch the units in the landscape and portrait stylesheets over to percentages. I used the updated values that were provided in the Maine instance override of that file. That results in the following output for me on Chrome, for example:

![image](https://user-images.githubusercontent.com/1042475/30560688-ad1a504a-9c86-11e7-893f-61aca822bab6.png)

When we initially did this work, we noticed lots of discrepancies with % and px for printing between browsers. Some of that is discussed here: https://github.com/CoastalResilienceNetwork/GeositeFramework/pull/763

@lesserj Let me know if you have any suggestions or input on other rules we should try and remove `!important` from.

lesserj:
Pretty much all of the important tags should be removed and be overridable.  This should be pretty straight-forward to do.  You just need to increase the specificity of the print.css so that it is more specific than the screen CSS it is overriding.  This should be doable without important tags.  Here is a good resource to help https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity

[...]

PR #988, opened 2017-09-18
Clean-up print stylesheets 

caseycesari:
## Overview

The following changes were made to improve the implementation of the print-specific stylesheets:

- Remove !important where possible to make rules easier to override by plugin developers
- Remove duplicate rules
- Where !important couldn't be removed, document why
- Remove rules for elements that no longer exist
- Condense rule blocks when multiple blocks existed for the same element.

Connects #984

## Testing Instructions

- Compare print output from `develop` with this branch, and verify that the output generally looks the same. There may be some small differences in padding, border, or size, but nothing that makes the output worse.
- Make sure to try both landscape and portrait layouts, with and without layers on the map, and framework and plugin print.

### DIFF
--- a/src/GeositeFramework/css/app-print.css
+++ b/src/GeositeFramework/css/app-print.css
@@ -5,63 +5,37 @@
         text-align: center;
     }
 
-    .print-sandbox-header {
-        display: flex;
-        flex-direction: row;
-        height: 0.95in;
-    }
-
     #print-map-container {
-        overflow: hidden !important;
-    }
-
-    #map-print-sandbox {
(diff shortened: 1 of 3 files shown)


caseycesari:
Also tagging @lesserj for input.

caseycesari:
For my own sanity, I went through this again with each browser. Here are the results:

Chrome
image

Firefox
image

Safari
image

IE 
image

caseycesari:
For reference, while I was making these changes, I was testing both framework and plugin (in-app) print.

Issue #938, opened 2017-03-30
Make legend one scrollable column


ajrobbins:
Currently, the legend dynamically expands to fit the content, and can be multiple columns side by side. TNC users have found this distracting and it covers their map space. TNC is requesting that the legend be one fixed width (one column) and be scrollable.

PR #990, opened 2017-09-20
Reduce map legend to one column

caseycesari:
## Overview

When the legend is auto-resizing, it will now only use one column instead of two. This is done by setting the min and max width to the same value.

Per request from TNC.

Connects to #938

### Demo

tnc3

### Notes

I decided to leave the existing implementation and just adjust the relevant constant because it made for a quick fix and there's a chance this decision may be revisited.

## Testing Instructions

- Activate the regional planning plugin.
- Activate layers until the legend needs to resize. As it resizes, verify that it stays one column wide.
- ~~Manually resize the legend to two columns. Verify that the legend keeps this shape even as layers are turned off.~~

### DIFF
--- a/src/GeositeFramework/js/Legend.js
+++ b/src/GeositeFramework/js/Legend.js
@@ -233,7 +233,7 @@ define(['use!Geosite',
             // legends. It will do so until the hard-coded limits
             // here are reached.
             var MAX_HEIGHT = 400, // Somewhat arbitrary
-                MAX_WIDTH = 485, // Just enough to fit two columns
+                MAX_WIDTH = 255, // Just enough to fit one column
                 MIN_WIDTH = 255, // Just enough to fit one column
                 LEGEND_BODY_PADDING = 22;
 


kellyi:
> Activate layers until the legend needs to resize. As it resizes, verify that it stays one column wide.

Confirmed that the column stays one column wide as layers are added.

> Manually resize the legend to two columns. Verify that the legend keeps this shape even as layers are turned off.

Looks like it's resizing to one column when then layers change if it's initially two:

tnc-legend

kellyi:
Just realized I may not have hard-refreshed the browser, so going to check again.

kellyi:
> Looks like it's resizing to one column when then layers change if it's initially two:

Double checked and it seems like it does sometimes resize the two-column legend to one on toggling layers.

Not sure what the condition is but it seems like layers which have only one symbology always toggle it from two columns back to one and that layers which have multiple symbologies (like "Population Density" in the first gif above) sometimes keep it at two.

caseycesari:
Thanks. Let me take a look. I may be misremembering how the original logic worked.

caseycesari:
Okay, I was wrong about the legend behavior. I assumed that once you resized it, it stopped adjusting, but that is not the case. I was able to reproduce a similar issue on `develop`. While you can resize the legend, it always tries to do a best fit of the legend items.

kellyi:
> While you can resize the legend, it always tries to do a best fit of the legend items.

Got it. So essentially: if the legend is at two columns but we toggle another layer on or off, the legend will resize conditionally based on whether there's space for it -- and we're good with the current default behavior being that it will occasionally try to squish it back down to one column?

caseycesari:
> Got it. So essentially: if the legend is at two columns but we toggle another layer on or off, the legend will resize conditionally based on whether there's space for it

That's right.

> we're good with the current default behavior being that it will occasionally try to squish it back down to one column?

I believe so, since the issue states that users did not prefer the two columns. If not, we can open another issue.

caseycesari:
Thanks!

matts5517:
Hey Casey,
There is one more issue that I have noticed with the legend, see bottom of image below. the legend body seems to overflow out of the legend container.

The fix for me was to set the .legend-body class to a height of 99% vs. 100%.

Thanks,
Matt

Issue:
legendoverlap

fix:
screen shot 2017-09-21 at 10 07 42 am

caseycesari:
Thanks for the heads up @matts5517. I opened a new issue so that this doesn't get lost: https://github.com/CoastalResilienceNetwork/GeositeFramework/issues/993.

Issue #972, opened 2017-08-22
Disable subregions when super web app mode is active


caseycesari:
- Make sure this feature is disabled and the subregion boundaries are removed from the map.

PR #991, opened 2017-09-20
Disable subregions in single plugin mode

caseycesari:
## Overview

The subregion "availablePlugins" configuration option conflicts with single plugin mode, as well as adds complications to supporting single plugin mode.

Connects to #972

## Testing Instructions

- Activate the app in regular mode. Verify the subregion geometries are on the map.
- Activate the app in single plugin mode. Verify the subregion geometries are removed from the map.


caseycesari:
Thanks!

PR #992, opened 2017-09-21
Ensure help pane is displayed on top of map scale

caseycesari:
Connects #974

## Demo

*Before*
image

*After*
image

## Testing Instructions

Toggle the `About` pane and verify that it is displayed on top of the map scale.


caseycesari:
Thanks!

Issue #993, opened 2017-09-21
Legend content extends beyond legend


caseycesari:
From @matts5517:

> There is one more issue that I have noticed with the legend, see bottom of image below. the legend body seems to overflow out of the legend container.

> The fix for me was to set the .legend-body class to a height of 99% vs. 100%.

> ![image](https://user-images.githubusercontent.com/19394069/30706077-8849c8a6-9eb4-11e7-9352-03e9d82d8df1.png)

> ![image](https://user-images.githubusercontent.com/19394069/30706145-bc3b2c4a-9eb4-11e7-9dcd-eb1b8bff5885.png)

PR #994, opened 2017-09-22
Adjust legend content size to reduce overflow

caseycesari:
Previously, once enough items were added to the legend, the last item in the list would very slightly extend beyond the bottom of the legend container. By slightly reducing the size of legend content, the overflow is eliminated. Reported and fixed by @matts5517.

Connects #993

### Demo

**Before**
image
*Note white line under legend*

**After**
image

## Testing Instructions

- Add and remove layers and verify that the legend never overflows like in the above *before* photo.

### DIFF
--- a/src/GeositeFramework/css/main.css
+++ b/src/GeositeFramework/css/main.css
@@ -1670,7 +1670,7 @@ header .nav-main {
 .content .legend .legend-body {
     padding: 10px 15px;
     overflow: auto;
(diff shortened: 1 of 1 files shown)


caseycesari:
@jwalgran Feel free to hold off on this until we meet.

caseycesari:
Thanks @matts5517 @jwalgran!

Issue #982, opened 2017-09-12
Move hamburger menu to the right side of header


caseycesari:
![image](https://user-images.githubusercontent.com/1042475/30178729-ffe7936c-93d7-11e7-9800-bf8c5be98e53.png)

PR #995, opened 2017-09-22
Move menu to right in single plugin mode

caseycesari:
## Overview

Per request from TNC. I believe this was requested so that the app title is the only thing on the left side of the header.

Connects #982 

### Demo

image

### Testing

- Load the app in single plugin mode in every browser, and verify that the menu is on the left and visible.

### DIFF
--- a/src/GeositeFramework/Views/Home/Index.cshtml
+++ b/src/GeositeFramework/Views/Home/Index.cshtml
@@ -40,7 +40,7 @@
                         <i class="dropdown-icon icon-angle-right"></i>
                     </a>
 
-                    <ul class="dropdown-menu">
+                    <ul class="dropdown-menu" id="region-links">
                         @RenderRegionLinks(link.Items)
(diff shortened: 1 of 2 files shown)


kellyi:
This is working well! I wonder if we may need to rethink how this is designed if it appears on the right? Here's what it looks like in non-single-app-mode:

screen shot 2017-09-26 at 2 07 36 pm

Wonder if there's a straightforward way to capture the same functionality while moving the whole thing to the left?

ajrobbins:
@jfrankl ^

jfrankl:
Can the list of locations move to the left of the initial dropdown so they don't overlap?

caseycesari:
Yeah, I'll give that a shot. Thanks.

caseycesari:
Pushed a fixup. This is what the list looks like now:

image

caseycesari:
Thanks!

Issue #987, opened 2017-09-18
Layout-specific print stylesheets are not removed


caseycesari:
These should be removed after the document is printed, or before the next print run is started.

![image](https://user-images.githubusercontent.com/1042475/30557679-1a42bdf6-9c7d-11e7-94f8-f61b9ed34aad.png)

PR #997, opened 2017-09-25
Fix removal of layout-specific print stylesheets

caseycesari:
## Overview

Class name should not have a leading ".". This is only required when
querying the DOM. The leading "." prevented the CSS import from being
found during print clean-up.

Connects to #987

## Testing Instructions

- Run through the print/"create map" workflow a few times and specify each page layout at least once.
- After this, ensure that any imports of `print-portrait.css` and `print-landscape.css` have been removed.


caseycesari:
Thanks!

Issue #1003, opened 2017-09-27
Implement CSS and DOM elements for mobile single plugin mode


caseycesari:
Also, wire these up to the appropriate calls. Depends on #1002 

![image](https://user-images.githubusercontent.com/1042475/30938558-6bf73f00-a3a8-11e7-82d7-8a12a7b4e901.png)

![image](https://user-images.githubusercontent.com/1042475/30938569-734f70ba-a3a8-11e7-833e-197ba19e7ecf.png)

Issue #1002, opened 2017-09-27
Implement map/sidebar toggle functionality and API


caseycesari:
This will be utilized on mobile viewports to toggle whether the map or the sidebar is shown. In reality, it may just be toggling the visibility of the sidebar.

There should be a way to do this programmatically (so that a plugin can do it) and through the UI.

The code may be toggling classes on the map and sidebar that effect the visibility of each item. Those classes should only have an effect on small viewports.

Also, this toggling is designed for single plugin mode, but it may make sense to make it available in any configuration.

ajrobbins:
Designs: [2017_09_25_phone_v1.pdf](https://github.com/CoastalResilienceNetwork/GeositeFramework/files/1340931/2017_09_25_phone_v1.pdf)

caseycesari:
Some issues that came up when I started implementing this:

- Assuming we'll use a media query to show the buttons and hide the map/sidebar initially, we'll have to make sure the programmatic route can override those rules.
- The framework needs some CSS rule revisions to make it more mobile-friendly in general.
- The search element can become collapsed via a configuration option, but how will we know to do this on small screens?
- Regarding the above point, and more in general, how will the JS implementation of this interact with the CSS? Querying for viewport sizes seems fragile. We could add a class to the body when we know the media queries will be in effect, but that will require keeping parts of the code and CSS in sync.

kellyi:
Got a working-ish version of the toggle buttons so far on this branch https://github.com/CoastalResilienceNetwork/GeositeFramework/tree/ki/mobile-super-web-app

![map-content-toggle-buttons](https://user-images.githubusercontent.com/4165523/31360215-5f86f2c0-ad1b-11e7-9def-89752d103225.gif)

Going to give this some thought in advance of discussing it tomorrow:

> The framework needs some CSS rule revisions to make it more mobile-friendly in general.

kellyi:
Dropping this here for possible use:

https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia

kellyi:
@caseycesari 

Do we want this:

> There should be a way to do this programmatically (so that a plugin can do it) and through the UI.

...to be available both in non-mobile and mobile mode? I assume it should only be available in single plugin mode, but should it always be available otherwise?

caseycesari:
Should be just mobile single-plugin mode for now.

PR #1004, opened 2017-10-10
Add mobile style rules & plugin API

kellyi:
## Overview

This PR adds some style rules & interfaces to enable the app to show a mobile version on devices sized up to iPhone 6/7/8s-sized widths. To accomplish this, I added a footer component with buttons which only display when the media query for a phone-sized device matches, along with some logic to handle toggling in the app.

The app also now adds two methods for plugins in single app mode: `toggleMobileMap` and `toggleMobileContent`. These toggle each of those on, respectively, and change the state such that the buttons & toggle icon are adjusted in tandem. I added some error messages & noops to handle cases in which the methods don't exist for whatever reason, but left them exposed & callable in non-mobile versions of the app. (We can adjust the names to something which doesn't include "mobile", btw, if we decide to keep them accessible for all screen sizes).

Connects #1002
Connects #1003 

## Demo

I tested this out with a few iPhone simulators and on a 6s; we also took a look at an Android version on @rajadain's phone

Here's iPhone SE

iphonese

And 7S:

iphone7s

## Notes

- Still working through some CSS stuff wrt to the search bar: I think the ideal behavior is that when it's expanded, it occupies the full breadth of the header from left margin to the search icon -- and also covers up the app header title and "About" button.

- I think that some of the app's external CSS rules warrant some changes to make them rely more on viewport size definitions and less on percentages. We can spin that off to another card.

## Testing
- get and rebuild this branch
- visit the app in the browser in single plugin mode in a regular browser-sized viewport and verify that it works as before
- change the viewport to a mobile sized browser then hard refresh and verify that you see mobile mode and that everything is functioning
- check a few mobile viewport sizes to verify that the app appears correctly in each of them

### DIFF
--- a/src/GeositeFramework/css/main.css
+++ b/src/GeositeFramework/css/main.css
@@ -2122,37 +2122,130 @@ body.x-body {
 .arcgisSearch .searchGroup .searchInput:-moz-placeholder { /* Firefox 18- */
   color: #fff;
 }
-#single-plugin-toggle-footer {
-    bottom: 0px;
-    width: 100vw;
-    height: 8vh;
-    background-color: lightgray;
-    z-index: 100;
-    text-align: center;
+
+@media not screen and (max-device-width: 737px) {
+    #single-plugin-toggle-footer {
+        display: none !important;
+    }
 }
 
-#single-plugin-toggle-buttons {
-    display: inline-block;
-    width: 100vw;
-    height: 8vh;
-    margin-top: 1vh;
-}
+@media screen and (max-device-width: 736px) {
+    #toggle-plugin-container {
+        display: none !important;
+    }
 
-.single-plugin-toggle-button {
-    width: 48vw;
-    height: 75%;
-    font-size: 20px !important;
-    font-weight: 500;
-    font-family: colfax, -apple-system, BlinkMacSystemFont, Roboto, Oxygen, Ubuntu, Cantarell, Fira Sans, Droid Sans, Helvetica Neue, Arial, sans-serif;
-    padding: 0 !important;
-    border: 0px !important;
-    border-radius: 0 !important;
-}
+    #single-plugin-toggle-footer {
+        bottom: 0px;
+        width: 100vw;
+        background-color: lightgray;
+        z-index: 100;
+        text-align: center;
+        display: block;
+    }
 
-#single-plugin-toggle-content-button {
+    #single-plugin-toggle-buttons {
+        display: inline-block;
+        width: 100vw;
+        height: inherit;
+        margin-top: 1vh;
+    }
 
-}
+    .single-plugin-toggle-button {
+        width: 48vw;
+        height: 75%;
+        font-size: 20px !important;
+        font-weight: 500;
+        font-family: colfax, -apple-system, BlinkMacSystemFont, Roboto, Oxygen, Ubuntu, Cantarell, Fira Sans, Droid Sans, Helvetica Neue, Arial, sans-serif;
+        padding: 0 !important;
+        border: 0px !important;
+        border-radius: 0 !important;
+    }
 
-#single-plugin-toggle-map-button {
+    #single-plugin-toggle-buttons .button-primary {
+        background-color: #0096d6 !important;
+        border-color: #0084bd !important;
+    }
 
+    #single-plugin-toggle-buttons .button-primary:focus {
+        box-shadow: unset !important;
+    }
+
+    .esriMapContainer {
+        width: 100vw !important;
+    }
+
+    @media (orientation: portrait) {
+        #single-plugin-toggle-footer {
+            height: 8vh;
+        }
+    }
+
+    @media (orientation: landscape) {
+        #single-plugin-toggle-footer {
+            height: 10vh;
+        }
+    }
+
+    .sidebar {
+        width: 100vw !important;
+        -webkit-overflow-scrolling: touch;
+    }
+
+    .control-container {
+        width: 100vw;
+    }
+
+    header {
+        width: 100vw;
+    }
+
+    .app-fullscreen {
+        width: 100vw;
+        position: fixed;
+    }
+
+    #single-plugin-mode-help-container {
+        z-index: 5010;
+    }
+
+    #search {
+        position: fixed !important;
+        right: 10vw !important;
+    }
+
+    .searchBtn {
+        background-color: inherit !important;
+    }
+
+    #search .searchExpanded .searchExpandContainer {
+        background-color: #5C95B5;
+        width: 75vw;
+        margin-left: 2vw !important;
+    }
+
+    .arcgisSearch .searchExpanded #search_input {
+        width: 60vw;
+        margin-left: 2vw !important;
+    }
+
+    .nav-main-dropdown {
+        position: fixed !important;
+        right: 2vw !important;
+    }
+
+    .nav-region-subtitle {
+        right: 22vw;
+    }
+
+    .nav-main {
+        padding-right: 5px;
+    }
+
+    .nav-main-title {
+        display: inline-block;
+        white-space: nowrap;
+        width: 60vw;
+        text-overflow: ellipsis;
+        overflow: hidden;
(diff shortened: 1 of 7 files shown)


kellyi:
About to update the demo gif btw.

src/GeositeFramework/js/SidebarToggle.js
```diff
@@ -47,20 +47,84 @@
...
+            this.mobileMapButton = $('#single-plugin-toggle-map-button');
+
+            this.listenTo(this.viewModel, 'change', this.adjustButtonCSS);
+            this.showMobileMap = this.showMobileMap.bind(this);
+            this.showMobileContent = this.showMobileContent.bind(this);
+
+            this.mobileMapButton.on('click', this.showMobileMap);
+            this.mobileContentButton.on('click', this.showMobileContent);
+
+            N.app.toggleMobileMap = this.showMobileMap;
```
caseycesari: I think I prefer the internal (`show...`) name of these functions over the public (`toggle...`) name of these functions, since they will always show and not toggle the appropriate content. Is that right?

src/GeositeFramework/js/SidebarToggle.js
```diff
@@ -47,20 +47,84 @@
...
+
+                // Unhighlight mobile content button
+                this.mobileContentButton.removeClass('button-primary');
+                this.mobileContentButton.addClass('button-secondary');
+            }
+        },
+
+        showMobileContent: function() {
+            var singlePlugin = this.viewModel.get('plugins').at(0);
+            singlePlugin.select();
```
caseycesari: Can this be omitted? `select` may have side-effects, like reseting controls and options made by the user to the plugin.

src/GeositeFramework/js/SidebarToggle.js
```diff
@@ -47,20 +47,84 @@
...
+            this.viewModel = options.viewModel;
+            this.mobileContentButton = $('#single-plugin-toggle-content-button');
+            this.mobileMapButton = $('#single-plugin-toggle-map-button');
+
+            this.listenTo(this.viewModel, 'change', this.adjustButtonCSS);
+            this.showMobileMap = this.showMobileMap.bind(this);
+            this.showMobileContent = this.showMobileContent.bind(this);
+
+            this.mobileMapButton.on('click', this.showMobileMap);
+            this.mobileContentButton.on('click', this.showMobileContent);
```
caseycesari: If possible, consider assigning these events using the Backbone [event delegation](http://backbonejs.org/#View-events) style.

src/GeositeFramework/js/SidebarToggle.js
```diff
@@ -47,20 +47,84 @@
...
-                this.$el.find('i').attr('class', 'fa fa-chevron-left');
             }
         }
     });
+
+    // Mobile toggle view for single plugin mode
+    N.views.MobileTogglePlugin = Backbone.View.extend({
+        initialize: function(options) {
+            this.viewModel = options.viewModel;
+            this.mobileContentButton = $('#single-plugin-toggle-content-button');
```
caseycesari: This won't have a big effect, but consider searching the view's `$el` for these elements.

src/GeositeFramework/js/Plugin.js
```diff
@@ -97,6 +99,28 @@ require(['use!Geosite',
             }
         }
 
+        function toggleMobileMap() {
+            if (!this.singlePluginMode) {
+                console.error('Programmatic map toggle only available in single plugin mode');
```
caseycesari: This is just my preference, but I don't feel that we need to add a message to the console is these cases. What do you think @mmcfarland?

src/GeositeFramework/css/main.css
```diff
@@ -2120,3 +2120,95 @@ body.x-body {
...
   color: #fff;
 }
+@media screen and (min-device-width: 737px) {
+    #single-plugin-toggle-footer {
+        display: none;
+    }
+}
+
+@media screen and (max-device-width: 736px) {
+    #toggle-plugin-container {
```
caseycesari: Are the `!important` flags used in this block required? I thought that the rules defined within the media query would take precedent over other rules when the media query is active.

caseycesari:
The about pane should be displayed above the toggle buttons since they will have no effect while the about pane is open:

image

kellyi:
> I tested this on my phone in mobile Safari, and I'm not able to scroll to the toggle buttons.

Ah, this is due to changing the app height from 100% to 100vh -- forgot that mobile Safari counts the bit beneath the navigation bar as part of the viewport.

src/GeositeFramework/js/SidebarToggle.js
kellyi: Yeah -- I'll change these throughout to `show...`.

src/GeositeFramework/js/Plugin.js
kellyi: Going to drop these error messages; without them, these functions are small enough that we can just make the assignments inline in the initializer:

```js
showMobileMap = (N.app.singlePluginMode && N.app.showMobileMap) ?
    N.app.showMobileMap : _.noop(),
showMobileContent = (N.app.singlePluginMode && N.app.showMobileContent) ?
    N.app.showMobileContent : _.noop();
```

src/GeositeFramework/js/SidebarToggle.js
kellyi: Let me check on this. My initial thought was that we'd have the buttons behave the same way as the sidebar collapse tab chevron button works in the desktop single web app: it calls `select` and `deselect` when the plugin's expanded and collapsed, respectively.

The idea here was that I wanted to keep them in harmony so now both the `TogglePlugin` view and the `MobileTogglePlugin` view listen for & change on the same model events. 

However, I think you're suggesting that we shouldn't ever deselect the plugin in mobile view -- and instead just show and hide it?

src/GeositeFramework/css/main.css
kellyi: Some of these do seem to be required, but I'll go through and check to make sure we keep only the minimal number of them.

kellyi: I pared these down to what seem to be the minimal required set.

src/GeositeFramework/js/SidebarToggle.js
caseycesari: That makes sense, but I think we don't actually want it to be in harmony with that button because we want to the plugin to maintain all of it's state in this case. Changes in `select/deselect` don't necessarily reset or lose the plugin state, but I think it can if a plugin chooses to do so. 

Does that sound right @mmcfarland?

kellyi: I updated these to use `show`/`hide` instead of select & deselect.

kellyi:
ac84e90 addresses the code comments, makes some CSS fixes, and also adjusts the geocoder search bar such that we now hide the app title & subtitle when the search bar's expanded & show it when it's collapsed.

I updated the demo gifs in the first comment above, too.

One additional testing instruction is that in during single app mobile mode, you should be able to call the following functions in the console & see the desired effects:

```
Geosite.app.showMobileMap();
Geosite.app.showMobileContent();
```

`showMobileMap` should show the map & toggle the buttons such that the map button's highlighted; `showMobileContent` should show the content pane & toggle the buttons such that the content button's highlighted.

src/GeositeFramework/js/SidebarToggle.js
kellyi: Oof, just realized I updated the wrong set. Pushing another fixup.

caseycesari:
Oh, also the button focus color issue!

kellyi:
814bc7a handles both hiding the mobile button footer for non-mobile devices & also dropping the green-ish focus class from the toggle buttons

kellyi:
Great! I just squashed this, then rebased it on develop & tested it out again. Everything's still working as expected, so I'm going to merge this in.

Thanks for all your help with this!

Issue #1009, opened 2017-10-12
Replace print preview in plugin print workflow with sandbox


caseycesari:
Rather than providing a print preview in the plugin print workflow, provide a sandbox contained within the print modal that developers have control over. They could place a print preview, or an input form, or whatever they want.

After the user clicks print, the contents of sandbox should be provided to the plugin developer before `window.print` is called.

Issue #985, opened 2017-09-18
Prompt for page size when printing


caseycesari:
This will allow us to resize the print area before browser print is initiated. Initially, we should support letter (default), legal, and A4, and resize the print area accordingly.

The user may need to reselect the correct page size in the browser print prompt.

![image](https://user-images.githubusercontent.com/1042475/30547905-a18d5498-9c5e-11e7-8c61-eb1c69f13b61.png)

caseycesari:
Make sure to support this in the plugin print workflow as well.

PR #1011, opened 2017-10-12
Add support for A4 page size for printing

caseycesari:
## Overview

Adds stylesheets for A4 page size (both landscape and portrait orientations) as well as an option to the framework print dialog that allows users to select their page size. The A4 stylesheets were based of the letter size stylesheets, but with difference in dimensions between A4 and letter added/subtracted.

The existing page orientation stylesheets were renamed to reflect that they are for letter size.

Plugins can utilize the page size and orientation stylesheets either by adding them to the page directly when printing, or getting user feedback in the plugin print dialog (to be added in #1009) and adding the appropriate stylesheets.

Connects to #985

## Testing Instructions
 
- Initiate the create map option and verify that the correct stylesheet is added for each set of size and orientation options. These stylesheets are added after the print button is selected to the bottom of the `head` element. You can see them being added by viewing the bottom of the `head` element in the developer console.
- Verify the map and legend elements are sized appropriately for each selected page and orientation option. This is a little tricky to see because the differences are subtle.
    - The best way I found to test this was to open the app in two tabs in Chrome and setup the maps in each tab identically (same extent, same layers added etc).
    - Initiate the create map option, and select the same orientation in each tab, but a different page size.
    - Click print.
    - In the Chrome print preview dialog, select the **same** page size for each map (it's a little easier to see this with A4 selected in portrait orientation).
    - One of the maps will now have the incorrect page size selected (compared to the initial option selected within the app).
    - Now note how the map elements are sized more appropriately for the map that has the correct page size selected (for example, in A4 portrait mode, the map will be a little larger for the tab with the A4 page size selected).

### DIFF
--- a/src/GeositeFramework/Views/Home/Index.cshtml
+++ b/src/GeositeFramework/Views/Home/Index.cshtml
@@ -478,14 +478,28 @@
                     <div class="popover-section">
                         <h3 class="i18n" data-i18n="Orientation">Orientation</h3>
                         <label class="form-component">
-                            <input type="radio" name="export-orientation" value="Landscape" checked="checked"/>
+                            <input type="radio" name="export-orientation" value="landscape" checked="checked"/>
                             <div class="check"></div>
-                            <span class="form-text i18n" data-i18n="Landscape">Landscape</span>
+                            <span class="form-text i18n" data-i18n="landscape">Landscape</span>
                         </label>
                         <label class="form-component">
-                            <input type="radio" name="export-orientation" value="Portrait"/>
+                            <input type="radio" name="export-orientation" value="portrait"/>
                             <div class="check"></div>
                             <span class="form-text i18n" data-i18n="Portrait">Portrait</span>
+                        </label>
+                        <p class="font-italic small i18n" data-i18n="Browser Print">* You may need to alter your browser print settings</p>
+                    </div>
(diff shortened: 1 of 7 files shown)


kellyi:
Tested this out and they're sized differently; I printed the same map to pdfs as both letter and a4 and a4 is slightly narrower:

screen shot 2017-10-12 at 3 00 22 pm

Similar difference happens in portrait mode.

The legend shows up appropriately. Not sure if this is a known limitation or bug but it seems like the actual layers aren't showing up on the printed map. Here's what they look like on screen:

screen shot 2017-10-12 at 3 04 03 pm

and here's the printed map:

screen shot 2017-10-12 at 3 04 32 pm

caseycesari:
Interesting. Good catch. This is built on top of work that should preserve the layers on the map. I'll take a look. I wonder if it's something with that layer in particular.

caseycesari:
Okay, so I had been testing this through the plugin print (which is not what I said to do in the instructions), which shows all of the layers that are on the map. Turns out I somehow broke the layers in the framework print in my previous PR. I'll fix it there and let you know. Thanks for finding this.

caseycesari:
Turns out I broke this in #988. Just pushed a commit to fix it. Mind taking another look?

caseycesari:
Thanks!

Issue #1006, opened 2017-10-12
Setup test region and send instructions for testing new features


caseycesari:
Setup a new region to test out super web app mode and the new print workflow.

After that is setup, send TNC the following:
- Link the region repo that has the region.json config for the new features
- Link to the plugin that demonstrates the new config
- The build system commands that can be used to install to their dev environment

We have done this before with the ESRI JS API 4.x branch: https://github.com/CoastalResilienceNetwork/GeositeFramework/wiki/Updating-a-Region-for-the-ArcGIS-JavaScript-API-4.x-Pilot#building-a-pilot-region

PR #1016, opened 2017-10-23
Clean-up identify point plugin and add print button to UI when in single plugin mode

caseycesari:
## Overview

Cleans-up the internals and UI of the identify point plugin, which will be used to demo single plugin mode and the revamped plugin print workflow. The base plugin code was also restructured so that plugins can trigger the plugin print workflow from single plugin mode.

Connects #1006 

### Notes

The diff of c009fd6 is a bit hard to read, but not much code was added. Code was mostly moved around. It may be easier to view the diff in split mode.

## Testing Instructions

- Start the framework in Single Plugin mode.
- Click the print button in the identify point plugin. Verify that the plugin print workflow is kicked off.
- Verify that the demo form controls on the second tab of the plugin still work.
- Restart the framework in normal mode.
- Open the identify point plugin. Verify that the print button is gone from the plugin body.

### DIFF
--- a/src/GeositeFramework/js/Plugin.js
+++ b/src/GeositeFramework/js/Plugin.js
@@ -430,6 +430,8 @@ require(['use!Geosite',
             view.$el.addClass(model.getId() + '-' + view.paneNumber);
 
             if (view.$uiContainer) {
+                assignEvents(view.$uiContainer, model, view.paneNumber);
+
                 if (view.model.selected === true) {
                     view.$uiContainer.show();
                 } else {
@@ -479,6 +481,14 @@ require(['use!Geosite',
 
             view.$uiContainer = $uiContainer;
 
+            // Attach to top pane element
+            view.$el.parents().find('.content .nav-apps').after($uiContainer);
+
+            // Tell the model about $uiContainer so it can pass it to the plugin object
+            model.set('$uiContainer', $uiContainer);
+        }
+
+        function assignEvents($uiContainer, model, paneNumber) {
             $uiContainer
                 // Minimize the plugin
                 .find('.plugin-minimize').on('click', function() {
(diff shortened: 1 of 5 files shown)


kellyi:
Checked this out locally & it's working as described: print button's visible and working in single plugin mode, but hidden in non-single-plugin-mode. The demo elements also still work in each. About to take a look at the code.

src/GeositeFramework/sample_plugins/identify_point/main.js
```diff
@@ -197,6 +137,18 @@ define(["dojo/_base/declare", "framework/PluginBase"],
                     html: '<img src="' + imageSourcePath + '"/>',
                     fixed: true,
                 });
+            },
+
+            render: function(frameworkParameters) {
+                $(this.container).append(template);
```
kellyi: Ah, we should have done this from the start.

src/GeositeFramework/js/Plugin.js
```diff
@@ -496,58 +506,59 @@ require(['use!Geosite',
...
+            // Add the plugin css
+            addCss(pluginCssPath, printCssClass);
+
+            // Hide the possible same plugin from the opposite map pane
+            addCss(oppositePaneHideCssPath, printCssClass);
+
+            // Give some time for the browser to parse the new CSS,
+            // if this wasn't slightly delayed, occasionally the override would
+            // not be present and print features wouldn't show up.
+            _.delay(parseDeferred.resolve, 200);
```
kellyi: Would this work with a delay value of `0`? Wondering if the `200` value is significant here or if it's just that this needs to be pushed to the back of the event loop queue.

caseycesari: This line is showing up new because of how things were moved around, but I don't know for sure. I imagine that using `0` didn't work in every case.

src/GeositeFramework/sample_plugins/identify_point/main.js
caseycesari: Yeah, it makes it a little easier to manage the DOM.

caseycesari:
Thanks for the review!

Issue #939, opened 2017-03-31
Enable using TINY modal in Launchpad/Getting Started


kellyi:
Enable TNC developers to use the TINY modal in the launchpad/getting started plugin. This plugin is configured by `region.json` file according to steps doucmented here:

https://github.com/CoastalResilienceNetwork/GeositeFramework/wiki/Launchpad-(Get-Started)

We'll need to update the plugin to accept configuring a TINY modal via the `launchpad` key in `region.json`, update the wiki page to document the new config option, and ensure that the changes to `region.json` don't break anything else in the framework.

PR #1022, opened 2017-11-08
Make launchpad asset paths relative

caseycesari:
This is needed so that when the app is served under a virtual folder/directory (like maps.coastalresilience.org/site-name), the assets are loaded correctly.

Connects to #939


caseycesari:
Thanks!

Issue #1001, opened 2017-09-27
Include custom plugin legends in printed legend


caseycesari:
They are currently ignored.

Some background here: https://github.com/CoastalResilienceNetwork/GeositeFramework/issues/963

PR #1027, opened 2017-11-21
Include plugin legends in printed map legend

caseycesari:
## Overview

Previously, plugin legends were not displayed in the printed map legend because they were not accounted for in the print stylesheets. This resulted in the legends being rendered off the page. Now, the plugin legends container is temporarily moved into the layer legends container. That, combined with modifications to the print stylesheets, makes the plugin legends visible on the printed page.

In the course of this work, the general legend print styling was also improved and a superfluous wrapping of legend elements in a special print class was removed.

It should be noted that while this does allow for the plugin legends to be displayed in the printed legends, most plugin legends are SVG and will not display well. They will likely be taller than the legend area, which will result in the legends being cut-off. However, the scope of this PR was just to get them to display. Plugin authors will be responsible for getting the legends to display well.

Connects to #1001

## Notes

The print workflow no longer works in Safari because of this dialog that appears when `windows.print()` is called. The timing of the workflow is now off.

image

## Demo

In the screenshots below, the "Likelihood of flooding" legend comes from a plugin. The other layers are from Regional Planning.

**Previously, in Chrome**
image

**Chrome**
image

**Firefox**
image

**Internet Explorer**
image

## Testing instructions

- Add a plugin to the framework that includes a legend. I used "Ecosystem effects of sea level change" in the above screenshots, so it may be worth trying a different plugin to get more test cases.
- Add layers from Regional Planning.
- Select the "Create Map" option.
- Verify the printed map legend contains both the Regional Planning layers and the plugin legend. It's easy to do this in Chrome because of the print preview. In Firefox and Internet Explorer, printing to PDF is the quickest way.

### DIFF
--- a/src/GeositeFramework/css/app-print.css
+++ b/src/GeositeFramework/css/app-print.css
@@ -93,42 +93,34 @@
     few legend style blocks results in the rules, despite not having any conflicting
     rules, not having any effect. It could have something to do with how the legend
     elements are moved around the DOM for printing */
-    .legend-body {
+    div.legend-body {
+        display: -webkit-box;
+        display: -ms-flexbox;
         display: flex;
-        flex-wrap: wrap;
         width: 100% !important;
         height: 85% !important;
         background-color: #fff !important; /* Needed for IE */
     }
 
-    .layer-legends {
-        height: 100% !important;
-        width: 100% !important;
+    div.layer-legends {
+        display: -webkit-box;
+        display: -ms-flexbox;
         display: flex;
-        flex-direction: column;
-        flex-wrap: wrap;
+        -webkit-box-orient: vertical;
+        -webkit-box-direction: normal;
+            -ms-flex-direction: column;
+                flex-direction: column;
+        -ms-flex-wrap: wrap;
+            flex-wrap: wrap;
+        width: 100%;
+        align-content: flex-start;
     }
 
-    .legend-layer {
+    div.legend-layer {
         width: auto !important;
-        font-size: 8px;
         height: auto !important;
-        display: flex;
-        flex-direction: column;
-        flex-shrink: 1;
-    }
-
-    .legend-title {
-        max-width: 75%;
-    }
-
-    .print-legend-coll {
-        display: flex !important;
-        flex-direction: column;
-        flex-wrap: wrap;
-        height: 60px;
-        align-content: stretch;
-        justify-content: center;
+        font-size: 8px;
+        margin: 0 10px 0 0;
     }
 
     .legend-layer img {
--- a/src/GeositeFramework/js/Export.js
+++ b/src/GeositeFramework/js/Export.js
@@ -50,7 +50,10 @@ require(['use!Geosite'],
                 orientDeferred = $.Deferred(),
                 resizeDeferred = $.Deferred(),
                 legendDeferred = $.Deferred(),
-                postPrintAction = _.noop;
+                postPrintAction = function() {
+                    var pluginLegend = $('.plugin-legends').detach();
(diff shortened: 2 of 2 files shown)


kellyi:
Taking a look at this.

kellyi:
Here's Chrome using regional planning and energy planning:

screen shot 2017-11-29 at 2 57 50 pm

kellyi:
Another one in Chrome:

screen shot 2017-11-29 at 3 14 15 pm

->

screen shot 2017-11-29 at 3 13 56 pm

->

screen shot 2017-11-29 at 3 15 47 pm

(Not sure what caused the layers to get moved from where they belong -- it happened on a resize.)

kellyi:
Oh: it looks like the post-print adjustments now move the basemap such that it's no longer in sync with the overlays. This happened in Firefox after printing:

screen shot 2017-11-29 at 3 18 00 pm

kellyi:
Zooming back out and back in fixes the layer misalignment, btw. I wonder if we should invalidate the map size to force a redraw after printing.

caseycesari:
I'll take a look at that. Thanks.

kellyi:
This is mostly working well! I noticed a few issues which we may want to address here or in other cards:

- some legends with numerous items still get cut off in landscape mode -- see the "Loggerhead Sea Turtles" example above
- the layers are misaligned with the basemap post-print. It's fixed on zooming out and back in; we may want to invalidate the map to force a redraw after the legend's repositioned?
- probably for another issue but: on Macbook-sized screens the "Create map" button's popup is hidden behind the legend if there are numerous items in the legend.

Tested this in Chrome and FF; still figuring out how to export a pdf in IE11.

caseycesari:
> some legends with numerous items still get cut off in landscape mode -- see the "Loggerhead Sea Turtles" example above

I noticed that, but I'm not sure what to do about it. Breaking up the legend across columns doesn't seem like a great approach. The user can sometimes select the portrait layout, which makes the legend taller, but that doesn't always ensure legends won't be cut-off. It's a bigger chunk of work, but we may want to switch the page layout to always have a vertically oriented legend (that would run along side the map instead of under it). It's possible that legends would get cut-off, but much more unlikely. I'll create an issue for that.

> the layers are misaligned with the basemap post-print. It's fixed on zooming out and back in; we may want to invalidate the map to force a redraw after the legend's repositioned?

I'm going to open a new issue for this since it's unrelated to this PR. I thought I fixed it in https://github.com/CoastalResilienceNetwork/GeositeFramework/pull/1005 with https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/develop/src/GeositeFramework/js/Plugin.js#L643. I'll have to look into this agin.

> probably for another issue but: on Macbook-sized screens the "Create map" button's popup is hidden behind the legend if there are numerous items in the legend.

Good catch. This should be a simple fix, but I'll open a new issue since it's unrelated to this PR and warrants it's own testing.

caseycesari:
Follow-up issues: https://github.com/CoastalResilienceNetwork/GeositeFramework/issues/1033, https://github.com/CoastalResilienceNetwork/GeositeFramework/issues/1034, https://github.com/CoastalResilienceNetwork/GeositeFramework/issues/1035.

kellyi:
New issues address most of my comments above; taking another look at this.

kellyi:
Checked this again and it's working well!

screen shot 2017-11-30 at 11 14 06 am

screen shot 2017-11-30 at 11 13 40 am

screen shot 2017-11-30 at 11 06 05 am

screen shot 2017-11-30 at 11 02 52 am

caseycesari:
Thanks for the thorough review!

Issue #1025, opened 2017-11-13
TINY modal for apps and Get Started


zferdana:
Some times the app icon obscures that title of the app in the title bar; resolved by refresh application. 
![screen shot 2017-11-13 at 2 47 49 pm](https://user-images.githubusercontent.com/5605392/32753476-d7a9155e-c881-11e7-94cc-80c4b9622019.png)

caseycesari:
This is happening because a number of plugins adjust the position of the plugin title, but do so in a way that it effects every plugin. These plugins almost all contain the same snippets of code that are called on activation/deactivation, and they will need to be modified to fix this.

On activation, [this line](https://github.com/CoastalResilienceNetwork/recreation_tourism/blob/1d7bc696ac9b7bbdbd2ea0eb71103abf924ca2bf/main.js#L90) is called, which moves all plugin titles over to the right 25 pixels. On deactivation, [this line](https://github.com/CoastalResilienceNetwork/northcarolina-region/blob/9224ea932f8159aa5dae190eeb129e3d57f8b51f/plugins/community-rating-system/main.js#L53) is called, which moves all plugin titles over to the left 25 pixels.

You only seeing this happen sometimes because if you only activate and deactivate one plugin, the code works as expected. However, after activating another plugin that contains this code, you will soon start to see things get out of sync.

Here's a list of the plugins that do this:

- [Community rating system](https://github.com/CoastalResilienceNetwork/community-rating-system)
- [Ecological effects SLC](https://github.com/CoastalResilienceNetwork/ecological-effects-slc)
- [Migratory blueways](https://github.com/CoastalResilienceNetwork/migratory-blueways)
- [Tuna fisheries](https://github.com/CoastalResilienceNetwork/tuna-fisheries)
- [Benefits explorer](https://github.com/CoastalResilienceNetwork/benefits_explorer)
- [Water security](https://github.com/CoastalResilienceNetwork/water_security)
- [Water funds](https://github.com/CoastalResilienceNetwork/water_funds)
- [Water share](https://github.com/CoastalResilienceNetwork/water_share)
- [Coral reef fisheries](https://github.com/CoastalResilienceNetwork/coral-reef-fisheries)

[...]

PR #1020, opened 2017-11-07
 Implement infographic support in the launchpad

caseycesari:
## Overview

Add support for the infographic feature that was developed by TNC plugin authors.

Site admins can now enable this feature in the region config, provide a HTML template, and a button will be added to the launchpad plugin title. When users click this button, a modal infographic is displayed.

This implementation is largely based on the feature implemented in https://github.com/CoastalResilienceNetwork/ecological-effects-slc.

Connects to #939

### Demo

tnc2

## Testing Instructions

- When the launchpad opens, click the infographic button, and verify the infographic modal opens.
- Disable the infographic in the config, and verify the app loads as normal, and the infographic button is not displayed.

PR #1028, opened 2017-11-22
Implement infographics for plugins

caseycesari:
## Overview

Adds a configuration option to plugins for infographics. When enabled, a button is added to the plugin title bar, which, when clicked, will open the provided infographic content in a modal.

Note that this approach is similar to what was implemented for the launchpad in #1020. However, because of differences in configuration (the launchpad is configured via the region.json file, while the plugins are configured directly in the code), the implementations could not be merged.

Connects to #1025

### Demo

tnc1

## Testing Instructions

- Open the Identify Point plugin.
- Click the infographic button in the title bar.
- Verify the infographic opens in a modal.


src/GeositeFramework/js/Plugin.js
```diff
@@ -461,6 +468,24 @@ require(['use!Geosite',
...
+            });
+        }
+
+        function showInfographic(infographicConfig, sourceFolder) {
+            TINY.box.show({
+                animate: true,
+                url: sourceFolder + '/html/infographic.html',
+                fixed: true,
+                width: infographicConfig[0],
+                height: infographicConfig[1]
```
kellyi: Should we add guards or defaults or throws in case `infographicConfig` is not null but also not a list with at least two elements?

caseycesari: Added!

caseycesari:
Thanks for the review.

Issue #1014, opened 2017-10-23
Add configuration option to toggle app minimization visibility


caseycesari:
By default, and without any configuration, the minimization button should be shown (this is the current behavior). Otherwise, plugins should not show the minimization button.

PR #1029, opened 2017-11-27
Add option to hide plugin minimize button

caseycesari:
## Overview

Adds an option to the plugin configuration that allows site admins to
hide the minimize button in a plugin's titlebar.

Support for this option has to be implemented in each plugin, and is
demonstrated here in the launchpad plugin.

Connects #1014

### Demo

image

## Testing Instructions

- Edit the Launchpad's `overrides.json` and add `hideMinimizeButton: true`.
- Reload the site (this may also require restarting the development server) and verify the Launchpad's minimize button is removed.

### DIFF
--- a/src/GeositeFramework/js/Plugin.js
+++ b/src/GeositeFramework/js/Plugin.js
@@ -475,7 +475,8 @@ require(['use!Geosite',
                     hasHelp: pluginObject.hasHelp,
                     hasCustomPrint: pluginObject.hasCustomPrint,
                     sizeClassName: getPluginSizeClassName(pluginObject.size),
-                    customWidth: getCustomPluginWidth(pluginObject.size, pluginObject.width)
+                    customWidth: getCustomPluginWidth(pluginObject.size, pluginObject.width),
+                    hideMinimizeButton: pluginObject.hideMinimizeButton
                 },
                 $uiContainer = $($.trim(N.app.templates['template-plugin-container'](bindings)));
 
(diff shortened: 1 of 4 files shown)


kellyi:
Taking a look at this.

kellyi:
Working well!

screen shot 2017-11-28 at 12 42 23 pm

src/GeositeFramework/js/Plugin.js
```diff
@@ -475,7 +475,8 @@ require(['use!Geosite',
                     hasHelp: pluginObject.hasHelp,
                     hasCustomPrint: pluginObject.hasCustomPrint,
                     sizeClassName: getPluginSizeClassName(pluginObject.size),
-                    customWidth: getCustomPluginWidth(pluginObject.size, pluginObject.width)
+                    customWidth: getCustomPluginWidth(pluginObject.size, pluginObject.width),
+                    hideMinimizeButton: pluginObject.hideMinimizeButton
```
kellyi: Not sure it's worth changing but out of curiosity I checked and it seems like the minimize button will be hidden if `overrides.json` has any truthy value for `hideMinimizeButton`. Since it's an override that's probably the behavior we want -- a plugin developer would have to intentionally add the `hideMinimizeButton` key to overrides.json in the first place, which seems like a pretty good proxy for intent.

caseycesari: Yeah, I think that's okay. I modeled the implementation in the launchpad from regional planning: https://github.com/CoastalResilienceNetwork/regional-planning/blob/master/main.js#L66

Issue #1017, opened 2017-10-24
IE11: "Object doesn't support this action"


kellyi:
![screen shot 2017-10-24 at 10 16 06 am](https://user-images.githubusercontent.com/4165523/31948273-71ce6d20-b8a4-11e7-9a5e-46e46ef343b3.png)

PR #1031, opened 2017-11-28
Replace ES6 method with non-ES6 method syntax

kellyi:
## Overview

This PR fixes a problem whereby the site wouldn't load properly in IE11. The cause was a stray bit of ES6 method declaration syntax which isn't supported in IE11; the fix is to use `name: function() {` syntax.

Connects #1017

### Demo

screen shot 2017-11-28 at 2 26 23 pm

## Testing Instructions
- get this branch, then rebuild in VS and view the project in Chome, IE11, FF, and Safari and verify that everything still works as expected

### DIFF
--- a/src/GeositeFramework/js/SidebarToggle.js
+++ b/src/GeositeFramework/js/SidebarToggle.js
@@ -56,7 +56,7 @@
             'click': 'togglePluginVisibility'
         },
 
-        adjustToggleIcon() {
+        adjustToggleIcon: function() {
             const newIconClass =
                 this.viewModel.get('pluginContentVisible') ?
                 'fa fa-chevron-right' : 'fa fa-chevron-left';


src/GeositeFramework/js/SidebarToggle.js
```diff
@@ -56,7 +56,7 @@
             'click': 'togglePluginVisibility'
         },
 
-        adjustToggleIcon() {
+        adjustToggleIcon: function() {
```
caseycesari: Nice find. I noticed I did this in a few other places, I think just out of habit.

kellyi: Wonder if it's possible to add a linter to catch these in PRs automatically.

kellyi:
Thanks!

Issue #1026, opened 2017-11-20
New bug - hamburger menu under left nav


zferdana:
Team - not sure when this started, but see graphic below showing the bug. Tried this on multiple master framework sites. I will try rebuilding...

![screen shot 2017-11-20 at 12 52 30 pm](https://user-images.githubusercontent.com/5605392/33041091-fc2295f2-cdf1-11e7-891d-74c4c405542d.png)

caseycesari:
May be a side-effect of https://github.com/CoastalResilienceNetwork/GeositeFramework/pull/968

kellyi:
Tried reproducing this on my local without success:

<img width="856" alt="screen shot 2017-11-28 at 2 51 25 pm" src="https://user-images.githubusercontent.com/4165523/33341059-b9f637f0-d44b-11e7-8851-52d83444db26.png">

Do see it happening on the site from the issue's original image, though:

<img width="809" alt="screen shot 2017-11-28 at 2 51 34 pm" src="https://user-images.githubusercontent.com/4165523/33341076-c9c5b494-d44b-11e7-8a8b-c6604b9109c5.png">

kellyi:
Tried reverting the change here https://github.com/CoastalResilienceNetwork/GeositeFramework/pull/968 but that didn't fix it.

Dropping the `.nav-apps` zindex below 2800 does fix it, although I don't know what other things that might break:

<img width="1116" alt="screen shot 2017-11-28 at 3 27 50 pm" src="https://user-images.githubusercontent.com/4165523/33342619-b892aa2e-d450-11e7-95c5-f6b72cc1a3c5.png">

kellyi:
`header` and `header .nav-main` have a zindex of 2800:

https://github.com/CoastalResilienceNetwork/GeositeFramework/search?utf8=%E2%9C%93&q=2800&type=

kellyi:
I set up my local instance of the framework to have the same plugins to see whether I could reproduce the bug that way, but wasn't able to recreate it that way.

Going to check into why changing the z-index would fix it and make a PR.

zferdana:
Hi Casey and Kelly - do you have a timeline on this? Would be good to know, as we are launching maps.bluewaysconservation.org tomorrow using framework master. Just let me know - thx

kellyi:
Hi Zach! I managed to reproduce this bug locally and I'm about to put up a pull request to fix it. It should be in `master` by or before 2pm ET tomorrow.

PR #1032, opened 2017-11-29
Lower .nav-apps z-index

kellyi:
## Overview

This PR fixes a bug reported in #1026 whereby the hamburger dropdown menu appeared behind the `.nav-apps` sidebar.

This was caused by having a z-index too high on `.nav-apps`: it was 5001, and the dropdown is 2800. Now `.nav-apps` is 2500.

Connects #1026

### Demo

screen shot 2017-11-29 at 4 16 05 pm

### Notes

This bug is present on `master` and not `develop`, so we may have inadvertently fixed it in the latter. Going to look through the git history to see if I can find where.

## Testing Instructions
- get this branch, then rebuild and serve the app in VS
- visit the site in Chrome, FF, Safari, and IE 11 and verify that the menu dropdown appears above the nav bar

### DIFF
--- a/src/GeositeFramework/css/main.css
+++ b/src/GeositeFramework/css/main.css
@@ -792,7 +792,7 @@ nav {
     width: 220px;
     background-color: #333;
     position: relative;
-    z-index: 5001;
+    z-index: 2500;
     padding-bottom: 27px;
     overflow: visible;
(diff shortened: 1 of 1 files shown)


kellyi:
Thanks!

Issue #1007, opened 2017-10-12
Exercise new print workflow in sample plugin


caseycesari:
- Add content to what was formerly the print preview sandbox. This content should take user feedback.
- Setup a basic print layout that is somehow different than the default.
- Modify map (plugin already does this by adding a new layer, perhaps we could also change the zoom level or something), and then clean-up after print is finished.

caseycesari:
We should also include some documentation in the wiki.

caseycesari:
Make sure to clean-up the print modal. This could include moving the ok button to the bottom of the modal, and adding a cancel button.

PR #1038, opened 2017-12-04
Transform print preview map into print modal

caseycesari:
## Overview

Repurpose the print preview map into a customizable print modal. Plugin authors can provide content, like a form, that will be used to populate the modal. Then, before printing is kicked-off, the plugin will again be given a reference to the modal content. A plugin author can parse the content for user input, and then adjust the print output as necessary.

Connects #1009 

### Demo
tnc1

### Notes

- The modal and demo element will be cleaned-up in #1007 
- Opting out of the print modal causes the print workflow to fail. This is also the case on `master` when opting out of the print preview map. I spent a little bit of time trying to figure this out, but I was unable to. I'll create a follow-up issue.

## Testing Instructions

- Activate the "Identify point" plugin.
- Click the print icon.
- In the modal that appears, check the checkbox.
- Verify that the text "Demo element" appears on the right side of the print preview page.
- Repeat this process, but do not check the box. Verify that the demo element is not on the print preview.

### DIFF
--- a/src/GeositeFramework/Views/Home/Index.cshtml
+++ b/src/GeositeFramework/Views/Home/Index.cshtml
@@ -593,15 +593,14 @@
         </div>
     </script>
 
-    <script type="text/template" id="template-map-preview">
-        <div id="plugin-print-preview">
+    <script type="text/template" id="template-plugin-print-modal">
+        <div id="plugin-print-modal">
             <div class="popover popover-header">
-                <h2><%= pluginName %> - <span class="i18n" data-i18n="Print Preview">Print Preview</span></h2>
+                <h2><%= pluginName %> - <span class="i18n" data-i18n="Print Options">Print Options</span></h2>
             </div>
-            <div class="print-preview-container">
-                <span class="instructions i18n" data-i18n="Zoom to the view that you want printed in your map.">Zoom to the view that you want printed in your map.</span>
-                <button id="print-preview-print" class="button radius i18n" data-class="Print">Print</button>
-                <div id="plugin-print-preview-map"></div>
+            <div class="print-modal-container">
+                <button id="print-modal-confirm" class="button radius i18n" data-class="Print">Ok</button>
+                <div id="plugin-print-modal-content"></div>
             </div>
         </div>
     </script>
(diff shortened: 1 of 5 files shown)


src/GeositeFramework/Views/Home/Index.cshtml
```diff
@@ -593,15 +593,14 @@
...
+        <div id="plugin-print-modal">
             <div class="popover popover-header">
-                <h2><%= pluginName %> - <span class="i18n" data-i18n="Print Preview">Print Preview</span></h2>
+                <h2><%= pluginName %> - <span class="i18n" data-i18n="Print Options">Print Options</span></h2>
             </div>
-            <div class="print-preview-container">
-                <span class="instructions i18n" data-i18n="Zoom to the view that you want printed in your map.">Zoom to the view that you want printed in your map.</span>
-                <button id="print-preview-print" class="button radius i18n" data-class="Print">Print</button>
-                <div id="plugin-print-preview-map"></div>
+            <div class="print-modal-container">
```
mmcfarland: This may be intended for #1007, but the `Ok` seems like it should be in a footer container and maybe also include a `Cancel` button that short-circuits the process.

src/GeositeFramework/js/PluginBase.js
```diff
@@ -40,11 +40,11 @@ define(["dojo/_base/declare",
...
 
-            // Show a print preview map for plugin printing
-            usePrintPreviewMap: false,
+            // Show a modal dialog for plugin printing
+            usePrintModal: false,
 
-            // The [width, height] of the print preview map
-            previewMapSize: [500, 400],
+            // The [width, height] of the print modal
+            printModalSize: [500, 400],
```
mmcfarland: The size was set specifically since the preview map needed them set explicitly.  Given that this will have form content going forward, are the sizes still needed?  I think it would be convenient if the container fit the contents, but it may be more trouble than it's worth.

src/GeositeFramework/Views/Home/Index.cshtml
caseycesari: Yeah, I was reserving that type of stuff for #1007.

src/GeositeFramework/js/PluginBase.js
caseycesari: I think this is a good idea, and I created a follow-up issue: https://github.com/CoastalResilienceNetwork/GeositeFramework/issues/1042.

caseycesari:
Thanks for the review.

Issue #1012, opened 2017-10-13
Fix scale bar placement in plugin print


caseycesari:
The scale bar should be placed in the bottom corner of the map, but for some reason is going to the bottom corner of the printed page in the plugin print workflow.

PR #1039, opened 2017-12-05
Place scalebar inside of map for plugin print

caseycesari:
## Overview

This change prevents the scalebar from moving to the bottom of the page, and keeps it within the map container so that it is always visible.

Connects #1012 

### Demo

Before:
image

After:
image

### Notes

Original comment triggering this issue: https://github.com/CoastalResilienceNetwork/GeositeFramework/pull/1005#issuecomment-336263041

## Testing Instructions

- Activate the identify point plugin.
- Click the plugin print button.
- Advance through the print modal.
- In the print preview window (or PDF, if you are using a browser other than chrome), verify that the scalebar is within the map.

### DIFF
--- a/src/GeositeFramework/css/print-a4-landscape.css
+++ b/src/GeositeFramework/css/print-a4-landscape.css
@@ -28,7 +28,7 @@
     }
 
     .page-bottom {
-        bottom: 1%;
+        bottom: 2%;
     }
 }
 
--- a/src/GeositeFramework/css/print-a4-portrait.css
+++ b/src/GeositeFramework/css/print-a4-portrait.css
@@ -28,7 +28,7 @@
     }
 
     .page-bottom {
-        bottom: 1%;
+        bottom: 2%;
     }
 }
 
--- a/src/GeositeFramework/css/print-letter-landscape.css
+++ b/src/GeositeFramework/css/print-letter-landscape.css
@@ -28,7 +28,7 @@
     }
 
     .page-bottom {
-        bottom: 1%;
+        bottom: 2%;
     }
 }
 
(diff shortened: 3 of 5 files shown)


kellyi:
Taking a look at this!

kellyi:
It appears on the map in the print pdf now, albeit in a way that cuts off the bottom label in landscape orientation:

screen shot 2017-12-05 at 2 32 15 pm

Portrait looks good though:

screen shot 2017-12-05 at 2 33 54 pm

kellyi:
This is probably fine but since we're already manipulating the scalebar here, I wonder if it makes sense to add a special CSS rule triggered when it's appended to `#map-0_root` to move it up like 5 or 10 pixels -- and avoid cutting off the bottom labels?

caseycesari:
> This is probably fine but since we're already manipulating the scalebar here, I wonder if it makes sense to add a special CSS rule triggered when it's appended to #map-0_root to move it up like 5 or 10 pixels -- and avoid cutting off the bottom labels?

Yeah, I'll make a fix here.

caseycesari:
I didn't realize it initially, but the above screenshots were from framework print, not plugin print. It was an easy fix though, so I pushed up a commit to adjust the scalebar position in framework print.

kellyi:
Taking another look at this.

kellyi:
Heads up that there's a merge conflict.

kellyi:
LGTM:

screen shot 2017-12-08 at 10 09 51 am

screen shot 2017-12-08 at 10 10 55 am

caseycesari:
Thanks for the review!

Issue #1034, opened 2017-11-29
Map tools dialog is hidden behind legend


caseycesari:
![image](https://user-images.githubusercontent.com/1042475/33400527-6a2b78fa-d524-11e7-900e-dabd7e057700.png)

PR #1040, opened 2017-12-05
Ensure map tools menu is displayed above legend

caseycesari:
## Overview

Adjust the z-index of the map tools parent container so that is displayed above the legend container.

Connects #1034

### Demo

Before:
image

After:
image

## Testing Instructions

- Open the regional planning plugin and add a bunch of layers to the map so that the legend container grows to it's max height.
- Click the map tools button, and ensure the menu is displayed above the legend container.

### DIFF
--- a/src/GeositeFramework/css/main.css
+++ b/src/GeositeFramework/css/main.css
@@ -1553,6 +1553,7 @@ header .nav-main {
 .control-container {
     position: absolute;
     width: 100%;
+    z-index: 102;
 }
 .control-container.subregion-active {
     top: 40px;
@@ -1645,8 +1646,6 @@ header .nav-main {
     width: 255px;
     min-width: 255px;
     z-index: 101;
-    /* show over plugin-container */
-
     font-weight: bold;
     background: #FFF;
     font-size: 12px;


kellyi:
Taking a look at this one too.

kellyi:
Chrome:

screen shot 2017-12-05 at 2 43 48 pm

IE11:

screen shot 2017-12-05 at 2 45 52 pm

src/GeositeFramework/css/main.css
```diff
@@ -1553,6 +1553,7 @@ header .nav-main {
 .control-container {
     position: absolute;
     width: 100%;
+    z-index: 102;
```
kellyi: For reference it seems like this is why we're using 102 here:

https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/ea036a0d3c9051bdd2876fcf48187b45db9aa10e/src/GeositeFramework/css/main.css#L1581

kellyi:
I noticed this comment beneath the z-index: 101 line for the legend:

https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/ea036a0d3c9051bdd2876fcf48187b45db9aa10e/src/GeositeFramework/css/main.css#L1582

Is it worth finding out whether that's still valid? If it's not, it's probably worth deleting the comment. If it is, we'll have to interleave these differently since the legend does now appear beneath the tools icon:

screen shot 2017-12-05 at 2 55 58 pm

FWIW my sense is that the order expressed in the comment isn't valid anymore since they're now in the sidebar.

caseycesari:
> I noticed this comment beneath the z-index: 101 line for the legend:

Yeah, I don't think that comment is relevant. I'll remove it.

caseycesari:
Pushed an updated commit that removes the comment.

caseycesari:
Thanks for the reviews.

Issue #1043, opened 2017-12-07
App does not appear in Single App Mode


cschneebeck:
- Hawaii region [development branch](https://github.com/CoastalResilienceNetwork/hawaii-region/tree/development)
- region.json specifies [ecological-effects-slc](https://github.com/CoastalResilienceNetwork/ecological-effects-slc/tree/development) as the pluginFolderName
- The site builds successfully, but the app does not load in the left pane and the left pane completely covers the map
- An error is thrown when trying to load underscore.js

![image](https://user-images.githubusercontent.com/11446978/33733512-3d825db8-db47-11e7-9567-7124c6e36138.png)

ajrobbins:
Related to https://github.com/CoastalResilienceNetwork/GeositeFramework/issues/669

cschneebeck:
FYI, to make sure the app wasn't causing the problem, I tested the same region with a different app. I switched ecological-effects-slc out with marsh-explorer and ran the build again. I received the same error trying it with marsh-explorer so I switched it back to ecological-effects-slc.

Issue #669, opened 2016-09-12
Underscore not available


mmcfarland:
We have two reports of the underscore library not resolving to an object and throwing errors when functions are accessed in plugin startup scripts:

Regional Planning
`main.js:66 Uncaught TypeError: _.isUndefined is not a function`

Inspecting the `_` variable in the `declare` function shows that  `_` is equal to the number `3`, which happens to be the last character in the `require:packages:location` property (`//cdnjs.cloudflare.com/ajax/libs/underscore.js/1.8.3`)

The problem was also reported from an app developer, and may indicate the error is not related to Regional Planning:

> The Puget Sound Coastal Defense plugin works with the master repo of the framework (at least the version on maps.coastalresilience.org) but is not compatible with the current development version.  When included it causes some exception/error when starting and the framework renders the header bar only.  What error is logged references the regional planning plugin (`Uncaught TypeError: _.isUndefined is not a function`) but I'm sure relates to some conflict with PS coastal defense and not regional planning as the same occurs for other plugins if regional planning is left out but coastal defense is included.

PR #1044, opened 2017-12-08
Ensure underscore is available through Use config

caseycesari:
## Overview

Import underscore locally and add it to the Use config so that plugins
can reference it.

Plugins that don't directly import underscore reference it through
Use, but it is properly made available through that interface.
Because of that, require.js tries to find it in the ESRI/DOJO package
and can't, which causes underscore to be undefined.

Connects to #1043 
Connects to #669

## Testing Instructions

- Before switching to this branch, bring this plugin into your framework: https://github.com/CoastalResilienceNetwork/ecological-effects-slc and temporarily remove all other plugins.
- Load up the framework. You should see the error reported in #669.
- Now clone this branch and restart the dev server.
- Reload the site, the error should be gone and the plugin should load.
- Try adding in other plugins, and ensure things still work.


caseycesari:
@mmcfarland @kellyi Assigned you both because this is one of those weird bugs that could be fixed in one environment and broke in another.

caseycesari:
@mmcfarland The plugin brings it in here: https://github.com/CoastalResilienceNetwork/ecological-effects-slc/blob/master/js/AgsDrawPolygon.js#L1, and makes use of it a few times, like here: https://github.com/CoastalResilienceNetwork/ecological-effects-slc/blob/master/js/AgsDrawPolygon.js#L123.

I noticed that underscore was available globally, but maybe at one time we had to use `use`, because I noticed other plugins that did this as well, like https://github.com/CoastalResilienceNetwork/flood-slr.

caseycesari:
Thanks for the reviews!

Issue #1036, opened 2017-11-30
Set border-radius to 0 for plugin buttons.


caseycesari:
Chrome has started to add a border radius to these buttons, which makes them look odd:

![image](https://user-images.githubusercontent.com/1042475/33443818-8b7464ca-d5c6-11e7-818a-86da9b979347.png)

![image](https://user-images.githubusercontent.com/1042475/33443804-824fe216-d5c6-11e7-8d48-a87ef99bde04.png)

PR #1052, opened 2017-12-15
Ensure plugin button border radius is set to 0

caseycesari:
## Overview

Chrome temporarily included base styling which gave buttons a border radius of 4. This resulted in the plugin buttons looking odd. Chrome appears to have removed this style rule, but an override is added here in case it comes back, and for older versions of the browser.

Connects to #1036

### Demo

image

## Testing Instructions

- Ensure the added rule is valid and assigned to the correct class.


caseycesari:
Thanks

Issue #1024, opened 2017-11-13
Sidebar toggle arrow is out of sync with plugin state


caseycesari:
![image](https://user-images.githubusercontent.com/1042475/32742789-682d593c-c878-11e7-8a73-090ef3e618f1.png)

![image](https://user-images.githubusercontent.com/1042475/32742795-6c17f7c8-c878-11e7-8de3-f038c61bfadf.png)

PR #1053, opened 2017-12-18
Fix single plugin mode plugin toggle button

caseycesari:
## Overview

The icon for the single plugin mode plugin toggle button was reversed based on the state of the plugin. That is fixed here.

Also, while working on the above issue, I discovered a bug related to the infographic button and single plugin mode. The button was being rendered in single plugin mode when it shouldn't have been. It was also being rendered multiple times. Both of those issues are fixed here.

Connects #1024 

### Demo

tnc1

## Testing Instructions

- Pull down this branch.
- Turn on single plugin mode.
- Toggle the visibility of the plugin.
- Verify that the icon in the toggle button stays in sync with the state of the plugin.
- Verify that the infographic icon is not rendered.
- Turn off single plugin mode and restart the dev server.
- Verify that the infographic button is present for the identify point plugin, and that there is only one button element in the DOM for that plugin.

### DIFF
--- a/src/GeositeFramework/js/Plugin.js
+++ b/src/GeositeFramework/js/Plugin.js
@@ -409,6 +409,11 @@ require(['use!Geosite',
             render(view);
             view.$el.appendTo($parent);
             createUiContainer(view, paneNumber);
+            if (model.get('pluginObject').infographic && !N.app.singlePluginMode) {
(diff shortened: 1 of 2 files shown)


caseycesari:
Thanks!

Issue #1050, opened 2017-12-12
Verify legends from plugins that use image services appear in printed legend


caseycesari:
Check out NC restoration explorer.

PR #1054, opened 2017-12-19
Fix rendering of plugin legends in print

caseycesari:
## Overview

Changes introduced in #1051 exposed an issue with the solution introduced in #1027 to include plugin legends on printed maps.

In #1027, the plugin legends DOM element was moved into the layer legends DOM element for printing. However, the layer legends DOM element is dynamically updated and gets redrawn when the map updates. Because the changes in #1027 force many more map updates to
ensure layers don't get out of sync when the map size changes, the plugin layers element is overwritten when the layer legends element is rerendered.

I was able to remove the solution introduced in #1027, and make a much simpler fix, which is to move the plugin legends element ahead of the layer legends element in the DOM. This does have the side-effect of placing the plugin legends first in the legend container. It's not totally clear why this solution works, other than it being a result of the CSS rules in place for the legend during printing. Much work has already been invested in these rules, so I did not want to modify them any further. I did, however, include some new style rules for plugin legends to add a minimal set of constraints on them since they are often very different from one another.

Connects to #1050

### Demo

Screenshot from Chrome demonstrating multiple plugin legends and layer legends coexisting.

image

Screenshot from Firefox demonstrating that the Multiexplorer/Restoration Explorer plugin legend does show up in the printed legend. It does not currently show up in Chrome.

image

## Note

#1050 was created to verify that legends for image layers showed up in the printed legend. Since all plugin legends are now included in the printed legend, it does not matter what types of layers a plugin uses. However, the legend for multi-explorer (this is the plugin that was suggested to us for testing) does not show up in the printed legend for Chrome. It does work in Firefox and IE, however. It's not clear why this is happening, but it most likely has something to do with how the legend for this particular plugin is constructed. Other SVG legends do show and are included in the example screenshot above.

## Testing Instructions

- Add several plugins to your framework installation.
- Toggle on the plugins as well as some layers from the regional planning plugin.
- Use the "Create Map" feature to print the map
- Verify any plugin legends that were in map legend are displayed in the printed legend.

### DIFF
--- a/src/GeositeFramework/Views/Home/Index.cshtml
+++ b/src/GeositeFramework/Views/Home/Index.cshtml
@@ -294,8 +294,8 @@
                             <a class="legend-close i18n" data-i18n="[title]Hide Legend" href="javascript:;">_</a>
                         </div>
                         <div class="legend-body">
-                            <div class="layer-legends"></div>
                             <div class="plugin-legends"></div>
+                            <div class="layer-legends"></div>
                         </div>
                     </div>
                     <div class="subregion-tooltip"></div>
--- a/src/GeositeFramework/css/app-print.css
+++ b/src/GeositeFramework/css/app-print.css
@@ -123,6 +123,15 @@
         margin: 0 10px 0 0;
     }
 
+    div.plugin-legends {
+        display: flex;
+    }
+
+    div.custom-legend {
+        min-width: 100px;
+        max-width: 300px;
+    }
+
     .legend-layer img {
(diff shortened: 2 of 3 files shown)


caseycesari:
That was actually one of the problems I discovered while testing plugins with image layers (the cause is described in the PR description), and that this code change is aimed at. I rebased this PR on `develop`, and tested it again, and I'm not seeing the behavior you reported anymore. Mind giving this another try?

Issue #1046, opened 2017-12-12
Single vs multiple app mode in develop branch


zferdana:
Casey - just want to make sure the develop branch can do both single and multiple app mode. I just tried --framework-branch=develop and got this:

![screen shot 2017-12-11 at 4 59 14 pm](https://user-images.githubusercontent.com/5605392/33862035-9d425900-de95-11e7-866d-d215677d7972.png)

I had to deploy the master framework in order to produce the dev oceanwealth region. I suspect we will need both modes working well on the develop branch to test both the super app in Hawaii as well as the in-app print (Jacob will test on dev oceanwealth).

caseycesari:
@zferdana Hi Zach. The `develop` branch can do both single and multiple app mode. It is configured via the region.json. There is likely something else causing the above problem.

Can you provide the full build command you ran so that I can try and reproduce this locally?
 Thanks.

zferdana:
Sure - I didn't change anything to region.json since I was building our traditional, multiple plugin framework mode.

build.py oceanwealth-region --prod --framework-branch=develop --region-branch=development

caseycesari:
@zferdana Can you try running the command again? I was able to reproduce the issue you were experiencing. I then merged https://github.com/CoastalResilienceNetwork/regional-planning/pull/95 into master for regional planning (this is related to https://github.com/CoastalResilienceNetwork/GeositeFramework/issues/1043), tried the build command again, and the site worked.

Thanks.

matts5517:
Hey, I am also getting the same problem as above when building in Develop
my code: 
build.py wisconsin-region FreshwaterNetwork --dev --framework-branch=develop --region-branch=development.

I want to use the legend single column functionality which i think is only under the develop branch for now?

Thanks,
Matt

My console:

![screen shot 2018-01-03 at 12 30 22 pm](https://user-images.githubusercontent.com/19394069/34536312-f8189648-f081-11e7-8ccf-6c22dd8d5c72.png)

caseycesari:
Hi @matts5517,

[...]

PR #1055, opened 2018-01-04
Attempt to support various underscore.js imports

caseycesari:
## Overview
Adjust the use config to support all of the various ways that underscore
is imported within plugins.

The previous fix implemented in #1044 caused problems with plugins that
bring in their own version of underscore in via a CDN. The adjustments
here should make that possible while still allowing plugins to import
underscore via Use.js.

Also, rename the framework version of underscore to reduce confusion.

Connects to #669, #1043, #1046

### Notes

I'm not totally sure why this is working. It was arrived at through various tweaks of the Use config until various plugins that use underscore were all working.

## Testing Instructions

- Bring in regional-planning, flood-slr, and living-shorelines-explorer, and any other plugins that you know of that use underscore.
- Launch the framework, and ensure all of the plugins are working and that there are no errors related to underscore in the console (either errors finding underscore or undefined underscore methods.).
- Ensure regional-planning works before and after https://github.com/CoastalResilienceNetwork/regional-planning/pull/95. This will exercise two different approaches to using underscore.


src/GeositeFramework/Views/Home/Index.cshtml
```diff
@@ -815,7 +815,8 @@
         useConfig['jqueryui'] = { attach: 'jqueryui' };
         useConfig['chosen'] = { attach: 'chosen', deps: ['jquery'] };
         useConfig['pips'] = { attach: 'pips', deps: ['jqueryui'] };
-        useConfig['underscore'] = { attach: 'underscore' };
+        useConfig['underscore'] = { attach: '_', deps: ['underscore'] };
+        useConfig['underscore-min'] = { attach: '_', deps: ['underscore'] };
```
caseycesari: Not sure why `underscore` has to be a dependency of `underscore` here, but if you don't do this underscore is defined as `3` in some plugins.

src/GeositeFramework/Views/Home/Index.cshtml
```diff
@@ -843,7 +844,9 @@
                 "jqueryui": location.pathname.replace(/\/[^/]+$/, "") + "js/lib/jquery-ui.min",
                 "chosen": location.pathname.replace(/\/[^/]+$/, "") + "js/lib/chosen.jquery.min",
                 "pips": location.pathname.replace(/\/[^/]+$/, "") + "js/lib/jquery-ui-slider-pips.min",
-                "underscore": location.pathname.replace(/\/[^/]+$/, "") + "js/lib/underscore-min"
+                "underscore": location.pathname.replace(/\/[^/]+$/, "") + "js/lib/underscore",
+                "underscore/underscore-min": location.pathname.replace(/\/[^/]+$/, "") + "js/lib/underscore",
+                "underscore/underscore": location.pathname.replace(/\/[^/]+$/, "") + "js/lib/underscore",
```
caseycesari: I think this is capturing cases where underscore is brought in via a CDN. I'm attempting to redirect them to the framework underscore.

kellyi:
Set up these plugins to test:

```sh
ecological-effects-slc/
flood-slr/
future_habitat_v2/
launchpad/
living-shorelines-explorer/
natural-coastal-protection/
recreation_tourism/
regional-planning/
subregion_toggle/
tuna-fisheries/
```

kellyi:
Tried it out with those plugins and verified that 1. they won't work on `develop` due to the underscore issue and 2. they do load on this branch:

screen shot 2018-01-04 at 11 52 25 am

I don't know whether this is known, prexisting, or newly-created problem but it looks like future habitat can't find `pips`:

screen shot 2018-01-04 at 11 54 18 am

kellyi:
(That was for the `master` branches of those plugins, btw; going to check the `development` branches now).

kellyi:
This seems to work in almost all cases. I was able periodically to reproduce the bug with Living Shorelines in Firefox with the following steps:

- have Living Shorelines, ENBA, and Launchpad as the installed plugins
- load the app in Firefox and by or before the time the launchpad screen opens, click on the Living Shorelines plugin to open it.

I'd see lines related to underscore errors in the console:

screen shot 2018-01-04 at 12 59 06 pm

It did take some work to cause that to happen though, but it makes me wonder whether it's an underlying timing issue -- and might pop up in slower browsers?

caseycesari:
Thanks for the thorough reviews.

Issue #1059, opened 2018-01-08
Look into repositioning map in print sandbox


caseycesari:
> 7. I am having difficulty positioning and sizing the map element on the page.  I used the following css to make adjustments to the position:

```
#plugin-print-sandbox .map .esriMapContainer {
    display: block;
    position: absolute;
    top: 10mm;
    height: 157mm;
    width: 202mm;
    left: 5mm;
    overflow: hidden;
}
```

> This is tricky, because the map and the legend are in the same container, but often need to be positioned and sized differently.  If the legend were cloned (like the above example) and placed outside of the map container, this would be easier.  Also, the map is not respecting my height and width values in physical units.  Adjusting those values does seem to adjust the size of the map.  However, it is not corresponding with what those units should actually represent.  The above values are the units I used in the Maine print plugin (and I confirmed by measuring with a ruler on paper).  However, they appear much smaller here, like some scaling factor is being applied.  Having the map as it's own standalone absolutely positionable object in the print page would make styling much easier.  At a minimum, some documentation should be provided on the best practice for positioning and sizing the map element when customizing print.

PR #1064, opened 2018-01-10
Reposition legend during plugin print

caseycesari:
## Overview

Based on feedback from @lesserj, moving the legend out of the map container for the plugin print workflow should aid plugin developers in positioning the map and legend on the printed page. Also updates the identify point plugin print css to demo how to position the legend for printing.

Connects #1059 

### Demo

*A demonstration of plugin print from the identify point plugin*
image

## Testing Instructions

- Activate the regional planning plugin and add a few layers to the map.
- Activate the identify point plugin, and click the print icon in the plugin header.
- Place a breakpoint [here](https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/36444eda3342c756c257bcf8de6fe24689257ed9/src/GeositeFramework/js/Plugin.js#L586).
- Click the print button.
- The breakpoint should be triggered. In the inspector, verify that the legend is in the print sandbox as a sibling of the map container.
- Continue the breakpoint, and return to the application.
- In the inspector verify that the legend is child of the map container.

### DIFF
--- a/src/GeositeFramework/js/Plugin.js
+++ b/src/GeositeFramework/js/Plugin.js
@@ -631,6 +631,10 @@ require(['use!Geosite',
                         map.resize();
                         map.reposition();
 
+                        // Move the legend out of the map container for easier styling
+                        var legendNode = $("#legend-container-0").detach();
+                        $(legendNode).appendTo($printSandbox);
+
                         modalConfirmDeferred.resolve();
 
                         // Pass the modal contents to the plugin,
@@ -646,9 +650,11 @@ require(['use!Geosite',
                 },
 
                 closejs: function () {
-                    // Move the map back to it's original container
+                    // Move the map and legend back to the original container
                     var mapNode = $('#map-0').detach();
+                    var legendNode = $("#legend-container-0").detach();
                     $('.map-container').append(mapNode);
+                    $('#map-0').append(legendNode);
                     map.resize(true);
 
                     // Move the scalebar back to it's original location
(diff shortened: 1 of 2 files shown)


caseycesari:
@lesserj If you have a minute, take a look at the code changes here to see if this will work as a solution to #1059. Thanks.

lesserj:
@caseycesari I'll give it a look shortly

lesserj:
I'm getting an error when trying to build the framework (Not related to this branch), so I haven't been able to test yet... but in general, I'm liking this.  I've been thinking about this a bit and I think the general approach here should be something like this:

Map and legend are in their own containers that are absolutely positioned so that they are not in the document flow with any plugins custom print content that is injected into the content area.  That way, there are three independently stacked layers that do not "push" each other around in the dom -- one for the map, one for the legend and one for the custom content.  We can then use CSS to independently position and size each of those components.  Setting a width and height css value on the map container and the legend container will control the width and height of the map and legend respectively.   So, if I specify that the map is 100mm wide and 50mm tall and is positioned 25mm from the top and 45mm from the left, you'll get an appropriately placed item.  Alternatively, we could use percentages here, if we were making the content flexible for multiple page sizes.

lesserj:
This change has made everything much easier for me and pretty much addresses the issues I was having with https://github.com/CoastalResilienceNetwork/GeositeFramework/issues/1059.  With this change and applying the right css selectors, I am able to position and style both the map and the legend independently.  Thanks @caseycesari

Issue #1062, opened 2018-01-08
Call print process after map has updated


caseycesari:
Attempt to have `window.print` (used [here](https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/develop/src/GeositeFramework/js/Plugin.js#L586)) called by the `update-end` event.

From Jacob Lesser:

> 9. If the map area on the screen is smaller than the map area on the print, the tiles do not finish loading before window.print is called and the map on paper does not fill the allocated size.  I addressed this in my Maine print plugin by loading a css rule to resize the map before calling print to the size I needed on the paper.  I then used the ArcGIS callback methods to resize the map and then wait for it to finish loading before calling the window.print function.  The css to resize the map needed to be a screen rule and not a print rule to make sure that it was executed before window.print suspended execution.

> ```
> this.map.on('update-end', function() {
>     window.print();
> }
> ```

> This has the added benefit of making sure other layers that the user may have toggled on before printing finish loading before calling the window.print function.

caseycesari:
FYI @lesserj 

After experimenting with adding this [within the framework](https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/71ef9ecc896178dda5db4b5ddbde76e245f526e5/src/GeositeFramework/js/Plugin.js#L586), I think this should be implemented in the plugin's [postPrintModal function](https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/71ef9ecc896178dda5db4b5ddbde76e245f526e5/src/GeositeFramework/js/PluginBase.js#L89). 

I created a demo of this functionality in https://github.com/CoastalResilienceNetwork/GeositeFramework/pull/1067.

I'm recommending this because there are some cases where map doesn't need to update before printing. In these cases, waiting for the `update-end` event to fire doesn't make sense. The Maine region/future habitat v2 plugin implementation get around this by checking if the map is updating before binding print to the `update-end` event. However, this requires adding a `setTimeout` call. If you immediately check if the map needs to be updated after adding a layer, for example, the map doesn't appear to need to be updated. This leads to a complicated implementation and a delay in the print process on the framework side for a case that may not happen all the time.

This also ensures that there are no setTimeouts in the core plugin print process. think it’s worth striving to maintain this since adding setTimeouts to the core process can introduce timing issues and race conditions.

lesserj:
[...]

PR #1067, opened 2018-01-15
Demo printing after map update

caseycesari:
## Overview

Update the identify point plugin print to with an asynchronous action (adding a layer to the map). This requires waiting for the map to update before proceeding with printing. 

The timeout is extended to give the map object time to realize it needs to update. Proceeding too quickly results in the layer being added after printing.

This implementation is largely based off of https://github.com/CoastalResilienceNetwork/future_habitat_v2.

Connects to #1062

## Testing Instructions

- Place a breakpoint on line 143 of `src/GeositeFramework/sample_plugins/identify_point/main.js`.
- Activate the plugin print process for the identify point plugin.
- In the modal that appears, add the north arrow and logo, but not the layer, and proceed.
- Verify the breakpoint is not triggered, and that the two images are on the map in print preview.
- Dismiss and activate the print process again.
- Check the box to add the layer to the map, and proceed.
- Verify that the breakpoint is triggered. 
- Proceed and verify that the layer is on the map in print preview.

### DIFF
--- a/src/GeositeFramework/sample_plugins/identify_point/html/print-form.html
+++ b/src/GeositeFramework/sample_plugins/identify_point/html/print-form.html
@@ -8,4 +8,9 @@
         <input type="checkbox" name="checkbox" value="value" id="tnc-logo">
         Add TNC Logo
     </label>
+    <br>
+    <label>
+        <input type="checkbox" name="checkbox" value="value" id="add-layer">
+        Add Sample Layer
+    </label>
 </form>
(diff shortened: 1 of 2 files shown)


caseycesari:
This is pointed at `cpc/hide-legend-controls-in-print` because that branch now contains many small fixes and improvements to the printed map position and sizing.

caseycesari:
After this makes it in to `develop`, I'm going to update the documentation to mention this approach.

caseycesari:
@kellyi Rebased on `develop`.

kellyi:
Size issue seems to be fixed; however, it also seems like I'm not able to print with the sample layer a second time after I've already printed with the sample layer once?

To reproduce:

- open identify point print, select all options and print
- verify that the print preview opens
- click cancel
- open identify point print, select all options and print
- print preview doesn't seem to open again

kellyi:
Oh, there's also a specific error in the console when trying to click print a second time:

screen shot 2018-01-16 at 10 56 12 am

caseycesari:
It looks like sometimes the map gets stuck in an updating state. I'm looking into it. Thanks.

caseycesari:
@kellyi Realized there was a typo in my above comment and it was not clear that I added a fixup to address the problem you reported.

kellyi:
Taking another look at this!

caseycesari:
Thanks!

PR #1070, opened 2018-01-22
Exclude some layers from legend

cschneebeck:
If layer name starts with underscore ("_"), don't add that layer to
legend.


cschneebeck:
I added maximum-scale=1 to the content portion of the viewport meta tag. Before it just had initial-scale=1. Once the site opened, a click on the chosen select menu caused the site to zoom in on phones. If a user didn't adjust the zoom, the map controls were not visible when users clicked the map button. Adding maximum-scale=1 prevents the site from zooming in on phones and offers a better user experience. 

<meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1">

Issue #1072, opened 2018-01-26
Print is triggered multiple times


lesserj:
The new print feature gets triggered multiple times when clicking the print button.  Best case scenario, this causes the system print window to appear several times after hitting cancel.  If the content insertion in the prePrintModal function uses appends rather than replaces, then content will be inserted multiple times.  I have not been able to figure out the root cause of the issue.  Anecdotally, the region framework has three plugins that have a custom print function, and the print is being called three times.  This may or may not be related.  The number of times it tries to print increases every time the plugin print button is clicked.

![multiple-print-calls](https://user-images.githubusercontent.com/2304500/35453231-46c8f9fa-027e-11e8-9674-b9be8dfa2a8f.gif)

kellyi:
Spent some time trying to figure out how to recreate a minimal version of this bug in advance of reviewing  #1080 and I am able to recreate it by:

- setting up the base GeositeFramework
- adding the regional planning plugin
- setting `"hasCustomPrint": true` in the regional-planning plugin's overrides.json file

Once that's done, clicking the plugin's custom print button seems to create three print windows, and on canceling each of them an error like this is reported in the browser console:

![screen shot 2018-08-29 at 3 46 32 pm](https://user-images.githubusercontent.com/4165523/44811504-d916c580-aba2-11e8-915f-aa28f8bd9995.png)

PR #1080, opened 2018-08-14
Make sure plugin click events are only bound once

lesserj:
## Overview

This PR resolves the issue in #1072.  Events are getting assigned more than once, and will trigger once for every assignment.  In the case of NCP, the print button will cause the print window to open three times and eventually cause the browser to crash in Chrome.

Connects #1072

### Demo

See #1072 for screenshot

### Notes

There are other ways to bind events that may be better, such as delegating the event to the events before the elements are even created and only calling the delegation once.  The implemented method is quick to implement without refactor and is performant.

## Testing Instructions

Click the print country summary button in NCP.  The print window should only appear once.


kellyi:
Hi @lesserj! Thanks for fixing this bug!

I was able to recreate the bug described in #1072 and narrowed it down to something happening in the regional-planning plugin's custom print functionality. The changes here fix the bug while also still allowing that plugin & others to print as expected.

I did want to ask if you could retarget this pull request at `develop` rather than `master`? I tried to update the base branch through the GitHub UI, but doing so caused the retargeted PR to also include two unrelated merge commits from past releases along with the bugfix in 23f6fd1d2bc5f7fb87b0a4014f5e6741d6fc809b

It appears that the branch on which you originally did the work branched off `master`, so the process for adjusting that will involve

1. rebasing the `prusikmapping-master` branch onto `develop` so that it will include the print-related change but not the merge commits
2. force pushing `prusikmapping-master` to your fork's GitHub repository, which will then update the branch here with the branch containing only the print bugfix.
3. updating the target branch for this PR from `master` to `develop` via the GitHub user interface by clicking the "Edit" button and adjusting the target branch from the dropdown depicted here:

screen shot 2018-08-29 at 1 56 13 pm

If any of these steps are unclear, definitely let me know and I can help facilitate the process!

After the rebasing and retargeting I can approve this PR so it can be merged into `develop`.

I'll also take a look at the regional-planning plugin's code to see how it was causing the error in the first place so we can determine whether a change there might be warranted too.

kellyi:
Hi @lesserj! I thought I'd comment again just in case this got missed in the shuffle of the holiday weekend.

Let me know if you'd like any help on rebasing this to prepare it for merging!

lesserj:
@kellyi This had fallen off my radar.  I will go ahead and rebase later today.

lesserj:
Closing in favor of https://github.com/CoastalResilienceNetwork/GeositeFramework/pull/1105 which is rebased off develop.

Issue #1081, opened 2018-08-15
Chrome JS version bug


laflessn:
A recent change in Chrome caused all older explorer apps to fail, therefore preventing the entire site from loading in chrome. Other browsers like Firefox worked fine. As a band-aid, we removed the problem apps which restored the sites in Chrome, but we need a long term fix. From the error message Casey found below, upgrading from JS 3.11 to 3.24 would likely solve this issue, this upgrade is also needed to bring in openstreetmap as an esri base map option. 

“The page is using the arcgi js API 3.11 (which refers to xstyle). The older xstyle versions do have a bug concerning pseudo-css-elements (they need empty strings as retunr value instead of the correct 'normal'). Chrome v67 now returns the correct 'normal' causing the xstyle APi to raise errors. 
 You should try to upgrade to the arcgis js API version 3.24”

caseycesari:
@laflessn The upgrade to 3.25 has been merged to the `develop` branch. Let us know if this fixes the above reported issue.

laflessn:
I have successfully confirmed and tested this in various regions. Please push to master.

caseycesari:
Thanks @laflessn. I'll let you know when this is in master.

PR #1095, opened 2018-09-03
Upgrade to Arcgis JS API 3.25

fungjj92:
## Overview

JSAPI v 3.11 uses an older version of dojo and xstyle. xstyle handles pseudo-css-elements in a way that is incompatible with Chrome and in a past CSS standard, and modern Chrome simply won't load the framework. Newer versions of JS API (3.24+) should fix.

ESRI [recently added](https://www.esri.com/arcgis-blog/products/js-api-arcgis/mapping/new-osm-vector-basemap/) the OpenStreetMap (OSM) basemap. Upgrading the JSAPI version should allow access.

Connects #1081 

### Demo

Missing icons on upgrade:
screen shot 2018-08-30 at 5 22 09 pm

Icons reinstanted with some code changes:
screen shot 2018-08-30 at 4 59 04 pm

### Notes

screen shot 2018-08-31 at 4 48 26 pm

There's still a bunch of console errors. The first 3 are also seen on develop or production sites, but the CORS errors are new. My instinct was to fake a header in `proxy.ashx` [sniffing out this ESRI CORS documentation](https://developers.arcgis.com/javascript/3/jssamples/exp_cors_buffer.html), but didn't succeed.  Ultimately, nothing is broken but I made this PR WIP and want to ask for help here.

## Testing Instructions

Pull down this branch and start the project. See that nothing is broken in all evergreen browsers.

### DIFF
--- a/src/GeositeFramework/Views/Home/Index.cshtml
+++ b/src/GeositeFramework/Views/Home/Index.cshtml
@@ -82,6 +82,7 @@
     <link rel='shortcut icon' href='img/favicon.ico' type='image/x-icon'/ >
 
     <link rel="stylesheet" href="//js.arcgis.com/3.25/dijit/themes/claro/claro.css">
+    <link rel="stylesheet" href="//js.arcgis.com/3.25/esri/css/esri.css">
     <link rel="stylesheet" href="//js.arcgis.com/3.10/js/dojo/dojox/layout/resources/ResizeHandle.css">
     <link rel="stylesheet" href="//js.arcgis.com/o/232546/geosite-v1.0.10/esri/css/esri.css">
 
--- a/src/GeositeFramework/Views/Home/Index.cshtml
+++ b/src/GeositeFramework/Views/Home/Index.cshtml
@@ -81,9 +81,9 @@
     }
     <link rel='shortcut icon' href='img/favicon.ico' type='image/x-icon'/ >
 
-    <link rel="stylesheet" href="//js.arcgis.com/3.11/dijit/themes/claro/claro.css">
+    <link rel="stylesheet" href="//js.arcgis.com/3.25/dijit/themes/claro/claro.css">
     <link rel="stylesheet" href="//js.arcgis.com/3.10/js/dojo/dojox/layout/resources/ResizeHandle.css">
-    <link rel="stylesheet" href="//js.arcgis.com/o/232546/geosite-v1.0.9/esri/css/esri.css">
+    <link rel="stylesheet" href="//js.arcgis.com/o/232546/geosite-v1.0.10/esri/css/esri.css">
 
     <link rel="stylesheet" href="css/TinyBox2.css">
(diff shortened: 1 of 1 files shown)


fungjj92:
Just found[ this](https://developers.arcgis.com/javascript/3/jshelp/ags_proxy.html) for addressing CORS errors too

mmcfarland:
> Just found this for addressing CORS errors too

@fungjj92 @caseycesari this is a good reminder that we'll be losing the local proxy endpoint that we ship with the app when we go backend-less.

caseycesari:
I've been looking into this this morning, and I'm a little confused because the "calcite" theme is not one of the themes you can choose to add to the JS bundle. It appears to be a separate library all together: https://github.com/Esri/calcite-maps. I think the CORS error is a red herring. It may be worth looking into the [libraries](https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/master/js_requirements.txt) that were included with the previous bundle versus what is included in the bundle now. You can see those libraries by clicking the "View Layer Dependencies" button on the bundle details page:

image

caseycesari:
On [this page](), it states the following:

> Starting with version 3.25, all arcgis.com domains, (i.e. *.arcgis.com), are automatically added to the list. It is no longer necessary to specify individual ArcGIS.com domains. Also, when both the web server and browser support CORS,  esri.request will not use a proxy to perform cross-domain requests.

This may be a case where we have to stop using the proxy.

Also, we may want to look into this

> The API includes some Esri servers by default so it's important to push items on to this array rather than overwriting it.
> ```
> require(["esri/config"], function(esriConfig) {
>  esriConfig.defaults.io.corsEnabledServers.push("servicesbeta.esri.com");
>  esriConfig.defaults.io.corsEnabledServers.push("server.organization.com");
> });
> ```

fungjj92:
The source is that Calcite font face and icons are referenced in all the ESRI CSS
`@font-face {font-family: 'CalciteWebCoreIcons';`
https://js.arcgis.com/3.25/esri/css/esri.css
https://js.arcgis.com/o/232546/geosite-v1.0.10/esri/css/esri.css

fungjj92:
Since the CORS warnings are non-breaking, I pinched them off into #1097 

I considered an upgrade to 3.24 instead of 3.25 to avoid some of the issues observed, but the arcgis bundler has queued my job...I'll check in tomorrow. There's no features in 3.25 we need that aren't in 3.24.

caseycesari:
Any updates on the 3.24 bundle?

fungjj92:
It was still in queue this morn. I stopped it, and restarted a bundle. No luck, though - still in queue. Odd because previous bundle jobs didn't delay.

caseycesari:
Okay. Is this otherwise ready for review? I can take a look if so.

fungjj92:
Yes, ready for review

caseycesari:
There are two follow-up cards for this: #1097 and https://github.com/CoastalResilienceNetwork/regional-planning/issues/100

Issue #1061, opened 2018-01-08
Convert pixels to physical units in print stylesheets


caseycesari:
> 3.  Many styles in the default print stylesheet still use pixels.  For anything print, we should be using physical units like mm, cm, in.  It seems the stylesheet is about 50/50 with physical units and pxs.  This may be the source of some other issues I'm seeing when attempting to size the map and legend elements, but haven't been able to verify yet.

kellyi:
In #1096 I'm devolving most of the app-print-specific style rules into the page-size-specific CSS files, which will make it easier to know which rules have been applied for whichever page size.

PR #1096, opened 2018-09-04
Use fixed rather than relative measurement units for print css rules

kellyi:
## Overview

This PR adjusts the print-related CSS stylesheets to use fixed unit measurements rather than relative units whenever possible. This should simplify the process of positioning elements for a specific paper size and orientation in future.

Connects #1061 

### Notes

Second commit here is there to make it easier to have a intervening step in the print debug workflow in which the print DOM is inspectable; otherwise it doesn't do anything.

## Testing Instructions

- get this branch plus regional planning
- use app print for the four supported paper shapes without a layer legend and verify that each works and prints on a single page
- add some layers, then use app print for the four shapes again with a layer legend and verify that each works and also prints on a single page

### DIFF
--- a/src/GeositeFramework/css/app-print.css
+++ b/src/GeositeFramework/css/app-print.css
@@ -1,35 +1,4 @@
 ﻿@media print {
-    .print-sandbox-header h1 {
-        width: 100%;
-        padding-top: 0.3in;
-        text-align: center;
-    }
-
-    #map-print-sandbox {
-        visibility: visible;
-    }
-
-    #print-map-container {
-        overflow: hidden;
-    }
-
-    .control-container {
-        display: none;
-    }
-
-    #export-print-preview-container {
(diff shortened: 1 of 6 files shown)


kellyi:
> Also, for the duplicated rules, like those for #legend-container-0, do you think it's more work to have them consolidated in app-print.css?

This is a good q. My rationale for splitting them out into the page-specific ones is rather than `app-print` was that I thought it'd make it easier to have the page-specific rules diverge if they weren't consolidated -- so although they're duplicated now, potentially elements can have different rules applied based on the page size.

kellyi:
4e68a02 reduces the page height for a4 landscape to have it fit on a single page. I checked it out in Chrome on both Windows and macOS and it fit on a single page in both browsers.

kellyi:
Great! Thanks for your help with this!

Issue #1010, opened 2017-10-12
Fix default plugin print styling for IE


caseycesari:
With the default plugin print styling, IE prints two pages, the first of which is blank and the second of which contains the map.

![image](https://user-images.githubusercontent.com/1042475/31499382-9ca6c090-af32-11e7-8466-61ab4550d3e1.png)

This differs from Chrome, Firefox, and Safari, all of which only print one page. The placement of the map and legend are also a little off in IE compared to these other browsers.

In order to get an inspectable version of the print layout, I followed the following steps:

- Remove the `print` media block from all of the print stylesheets so that they will apply in the browser (switching them to `screen` should also work).
- Remove call to `previewDeferred.resolve()` so that `window.print` isn't actually called.
- Also remove call to `TINY.box.hide()`, which triggers the print clean-up process.
- Run the site and open it in IE.
- Activate the sample plugin, click a point, then click the plugin print button. The webpage should turn white because the print stylesheets have been added, but they now apply to `screen`.
- Open the developer tools and the inspector, select the `body` element and uncheck the `visibility: hidden` rule added by `print.css`. 
- Click "Print" in the now visible print modal.
- Select the modal in the inspector and delete it, along with the mask layer it adds.
- Select the body and recheck the `visibility: hidden` rule added by `print.css`.
- The DOM should now be in a state that mimics the printable DOM. You can now use the browser's print preview feature to see how the printed page will render.

PR #1102, opened 2018-09-05
Enable debugging print styles in browser

kellyi:
## Overview

This PR adds code to the `Export.js` class to facilitate inspecting the print-specific stylesheets in the browser for debugging purposes.

It works by allowing the user to add a querystring like `?debug=true&size=letter&orientation=portrait` to the site URL, which is then read in `Export.js` to determine whether the app is in print debug mode.

If it's not in debug mode, everything will work as before with regard to printing

If it is in debug mode, clicking "Generate Map Export" in the "Create Map" modal will create the same print-styled document that's output for printing when a developer selects "print" from the "Emulate CSS Media" tool depicted here:

screen shot 2018-09-05 at 5 31 30 pm

This is essentially prep work for #1010 and #1061. #1010 has some pretty involved instructions for handling printing stylesheet setup, and #1061 was difficult to debug past a certain point without being able to inspect the DOM.

Connects #1010 
Connects #1061 

### Demo

screen shot 2018-09-05 at 5 26 01 pm

### Notes

It probably makes sense for this to work only in development (and possibly staging) versions, so I'm happy to add some additional guards to prevent debugging from being active in production.

## Testing Instructions

- build this branch, then visit the main page at the `/` path and use framework print to "Generate Map Export", verifying that it works as it does on `develop`.
- reload the app with the following querystrings and try the sequence below to create an inspectable map print document in the browser:

`?debug=true&size=letter&orientation=portrait`
`?debug=true&size=letter&orientation=landscape`
`?debug=true&size=a4&orientation=portrait`
`?debug=true&size=a4&orientation=landscape`

For each of those the steps to generate an inspectable map document are:

- click "Create Map" in the tools
- click "Generate Map Export" in the modal
- when the background map disappears, open the "Rendering" tab of Chrome developer tools and then select "Print" from the "Emulate CSS Media Dropdown":

screen shot 2018-09-05 at 5 31 30 pm

Note that you'll have to refresh between debugging sessions to restore the map properly.

### DIFF
--- a/src/GeositeFramework/js/Export.js
+++ b/src/GeositeFramework/js/Export.js
@@ -83,6 +83,98 @@ require(['use!Geosite'],
             map.reposition();
         }
 
+        function createPrintableMap(
+            context, previewDeferred, orientDeferred, resizeDeferred, legendDeferred, postPrintAction
+        ) {
+            var mapNode = $("#map-0").detach();
+            var exportMap = $("#export-print-preview-map");
+
+            $('.print-sandbox-header h1').text($("#export-title").val());
+            exportMap.append(mapNode);
+            invalidateSize(context.map);
+
+            context.mapReadyDeferred.then(function() {
+                exportMap.detach().appendTo($("#print-map-container"));
+                invalidateSize(context.map);
+            });
+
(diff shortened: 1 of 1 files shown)


kellyi:

Issue #1104, opened 2018-09-06
Translate additional elements


caseycesari:
- [x] Print prompt window content
- [x] Save and share window content
- [x] Get started headers and “Go” button
- [x] Tour title and content
- [x] Zoom to extent and Opacity labels for each layer in regional planning (although currently listed in the es.json) are not being properly translated

PR #1109, opened 2018-09-14
Translate additional framework elements

caseycesari:
## Overview

Fixes some translation issues and adds more translations for various parts of the framework and plugins. See commit messages for details.

Connects #1104 

### Demo

image

image

image

### Notes

- It was requested that we support translation of the tour content. However, this content is provided via a partial and can be translated directly when setting up a region. 
- Due to https://github.com/CoastalResilienceNetwork/regional-planning/issues/101, I was unable to verify that the opacity controls are now translated. I'll double-check that they are once a fix is in.

## Testing Instructions

- In addition to checking out this branch, check out https://github.com/CoastalResilienceNetwork/regional-planning/pull/102 for the regional planning plugin.
- Switch the framework language to "es" in `region.json`.
- Verify that all of the areas enumerated in the issue have been translated.

### DIFF
--- a/src/GeositeFramework/Views/Home/Index.cshtml
+++ b/src/GeositeFramework/Views/Home/Index.cshtml
@@ -245,7 +245,7 @@
         }
         <nav class="nav-apps plugins nav-apps-narrow @(Model.SinglePluginMode ? "single-plugin-mode" : "")">
             <div id="sidebar-help-area">
-                <a id="help-overlay-start" href="#" data-i18n="Tour">Tour</a>
+                <a id="help-overlay-start" href="#" class="i18n" data-i18n="Tour">Tour</a>
                 <div id="sidebar-toggle">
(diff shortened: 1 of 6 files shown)


kellyi:
I see all the depicted elements above translated on my local test branch. The tour title has also been translated, but the content still seems to be in English:

screen shot 2018-09-17 at 1 57 27 pm

kellyi:
Aside from the tour content, everything else looks good! I approved the regional planning change separately in https://github.com/CoastalResilienceNetwork/regional-planning/pull/102

caseycesari:
Thanks, the tour content will have to be translated separately, on a region by region basis. Here's a related note from above:

> It was requested that we support translation of the tour content. However, this content is provided via a partial and can be translated directly when setting up a region.

kellyi:
> Here's a related note from above:

Thanks! Totally missed that.

caseycesari:
Thanks for the review!

PR #1110, opened 2018-09-18
Develop > Master

caseycesari:
## Overview

Pushes recent bug fixes and improvements from `develop` into `master`.

## Testing instructions

- Run the framework and verify that some of the recent fixes and changes captured here work as expected.


kellyi:
- plugin print demo map working
- mobile single plugin mode settings visible
- translations for regional planning and other elements are there
- a4/letter landscape/portrait print sizes without a layer legend are working

kellyi:
For some reason I can't seem to get a layer legend to show up on the map after adding some layers via regional-planning.

kellyi:
Taking another look.

kellyi:
Tested this out and everything seems to be working well! I did not try out supplying a URL to the OSM map layer option, but using the empty string shows the default-ish OSM basemap.

Issue #1086, opened 2018-08-28
Move templating and conditionals based off region.json to development server


caseycesari:
- Choose a templating engine (in the past we have used http://jinja.pocoo.org/)
- Install this via pip and, for development purposes, use virtualenv. Add some instructions to the README on how to do this.
- Integrate templating into the development server. Keep in this mind that the templating will need to be run when the project is built through [build.py](https://github.com/CoastalResilienceNetwork/geosite-framework-build/blob/master/build.py).
- Templating at this stage will involve reading in `region.json`, and templating variables like [this](https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/128f17cb97401d41ee359f33cb06ae99b6804161/src/GeositeFramework/Views/Home/Index.cshtml#L75).

Depends on #1085

Issue #1100, opened 2018-09-05
Static site clean-up tasks


caseycesari:
- [x] Move the contents of the [project source folder](https://github.com/CoastalResilienceNetwork/GeositeFramework/tree/master/src/GeositeFramework) up a level into the `src` directory. All the parents of this folder will be removed when the project is static.

- [x] Ensure broken icon refs are fixed in the front end (#1108)
- [x] Ensure project README is complete with clear, working instructions to set up the new static site development environment
- [x] Rebase on `develop`.
- [x] Remove all Visual Studio and C# traces

caseycesari:
Requires #1129

caseycesari:
The `feature/task-1` branch should be rebased on `develop` before directories are reorganized and the C#-related files are removed.

Issue #1087, opened 2018-08-28
Migrate HTML partials


caseycesari:
- Ideally these files can remain in their current location, with their current name, etc.
- Drop them into `index.html` when they are referenced.
- Test that all three examples still work as before.
- Partials located [here](https://github.com/CoastalResilienceNetwork/GeositeFramework/tree/master/src/GeositeFramework/Views/Shared).

Depends on #1085 
Depends on #1086

PR #1111, opened 2018-09-18
Setup templating engine and inject region.json into app

fungjj92:
## Overview

The next step in the static site conversion is that we add a templating engine that subs in data from custom JSON into templates that comprise the app. I added Jinja2 as the templating engine and process `region.json` for `index.html` as proof of concept that Jinja2 works for our needs. While we don't use Jinja2 frequently, its template syntax is familiar as it is strongly inspired by Django. Jinja2 is very lightweight as it only provides templating functionality and is very popular among pythonistas.

Connects #1086 

### Demo

Serving the templated `index.html` shows that variables are substituted.
screen shot 2018-09-18 at 11 33 52 am

Running the server looks like
screen shot 2018-09-18 at 11 40 14 am

### Notes

There's a lot of context in the commit messages, check them out.
The script `main.py` writes the jinja template to disk. This workflow works for our dev envir and `build.py` in the geosite-build process whereby we will return zipped static site files to the client to serve on their own.

## Testing Instructions

Create the container, install deps, and run the server
`./scripts/update`
`./scripts/server`
Visit http://localhost:54633 and see that variables are slotted in from region.json. The most prominent one is the text "Geosite Framework Sample" in the middle


scripts/server
```diff
@@ -1,32 +1,23 @@
...
-PORT = 54633
-PATH_TO_PROJECT = 'src/GeositeFramework/'
+if [[ -n "${GEOSITE_DEBUG}" ]]; then
+    set -x
+fi
 
-logging.basicConfig(level=logging.INFO)
+function usage() {
+    echo -n "Usage: $(basename "$0")
+    Starts the python development server.
```
kellyi: Consider specifying that it's starting a Docker container Python development server.

scripts/server.py
```diff
@@ -0,0 +1,32 @@
+#! /usr/bin/env python
+
```
kellyi: Wonder if we should keep the name for this as `server` and rename the Docker script to something like `docker-server` or something to make them more distinct?

scripts/update
```diff
@@ -0,0 +1,29 @@
...
+
+set -e
+
+if [[ -n "${GEOSITE_DEBUG}" ]]; then
+    set -x
+fi
+
+function usage() {
+    echo -n "Usage: $(basename "$0")
+    Attempts to install dependencies.
```
kellyi: We may have considered this, but would it make sense to have these installed via Dockerfile?

kellyi: Nevermind, I see that this is done in https://github.com/CoastalResilienceNetwork/GeositeFramework/pull/1111/commits/e3af147593d39e6cdeb49e6841ddc5cb6cfb259a

scripts/server.py
kellyi: Follow up: we discussed this a bit in Slack and if it were possible I think it'd be nice for these scripts to be able to execute either via `scripts/server` (for environments which have Python 2.7) or via `docker-compose exec app scripts/server` (for environments which have the Dockerized Python runtime).

I believe that since we can assume that `#!/usr/bin/env python` will be set in both cases, these can drop the `.py` file extensions.

In general that should save from having two different sets of scripts -- and we can just add a note to the README to explain how to prefix things with `docker-compose exec` for environments with Docker.

scripts/update
```diff
@@ -0,0 +1,28 @@
...
+}
+
+if [ "${BASH_SOURCE[0]}" = "${0}" ]
+then
+    if [ "${1:-}" = "--help" ]
+    then
+        usage
+    else
+        if docker-compose --version; then
+            docker-compose build
```
kellyi: If we are standardizing on a single set of scripts, I think this can just do a `pip install` from the requirements.txt file

scripts/server.py
caseycesari: 👍

mmcfarland: I'm not sure one can rely on `#!/usr/bin/env python` working on windows.  Perhaps using git shell?  Is the plan to require that?

scripts/main.py
```diff
@@ -0,0 +1,24 @@
...
+        return data
+
+def template_index():
+    j2_env = Environment(loader=FileSystemLoader(''),
+                         trim_blocks=True)
+    region_json = convert_region_json()
+    templated = j2_env.get_template('./GeositeFramework/template_index.html').render(region_json)
+
+    # write jinja template to disk, to be used in geosite static assets build
+    # as well as served the this project's development server
```
caseycesari: Typo: `the` -> `from`, or something similar.

scripts/server.py
```diff
@@ -0,0 +1,38 @@
...
+        HTTPServer.__init__(self, *args, **kwargs)
+        self.RequestHandlerClass.base_path = base_path
+
+class CustomRootHTTPRequestHandler(SimpleHTTPRequestHandler):
+    def translate_path(self, path):
+        words = filter(None, path.split('/'))
+        modified_path = self.base_path
+        for word in words:
+            modified_path = os.path.join(modified_path, word)
+        return modified_path
```
caseycesari: I know this is from a previous PR, but can you add a comment to this function to describe it's purpose?

scripts/server.py
caseycesari: I think we should rely on using the command prompt, at least for the sake of keeping the build tool requirements simple.

kellyi: That makes sense. Ideally it seems like these scripts should be runnable both via

```
python scripts/server
```

and via

```
docker-compose exec app scripts/server
```

and each script's internals can be indifferent to how it's being run.

docker-compose.yml
```diff
@@ -0,0 +1,12 @@
+version: '3'
+services:
+  app:
+    build:
+      context: ./src/GeositeFramework
+    working_dir: /usr/src/
+    ports:
+      - "54633:54633"
```
kellyi: Can we have this run on port 54634? That'd make it possible to serve both the static site and the existing site simultaneously.

scripts/server.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def do_GET(self):     
+        main.template_index() ## create template files    
+
+        return SimpleHTTPRequestHandler.do_GET(self)
+
+def serve():
+    server = CustomRootHTTPServer(PATH_TO_PROJECT, ('', PORT), CustomRootHTTPRequestHandler)
+    logging.info('Now serving on http://localhost:54633')
```
kellyi: Nitpick but it might be worthwhile to use `format` to configure this using the `PORT` value to facilitate future changes.

src/GeositeFramework/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM python:2.7-slim
+
+RUN mkdir -p /usr/src/GeositeFramework
```
kellyi: Will this dir already exist by virtue of having been set up as a volume in the docker-compose file?

scripts/server.py
kellyi: Made this https://gist.github.com/kellyi/45d4defe3a154ea8f0b71c3af9cfc90b as an example of how we might have the scripts work in both environments. It looks like we'll need to use 

```
docker-compose run app python scripts/server.py
```

or whatever script instead of `exec`. I'm not sure how to get the requirements.txt installed libraries to be installed in the container without a Dockerfile; for Node applications we often mount the `node_volumes` dir as a volume in docker-compose. Does `pip install` have a similarly discrete dir for its libraries?

kellyi: I believe we may be able to have `pip install` run as the declared `command` in the docker-compose file for the app service in order to have it ensure that that happens when the container is brought up.

docker-compose.yml
```diff
@@ -0,0 +1,12 @@
+version: '3'
+services:
+  app:
```
kellyi: This probably doesn't super-matter but I wonder if it'd make sense to call this service `server` instead of `app`?

fungjj92:
I believe all comments are addressed now.

The big picture changes were:
- Ensuring the static files could be compiled and served from the local machine as well as docker (in which case the dev is responsible for their python install being clean)
- Ensuring the STRTA workflow was more or less the same between local and docker setups. The difference is adding a `-d` flag.

kellyi:
Taking a look at this now.

Can we add `.pyc` to `.gitignore`?

README.md
```diff
@@ -10,21 +10,47 @@ Copyright (C) 2013 The Nature Conservancy
 ## Static Site Conversion (**work in progress**)
 This project is actively being converted into a static site and away from using .NET on the `feature/task-1` branch. During this time, the project will continue to work off `develop` via the typical set up in Visual Studio on Windows. It may break with Visual Studio on `feature/task-1` and its child branches.
 
+*Note the static site port is 54634 instead of 54633*
+
 ### Run a python development environment
-To run the project locally, on any OS, clone the repo and run `./scripts/update` from root to set up. To run the project server thereafter, `./scripts/server`.
+To run the project locally, on any OS, clone the repo and ensure you have `Python 2.x`. You can choose to work on your host machine or within a docker container. It is advised to work within docker if you'd rather not adjust your local python setup.
```
kellyi: Nitpick but can we capitalize `Docker` and `Python` here?

README.md
```diff
@@ -10,21 +10,47 @@ Copyright (C) 2013 The Nature Conservancy
...
+
+#### Host
+Ensure `pip` is installed.
+
+Install the python requirements:
+`python ./scripts/update.py`
```
kellyi: This may be a quirk related to my machine setup but I had to prefix this command with `sudo` to get it to work. After that it did work, however.

README.md
```diff
@@ -10,21 +10,47 @@ Copyright (C) 2013 The Nature Conservancy
...
+
+Run the development server and serve the static assets:
+`python ./scripts/server.py`
```
kellyi: Ran this and it worked well!

README.md
```diff
@@ -10,21 +10,47 @@ Copyright (C) 2013 The Nature Conservancy
...
+If requirements should change, dependencies can be updated via script:
+`python ./scripts/update.py -d`
+
+Containers can be accessed by bash in the usual way:
+`docker-compose exec <container_name> /bin/bash`
 
-The project uses docker-compose and the app container can be accessed in the usual way:
-`docker-compose exec app /bin/bash`
+To stop the docker server:
+`docker kill server`
```
kellyi: I think this should be

```
docker-compose stop
```

to stop the server

and it might also be worth adding a bit about how you can remove the whole container by running

```
docker-compose down
```

README.md
```diff
@@ -10,21 +10,47 @@ Copyright (C) 2013 The Nature Conservancy
...
-| Name           | Description                                                   |
-| -------------- | ------------------------------------------------------------- |
-| `server`       | Run a Python SimpleHTTPServer serving the templated app       |
-| `update`       | Create app container and install dependencies                 |
\ No newline at end of file
+| Name               | Description                                                   |
+| ------------------ | ------------------------------------------------------------- |
+| `server.py`        | Run a Python SimpleHTTPServer serving the templated app       |
+| `update.py`        | Install python dependencies                                   |
+| `create_static.py` | Write and compile static assets directly                      |
```
kellyi: Right before this table in the section, it'd probably be worthwhile to add a bit about prefixing these scripts with `python` to have them run.

scripts/create_static.py
```diff
@@ -0,0 +1,28 @@
...
+
+if args.d:
+    commands = ['echo "Compiling static assets..."',
+                'docker-compose run --rm server -c' +
+                    '''"import subprocess;subprocess.call('./scripts/main.py')"''',
+                'echo "Finished compiling static assets."']
+    for command in commands:
+        subprocess.call(command, shell=True)
+else:
+    main.template_index()
```
kellyi: Would it be possible and desirable to have this command print some output when it runs successfully. I ran it locally and it exited successfully but there wasn't any output indicating what happened:

```
kinnes@allegheny:~/dvlp/GeositeFramework     (jf/port-region-json-templating *)$ python scripts/create_static.py
kinnes@allegheny:~/dvlp/GeositeFramework     (jf/port-region-json-templating *)$ echo $?
0
```

scripts/server.py
```diff
@@ -0,0 +1,29 @@
...
+
+"""
+Static site server script
+
+Usage: python ./scripts/server.py [OPTIONS]
+
+"""
+import argparse
+import subprocess
+import  py_server
```
kellyi: Looks like there's an extra space here.

kellyi:
I followed the instructions in the README to build and run the project with the non-Docker Python scripts. Everything worked as described, with the tiny exception of having to prefix `python scripts/update.py` with `sudo` to get it to work on my machine.

Trying out the Docker versions now.

README.md
```diff
@@ -10,17 +10,47 @@ Copyright (C) 2013 The Nature Conservancy
...
+
+Install the python requirements:
+```python ./scripts/update.py```
+
+Run the development server and serve the static assets:
+```python ./scripts/server.py```
+
+
+#### Docker container
+Ensure [`docker`](https://www.docker.com/) and `docker-compose` are installed.
```
kellyi: Small addition: docker needs to be running, too, at least on macOS.

kellyi:
Docker versions of these scripts work great!

I was able to run `./scripts/update.py -d` and `./scripts/server.py -d` and `./scripts/create_static.py -d` and each of them worked.

Question: where are the static assets supposed to be after running `create_static.py -d`? I checked in the `dist` dir but don't see anything there.

I made a few comments inline about adjusting the README instruction for how to stop the service to make it more precise. Aside from that this is nearly good to go, though!

scripts/create_static.py
fungjj92: Good point. I'll standardize logging!

scripts/main.py
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env python
+
+import os
+import json
+from jinja2 import Environment, FileSystemLoader
+
+def convert_region_json():
+    with open('./src/GeositeFramework/region.json') as f:
```
mmcfarland: All the file path references in these scripts can be made portable between mac/linux/windows by using the [os.path](https://docs.python.org/2/library/os.path.html) module.  The `os.path.join` calls will add some noise, but cross platform compatibility seems valuable, especially if it's easily had with a few modifications.

In doing so, we can also add some of the parts to constants and not have the path strings sprinkled around.  Something along the lines of this:

```python
BASE_DIR = os.path.join('src', 'GeositeFramework')
REGION_FILE = os.path.join(BASE_DIR, 'region.json')
TMPL_FILE = os.path.join(BASE_DIR, 'template_index.html)
...
```

If they're used across scripts, consider defining them in a file in the module so they could be imported everywhere.

fungjj92: Good suggestion I didn't think about! Building for all platforms is tricky. I anticipate the file refs will only be used here -- but should that change, porting them to a constants file is trivial.

fungjj92:
@kellyi The only static file generated is `index.html`. I imagined that static file compilation and zipping would happen in a later card, such as #1087 or #1100

fungjj92:
Oops! I thought I had made that change but it snuck away. Thanks for the multiple speedy reviews! I will update and merge.

fungjj92:
Updated ~

Issue #1113, opened 2018-09-18
Plugin print demo enhancements


caseycesari:
To further enhance the plugin print workflow demo in the identify print plugin, do the following:

- The size ratio of the print preview map should match the ratio of the map in the printed document. In other words, what the user sees as the extent of the map in the print preview should match the extent of the print preview map.
- Now that there is a print preview map, if a user checks the "add sample layer" checkbox, the layer should appear on the print preview map.

PR #1120, opened 2018-09-20
Hide map legend in demo print preview window

caseycesari:
## Overview

The legend is not necessary or helpful at this stage of the workflow. It
will still display on the printed document.

Connects #1113

### Demo

Before

image

After

image

## Testing Instructions

- Open the identify point plugin and click the print button.
- Add the sample layer, and verify that the legend is not added to the print preview window.
- Click print, and verify that the legend is displayed on the printed document.


caseycesari:
Thanks!

PR #1122, opened 2018-09-21
Print plugin demo stylesheet clean-up

caseycesari:
## Overview

Ensures the document is only one page when printed in major browsers by making the document take up 90% of the page, instead of 100%, which leaves room for margins.

Also removes some rules that were no longer applicable and makes sure the north arrow is not covered by the legend.

Connects #1010

### Demo

**Chrome**
image

**Firefox**
image

**IE**
image

**Edge**
screen shot 2018-09-21 at 2 40 46 pm

## Testing Instructions

- Activate the print process for the identify point plugin.
- Check all of the boxes in the dialog, and click print.
- Ensure the printed document is confined to one page.
- Repeat this process in Chrome, Firefox, IE, and Edge.

### DIFF
--- a/src/GeositeFramework/css/print.css
+++ b/src/GeositeFramework/css/print.css
@@ -29,8 +29,8 @@
 
     #plugin-print-sandbox {
         visibility: visible;
-        height: 100%;
-        width: 100%;
+        height: 90%;
+        width: 90%;
     }
 
     #plugin-print-sandbox .control-container {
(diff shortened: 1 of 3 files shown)


kellyi:
Seems like this is working well! I tested it out in Chrome and Firefox and printing worked as expected. It also worked in IE11, although I couldn't figure out how to export a PDF file so I did XPS instead. The XPS file makes the map shape very narrow, but I don't know that that would be representative of an actual print:

screen shot 2018-09-24 at 10 19 51 am

caseycesari:
Thanks for the review. Let me double-check IE11.

caseycesari:
Pushed a fixup. It seems this was a timing issue. I extended the `setTimeout` that kicks off the browser print, and haven't been able to produce it since.

image

kellyi:
Taking another look at this!

kellyi:
Had trouble getting a pdf printer set up in IE11 but the XPS output looks correct and prints correctly now:

screen shot 2018-09-26 at 11 41 34 am

caseycesari:
Thanks for the thorough review!

Issue #1089, opened 2018-08-28
Migrate region.json validation


caseycesari:
- Find a python library to validation JSON schemas
- Add this library to the requirements
- Integrate into the development server. Keep in this mind that the validation will need to be run when the project is built through [build.py](https://github.com/CoastalResilienceNetwork/geosite-framework-build/blob/master/build.py).
- Where this code lives: [1](https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/master/src/GeositeFramework/Helpers/JsonDataRegion.cs), [2](https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/master/src/GeositeFramework/App_Data/regionSchema.json)

Depends on #1085

kellyi:
I imagine there may be a webpack loader to do JSON schema validation.

fungjj92:
The top-level issue description has been updated since merging the [ADR](https://github.com/azavea/civic-apps/pull/331/files), going with a python environment (to clarify the above suggested to use webpack/JS)

PR #1123, opened 2018-09-26
Validate region.json before static site generation

fungjj92:
## Overview

The Geosite build process pulls in various config files including`region.json`, a plug-in JSON config, etc. to slot into the framework. JSON validation is run on said configs to verify that they are compatible before wasting time on templating, bundling, etc. All of this logic exists in C#. This PR adds JSON validation and builds on the new and growing python setup.

Connects #1089 

### Notes

Looking for inspiration in how logging is done.
Because of the use of the `subprocess` call in `create_static.py`, any errors that should occur in `main.py` aren't easily propagated up to stop the parent script. Thus, the completion logging occurs error or not when running `create_static.py` in docker. 
screen shot 2018-09-26 at 2 07 42 pm
screen shot 2018-09-26 at 2 08 23 pm

## Testing Instructions

Edit `region.json` to be valid and invalid. Run the create static script at each go. Also run the server and make the same changes, see console errors and the page not loading as appropriate.
- change a value to be the wrong type
>     "language": "en",
becomes
>     "language": 1,

- erroneous json formatting, i.e. drop a comma

Test locally:
`python ./scripts/create_static.py`
`python ./scripts/server.py`

Test in docker:
`python ./scripts/create_static.py -d`
`python ./scripts/server.py -d`

### DIFF
--- a/scripts/main.py
+++ b/scripts/main.py
@@ -1,9 +1,10 @@
 #!/usr/bin/env python
 
 import os
+import sys
 import json
 from jinja2 import Environment, FileSystemLoader
-from jsonschema import validate
+from jsonschema import validate, exceptions
 
 BASE_DIR = os.path.join('src', 'GeositeFramework')
 REGION_FILE = os.path.join(BASE_DIR, 'region.json')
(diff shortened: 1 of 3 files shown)


fungjj92:
Updated logging ^_^ 

screen shot 2018-09-26 at 4 27 17 pm

fungjj92:
Thanks for review and trying to address my outstanding note ~

scripts/main.py
```diff
@@ -1,29 +1,48 @@
...
+    except ValueError:
+        raise
+        sys.exit(1)
+    
+    # validate the json against their JSON schema spec
+    try:
+        validate(region_json, region_schema_json)
+    except exceptions.ValidationError:
+        raise
+        sys.exit(2)
```
caseycesari: Very nice!

Issue #1114, opened 2018-09-18
TinyBox error when usePrintModal is set to false


caseycesari:
To reproduce:

- In the opening section of a plugin, set `hasCustomPrint` to `true` and `usePrintModal` to `false`.
- Open the framework and activate the plugin.
- Click the plugin's print button.
- The print process will proceed, but the printed document appears blank, and there is a console in the error:

![image](https://user-images.githubusercontent.com/1042475/45711415-7d16df80-bb57-11e8-8a22-dbb0886f295d.png)

PR #1124, opened 2018-09-26
 Fix plugin print if no modal is used

caseycesari:
## Overview

Previously, opting to not use the plugin print modal broke the print process. The base plugin print code was not written in a way that could accommodate no modal being used. It is reworked here to support both using a modal and not using a modal.

Connects #1114 

### Demo

sep-26-2018 12-10-11

## Testing Instructions

- Activate the identify point plugin, and click the print button. 
- In the modal that appears, select some of the options, drag the map, etc, and click print.
- Verify the printed page is constructed correctly in the preview.
- Cancel the print.
- Modify [this line](https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/9d540d6ea9f6effe589cd900b55e9dd94eca4f46/src/GeositeFramework/sample_plugins/identify_point/main.js#L19) so that `usePrintModal` is set to `False`.
- Refresh the page, activate the identify point plugin again, and click the print button.
- Verify that no modal appears, and that the browser print preview is shown with correctly constructed printed page.

### DIFF
--- a/src/GeositeFramework/js/Plugin.js
+++ b/src/GeositeFramework/js/Plugin.js
@@ -600,7 +600,11 @@ require(['use!Geosite',
                     window.print();
 
                     // Close out the modal, which calls the closejs method
-                    TINY.box.hide();
+                    if ($('.tbox').is(':visible')) {
+                        TINY.box.hide();
+                    } else {
+                        printReset(map, pluginObject);
+                    }
                 });
             });
         }
@@ -617,10 +621,13 @@ require(['use!Geosite',
             // If the plugin is not set up for a print modal,
             // resolve any pending pre-print map operations
             if (!pluginObject.usePrintModal) {
-                modalConfirmDeferred.resolve();
-                postModalDeferred.resolve();
-                mapReadyDeferred.resolve();
-                return mapReadyDeferred;
+                // Wrap in a setTimeout to give the map time to update
+                // before initiating print.
+                window.setTimeout(function() {
+                    printSetup(map, pluginObject, $printSandbox, modalConfirmDeferred, postModalDeferred);
+                }, 750);
+
+                return mapReadyDeferred.resolve(map);
             }
 
             // Setup a print modal window for the user.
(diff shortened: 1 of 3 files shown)


fungjj92:
Tested and what is described works. However, regardless of using the print modal or not, I'm seeing this map zoom jumpiness on IE and Chrome:
ie-print-map-wild-zooming

caseycesari:
Re: map jumpiness, that's due to two things:

1. The map is being removed from it's current place in the DOM into another place in preparation for printing.
2. As part of the demo, I'm recentering the map on NJ to show plugin developers how to manipulate the map.

It's not a great side-effect, but I'm not sure that there is anything I can do to make it not as jumpy.

src/GeositeFramework/js/Plugin.js
```diff
@@ -617,10 +621,11 @@ require(['use!Geosite',
             // If the plugin is not set up for a print modal,
             // resolve any pending pre-print map operations
             if (!pluginObject.usePrintModal) {
-                modalConfirmDeferred.resolve();
-                postModalDeferred.resolve();
-                mapReadyDeferred.resolve();
-                return mapReadyDeferred;
+                window.setTimeout(function() {
```
fungjj92: What does the timeout help with here? Maybe add a comment?

src/GeositeFramework/js/Plugin.js
```diff
@@ -719,6 +680,62 @@ require(['use!Geosite',
             return pluginObj.hasHelp;
         }
 
+        function printSetup(map, pluginObject, $printSandbox, modalConfirmDeferred, postModalDeferred) {
```
fungjj92: its really hard to tell without comparing line by line -- is there any new code in `printSetup()` or `printReset()` ?

fungjj92:
A few comments, otherwise code looks fine and the issue is addressed. The jumpiness is not a new misbehavior, I just hadn't noticed it before. Nice refactor & reorganization!

src/GeositeFramework/js/Plugin.js
caseycesari: Nope! I just separated them out into their own functions so they can be called independently of the modal.

src/GeositeFramework/js/Plugin.js
caseycesari: Good call. Will do.

Issue #1108, opened 2018-09-12
Fix broken file paths


fungjj92:
With the conversion to a static site in #1085, some assets are returning 403 and 404 because their paths are incorrect. The files that are trying to be brought in via the [backend](https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/128f17cb97401d41ee359f33cb06ae99b6804161/src/GeositeFramework/Models/Geosite.cs#L339) remain to be fixed.

![screen shot 2018-09-12 at 4 23 11 pm](https://user-images.githubusercontent.com/10568752/45451294-aa234800-b6a8-11e8-8ed2-192ce5991874.png)

Note the font files and `use.js`

caseycesari:
The use.js error mentioned here is fixed by https://github.com/CoastalResilienceNetwork/GeositeFramework/pull/1130, which corrected the `use` config in the index. This card should focus on fixing the icons, though do note that some of the font files do not resolve on the existing site.

Issue #1084, opened 2018-08-28
Move templating and creation of dropdown menu to front-end


caseycesari:
- Refers to this menu 
![image](https://user-images.githubusercontent.com/1042475/44744503-49561600-aad3-11e8-812e-eeaabbee468b.png)

- Code: https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/128f17cb97401d41ee359f33cb06ae99b6804161/src/GeositeFramework/Views/Home/Index.cshtml#L5-L66

Depends on #1086

PR #1125, opened 2018-09-28
Convert C# dropdown creation to jinja and cleanup

fungjj92:
## Overview

Continue converting templating of what was `index.html` (now `template_index.html`). This PR converts dropdown creation and also cleans up the file's jinja2 formatting generally. Getting closer! The icons (known issue bringing in font files #1100 #1108)

Connects #1084 

### Demo

screen shot 2018-09-28 at 1 14 09 pm

All the links work too 🎉 

## Testing Instructions

`python ./scripts/server.py` or & `-d`
Add/remove the singlePluginMode section from `region.json` and refresh the page. See that the site changes.

### DIFF
--- a/src/GeositeFramework/template_index.html
+++ b/src/GeositeFramework/template_index.html
@@ -1,70 +1,3 @@
-@{
-    Layout = null;
-}
-
-@using GeositeFramework.Models
-@model Geosite
-
-@helper OptionalLinkText(Geosite.Link link)
-{
-    if (link.Url != null)
-    {
-        <a target="_blank" href="@link.Url">@link.Text</a>
-    }
-    else
-    {
-        @link.Text
-    }
-
-}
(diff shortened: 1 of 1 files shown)


src/GeositeFramework/template_index.html
```diff
@@ -232,14 +164,14 @@
     </form>
 
     <script type="text/template" id="template-pane">
-        @if ({{ singlePluginMode }}) {
+        {% if singlePluginMode %}
         <text>
         <div id="single-plugin-mode-help-container">
             Help
         </div>
         </text>
```
kellyi: We should make an issue to remove these `<text>` tags which I think must be valid c# templating but aren't valid html. However it's out of scope for this issue -- just wanted to make sure it didn't get lost.

PR #1126, opened 2018-10-01
Template in HTML partials

fungjj92:
## Overview

Inject the HTML partials into the HTML that comprise other parts of the app conditionally. They were previously brought into the app by C# templating and now are being injected via jinja.

Connects #1087 

### Demo

screen shot 2018-10-01 at 2 21 07 pm

The last 3 elements in the console are injected partials. In CSS the all default to `display: none` hence not seeing them on the browser.

### Notes

I resaved the partials as `utf-8` because jinja requires that templates be this charset and cant parse special unicode chars that come along with other formats.

I didn't move the `.cshtml` files as specified by the issue. All of the C# refs will eventually be torn out of the app and all static site work is on a feature branch, so I'm wondering if I should rather forgo that recommendation and reorganize all the partials. Am I missing something that would prevent me from doing so?

## Testing Instructions

Serve the project.

### DIFF
--- a/scripts/main.py
+++ b/scripts/main.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 
 import os
+import codecs
 import sys
 import plugin_loader
 from jinja2 import Environment, FileSystemLoader
@@ -12,6 +13,7 @@ REGION_SCHEMA_FILE = os.path.join(BASE_DIR, 'App_Data/regionSchema.json')
 TMPL_FILE = os.path.join(BASE_DIR, 'template_index.html')
 IDX_FILE = os.path.join(BASE_DIR, 'index.html')
 
+PARTIALS_DIR= os.path.join(BASE_DIR, 'Views/Shared')
 
 def template_index():
     # create a jinja environment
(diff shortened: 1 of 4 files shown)


caseycesari:
> I didn't move the .cshtml files as specified by the issue. All of the C# refs will eventually be torn out of the app and all static site work is on a feature branch, so I'm wondering if I should rather forgo that recommendation and reorganize all the partials. Am I missing something that would prevent me from doing so?

The plan is to leave the files that are configured for each region (like the partials) where they are, so that region administrators have to make as few changes as possible when migrating to the static site. For an  example, the [gulfmex region](https://github.com/CoastalResilienceNetwork/gulfmex-region) implements the `TourInfo` partial.

caseycesari:
> I resaved the partials as utf-8 because jinja requires that templates be this charset and cant parse special unicode chars that come along with other formats.

Do you remember the error that was thrown? I wondering if this is something that we can do dynamically. I'm hoping to avoid asking site admins to resave all of their partials in case some or all of them are not in utf-8.

fungjj92:
screen shot 2018-10-02 at 12 06 49 pm

Included the whole trace but the last line is really the one.

fungjj92:
Thanks for explaining, that makes sense. I could read in all the partials in `main.py` and rewrite them as `utf-8` alongside wherever we end up writing `index.html` https://stackoverflow.com/questions/19591458/python-reading-from-a-file-and-saving-to-utf-8/19591815

caseycesari:
It looks like this branch still needs some patching up. ca10fb6 looks like it's already on `feature/task-1` as 97e25d0.

fungjj92:
Ah, thanks -- didn't see that. Just had to rebase the branch on `feature/task-1` ~

fungjj92:
Thanks for review ~~

Issue #1129, opened 2018-10-04
Ensure JS loads and works correctly


caseycesari:
Meant as a follow-up card in case fixes in #1108 does not address all of the issues.

The expectation that, after this card is completed, the site should generally be functional and resemble the existing site in functionality. 

Requires #1108

PR #1131, opened 2018-10-05
Fix various asset path issues with static site

caseycesari:
## Overview

See commit messages for details.

Connects #1108 

### Notes

New plugin JS errors will be addressed in #1129 

## Testing Instructions

- Build and serve the static site.
- Verify that the path errors noted in #1108 are fixed by this PR (There will be plugin errors, but those will be addressed in #1129).


fungjj92:
screen shot 2018-10-08 at 12 15 05 pm

rationalPaths only! Nice to see icons and a clean console. I saw this error above, where `layers.json` doesn't exist. Plugin devs probably know to add their own `layers.json` config, but could be worth an error message as per Task 2.

PR #1112, opened 2018-09-18
Add translation files for the launchpad

caseycesari:
## Overview

These should have been checked in along with c9be5c.

Discovered while verifying #1109 

Connects to #1104

### Demo

image

## Testing Instructions

- Switch the framework language to "es" in the region.json.
- Restart the development server
- Load the app, and verify that the launchpad headers and buttons are translated.

Issue #1088, opened 2018-08-28
Migrate translation processing


caseycesari:
- Currently, there is some translation preprocessing that happens on the back-end.
- Reference https://github.com/CoastalResilienceNetwork/GeositeFramework/pull/602 for details on how translation happens in the app, and what needs to be moved to a python script.
- Keep in mind that this step will be called by build.py when the site is built.
- We may need to use a python i18n library.

Depends on #1085

PR #1133, opened 2018-10-08
Migrate internationalization to static site

fungjj92:
## Overview

The Framework can be internationalized by providing translation dictionaries for plugins, core and then specifying what language in `region.json`. This PR ports that ability from the C# site to the static site -- this amounts really to compiling the translation dicts across the various folders using python. We are able to use the already implemented `i18next ` library to template in the values-- it automagically just works behind the scenes. TNC developers won't have to make any updates for translations to work.

Connects #1088 

### Demo

screen shot 2018-10-08 at 2 00 13 pm

Using spanish.

### Notes

There are some known translation fails that were fixed in https://github.com/CoastalResilienceNetwork/GeositeFramework/pull/1109 and https://github.com/CoastalResilienceNetwork/GeositeFramework/pull/1112 . These fixes are on develop.

Some instances of English ("Zoom to Extent") are also because there is no provided translation.

## Testing Instructions

- Serve the static site. 
- ** Ensure there are no (javascript) errors in the console for the map and text to load. If you have any custom plugins downloaded for testing that are erroring, you may want to just delete them.
- The site should be in english

- Change the language in `region.json` to Spanish `es` instead of English (`en`).
- Refresh the web page
- See that translations (as far as they exist).

### DIFF
--- a/scripts/main.py
+++ b/scripts/main.py
@@ -1,18 +1,59 @@
 #!/usr/bin/env python
 
 import os
+import re
 import codecs
 import sys
 import plugin_loader
 from jinja2 import Environment, FileSystemLoader
 from jsonschema import validate, exceptions
-from script_helpers import to_json, BASE_DIR
+from script_helpers import (to_json,
+                            BASE_DIR,
+                            REGION_FILE,
+                            extract_filepaths_from_dirs)
 
-REGION_FILE = os.path.join(BASE_DIR, 'region.json')
 REGION_SCHEMA_FILE = os.path.join(BASE_DIR, 'App_Data/regionSchema.json')
 TMPL_FILE = os.path.join(BASE_DIR, 'template_index.html')
 IDX_FILE = os.path.join(BASE_DIR, 'index.html')
 PARTIALS_DIR = os.path.join(BASE_DIR, 'Views/Shared')
+
+
+def prepare_languages():
+    # get app-wide translation files
+    try:
+        language_dir = os.path.join(BASE_DIR, 'locales')
+        plugin_loader.verify_directories_exist([language_dir])
+        app_json_files = [os.path.join(language_dir, file) for file in
+                          os.listdir(language_dir) if file.endswith('.json')]
+    except ValueError as e:
+        msg = 'Failed! Check the app has JSON translation files. {}'.format(e)
+        sys.exit(msg)
+
+    # get plug-ins' translation files
+    plugin_folder_paths = plugin_loader.get_plugin_folder_paths()
+    plugin_locales = [str(path+'/locales') for path in plugin_folder_paths
+                      if os.path.exists(path+'/locales')]
+    plugin_json_files = extract_filepaths_from_dirs(plugin_locales)
+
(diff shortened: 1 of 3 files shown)


scripts/plugin_loader.py
```diff
@@ -64,6 +63,19 @@ def verify_directories_exist(dirs):
             raise ValueError("{} does not exist".format(path))
 
 
+def get_plugin_folder_paths(**kwargs):
```
mmcfarland: It's not clear that `**kwargs` is necessary here, as it seems to expect only a single argument.  If you're trying to have an optional arg that accepts a default, writing it this way is more conventional:

```python
def get_plugin_folder_paths(base_path=BASE_DIR):
   region_json = to_json(REGION_FILE)   
   plugin_directories = get_plugin_directories(region_json, base_path)
   ...
```

scripts/main.py
```diff
@@ -1,20 +1,65 @@
...
+        plugin_loader.verify_directories_exist([language_dir])
+        app_json_files = [os.path.join(language_dir, file) for file in
+                          os.listdir(language_dir) if file.endswith('.json')]
+    except ValueError as e:
+        msg = 'Failed! Check the app has JSON translation files. {}'.format(e)
+        sys.exit(msg)
+
+    # get plug-ins' translation files
+    plugin_folder_paths = plugin_loader.get_plugin_folder_paths()
+    plugin_locales = [str(path+'/locales') for path in plugin_folder_paths
```
mmcfarland: These concats assume linux/mac file paths, are we still anticipating this be run on windows machines?

scripts/main.py
```diff
@@ -1,20 +1,65 @@
...
+                      if os.path.exists(path+'/locales')]
+    plugin_json_files = extract_filepaths_from_dirs(plugin_locales)
+
+    # merge app and plugin translation dicts keyed to language code
+    # prefer app dict translations, if conflict
+    all_json_files = plugin_json_files + app_json_files
+    translations = {}
+
+    for file in all_json_files:
```
mmcfarland: `file` is a reserved keyword and valid type in python, which is being overwritten for this variable.  Consider renaming this variable to something else.

scripts/plugin_loader.py
fungjj92: A ha! Thanks, that is what I was trying to go for.

scripts/main.py
caseycesari: Good catch. Yes, we do except this to still be run on Windows. We'll need to use the [os.path](https://docs.python.org/2/library/os.path.html) module here.

scripts/main.py
fungjj92: Good catch, thanks!

scripts/main.py
fungjj92: ooh, slipped from using `os.path.join`

fungjj92:
Thanks for review all ~ I will leave this PR up until EOD today in case there's comment on the changes I pushed.

PR #1147, opened 2018-10-24
Static site clean-up

caseycesari:
## Overview

Removes C# components and moves the `src` folder up a directory to make things a bit cleaner.

Connects #1100 

### Notes

After this is merged, we'll have to keep track of changes that are made on `develop` that need to be manually migrated to this branch. This is because of the moving of the `src` folder, which prevents us from rebasing on `develop`.

## Testing Instructions

- Tested through https://github.com/CoastalResilienceNetwork/geosite-framework-build/pull/45


mmcfarland:
Consider including renaming files `cshtml` -> `html`, and insuring they don't have C# or Razor code in them as part of this clean up.

caseycesari:
> Consider including renaming files cshtml -> html

The only `.cshtml` files left are the HTML partials. Changing the extensions would also require that region admins to update the extension of their partials. I wanted to avoid requiring any changes to convert to the static site, so I avoided doing that. 

> and ensuring they don't have C# or Razor code in them as part of this clean up.

All razor code has been removed!

caseycesari:
Thanks!

Issue #1097, opened 2018-09-04
Remove or reimplement proxy after migrating off IIS


fungjj92:
We use a proxy to intercept and adjust requests to ESRI. It runs on IIS currently, and will go away when we switch over to a static site. If its use cases prove essential upon conversion, we'll have to figure out a solution to reinstate the proxy.

Should we have to reinstate the proxy, this issue may connect #1095 which has fall out CORS console errors for getting Calcite icons. They're unsightly and[ potentially solvable ](https://developers.arcgis.com/javascript/3/jssamples/exp_cors_buffer.html)via [the proxy](https://developers.arcgis.com/javascript/3/jshelp/ags_proxy.html). Weirdly, it looks like calcite resources are included in our custom arcgis js api bundle. Fortunately, there's no visible problems in the app and icons are visible. 
![screen shot 2018-08-31 at 4 48 26 pm](https://user-images.githubusercontent.com/10568752/45056356-b08a4200-b060-11e8-852b-2c4bbe650d07.png)
![screen shot 2018-09-04 at 4 15 07 pm](https://user-images.githubusercontent.com/10568752/45056357-b122d880-b060-11e8-8712-c76631e6459b.png)

caseycesari:
A good region to test if the proxy is still required is the gulfmex region. A lot of the plugins make use of the proxy. For example, here are the requests made by the Future Habitat plugin:

![image](https://user-images.githubusercontent.com/1042475/47175370-efa6e680-d2e0-11e8-9a18-2352609f2ae9.png)

PR #1148, opened 2018-10-25
Replace proxy with CORS whitelist

caseycesari:
## Overview

In place of the built-in proxy server, use the CORS functionality built into the ESRI JS API. For most cases, this serves the same purpose as the proxy, but is more modern.

Connects to #1097

## Testing Instructions

- Use the build tool on the `static-site` branch to build a region on this branch, for example:

```
python build.py gulfmex-region --region-branch master --framework-branch cpc/proxy-fix
``` 

- Run the site.
- Activate each plugin, interact with it and verify that no proxy errors are thrown (you would see a large message in the console).
- Repeat the above process with a different region (I tried gulfmex, new jersey).


mmcfarland:
Interestingly, your test build command built a "geosite framework sample" and still had `cors` errors, even though the cors stuff seemed to be set up.

screenshot from 2018-10-25 13 32 42

mmcfarland:
When I run on the `task-1` branch, it builds the right region

`python build.py gulfmex-migrations-region --framework-branch=feature/task-1`

but when I do `cpc/proxy-fix` it goes with sample

`python build.py gulfmex-migrations-region --framework-branch=cpc/proxy-fix`

caseycesari:
@mmcfarland Added a commit to fix the issue that resulted in the site being halfway between normal mode and single plugin mode. I'm not sure if this will fix the issue you were seeing, but mind giving it another test?

mmcfarland:
I can build this successfully, if I use a `develop` region branch:
```
python build.py gulfmex-region --framework-branch=cpc/proxy-fix --region-branch=develop
```
This gets the new CORS code, but is likely not building the correct branch for the region or plugins other than `regional-planning`.  I've brought this issue up in a different forum.  For this PR, I do get CORS errors on a font resource served from js.arcgis.com.  I tried adding that domain to the whitelist, but it didn't appear to have an effect:

screenshot from 2018-10-29 14 36 02

I don't see a noticeable problem with the missing font.

mmcfarland:
I found this that seems related, but I'm not sure it is - also not sure if the revert is including in a recent jsapi release.

https://github.com/Esri/calcite-web/issues/819

caseycesari:
Re: the font issues, I thought I had posted something in the notes, but it appears I didn't do that. Here are some more details.

The calcite web theme appears to be an independent framework and doesn't appear to be included in the ESRI JS bundle in whole. However, there are a few references in the CSS (some classes for tables, a few sprites, and these fonts). The references to the fonts are in the ESRI css bundle that gets created along with the custom JS bundle. I inspected the existing bundle, and the Calcite library is not listed as being included. I tried creating a new bundle, and adding Calcite to the library, but it isn't even an option. 

By default, all of the esri.com domains are included in the CORS enabled site list:
> Starting with version 3.25, all arcgis.com domains, (i.e. *.arcgis.com), are automatically added to the list. It is no longer necessary to specify individual ArcGIS.com domains.

It's really strange that they are still throwing CORS errors given that. I also verified that the fonts are included in the bundle ([example](http://js.arcgis.com/o/232546/geosite-v1.0.10/esri/themes/calcite/icons/fonts/CalciteWebCoreIcons.woff)). 

All that said, it's annoying to see the error in the console, but I don't think it's causing any issues besides the message. I'm hoping that with a new release of the ESRI JS API, it'll be fixed, but it's hard to know for sure.

mmcfarland:
Doesn't seem like there's much to do about it then. :woman_shrugging:

caseycesari:
Thanks!

PR #1149, opened 2018-10-26
Fix docker setup

fungjj92:
## Overview

Oops we forgot about the docker set up when moving the project dir up a layer, so it broke.

Connects #1147

## Testing Instructions

Ensure docker scripts run correctly
`python ./scripts/server.py -d` and check http://localhost:54634/
`python ./scripts/update.py -d`
`python ./scripts/create_static.py -d`

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,11 +2,11 @@ version: '3'
 services:
   server:
     build:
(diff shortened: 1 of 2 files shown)


fungjj92:
thx for review! 😅

PR #1151, opened 2018-10-29
Set minZoom on map

fungjj92:
## Overview

TNC reported a bug (see #1142 for gif) whereby switching from a regional to a global view in the map made the map very tiny, offset, and cut off. Very odd.

From the best of my investigation, it seems like a mixture of problems some our code, some behind the scenes ESRI map CSS. A quick fix that seemed to work was setting a minimum zoom on the map such that we can't zoom out enough to encounter the issue. This seems like an overall positive move because there's no use to see the map so zoomed out and we avoid black space views like:

screen shot 2018-10-29 at 11 58 17 am

Connects #1142

### Demo

bug1142

### Notes

Even though this is a Task 4 contract bug fix, I based and pointed this branch at the static site feature branch for easier testing. If we would rather cherry-pick the commit onto master, I can do that.

## Testing Instructions

There's a lot of ways you could test this. Here's two easier ones:

From the `GeositeFramework` repo:
- Pull down this branch
- Clone [the natural-coastal-protection plugin](https://github.com/CoastalResilienceNetwork/natural-coastal-protection) into the plugin repo
- Switch to the `development` branch of the natural-coastal-protection plugin (defaults to master, which won't build)
- Update the `region.json` to include the natural-coastal-protection plugin in single plugin mode
- Serve the project `python ./scripts/server.py` or & `-d`
- Visit http://localhost:54634/
- Adjust your browser to mobile mode
- Toggle between regions and global

Static site build from the `geosite-framework-build` repo, `static-site` branch:
- clear out your `output folder` (`rm -rf output/`)
- `python build.py coastal-protection-region --framework-branch jf/bugfix#1142 --region-branch development`
- unzip the built project and hit the files with a python server (`cd output`, `unzip coastal-protection.zip`, `python -m SimpleHTTPServer 7979`)
- Visit http://localhost:7979/ or wherever
- Adjust your browser to mobile mode
- Toggle between regions and global

You shouldn't see the issue anymore, and the minzoom should be enforced with no detriment to the map.

PR #1152, opened 2018-10-30
Set minZoom on map

fungjj92:
Prefer over #1151, we want to merge the change into develop instead of the static-site.


fungjj92:
Thanks! Merging.

Issue #1158, opened 2018-11-15
Map util files can't be found when serving from IIS application folder


caseycesari:
If you set up a site in IIS, add an application for the region, and serve it, the map widgets JS can't be found. These files use a relative path that breaks when served from a folder on a domain.

Example: https://github.com/CoastalResilienceNetwork/GeositeFramework/blob/feature/task-1/src/js/widgets/map_utils/measure/main.js#L3

Find a way to import these files that works when the app is served from the top-level or a subfolder.

PR #1159, opened 2018-11-16
Fix paths to map util libraries and plugin icons

caseycesari:
## Overview

Ensures the site works correctly when serving from the root location of a server and from a subfolder.

Connects #1158 

### Demo

**Served from the root**

image

**Served from a subfolder**

image

## Testing Instructions

- Build a region using this branch. Example:
```
python build.py gulfmex-region --framework-branch cpc/fix-paths
```
- Extract the zip to a folder within a folder on your system. 
- Start a python server from within the top level folder and another server from within in extract site contents folder.
- Visit both sites and ensure both of them work and don't have JS errors in the console.

### DIFF
--- a/src/js/Legend.js
+++ b/src/js/Legend.js
@@ -3,7 +3,7 @@ define(['use!Geosite',
          'dojo/window',
          'dojo/dom-geometry',
          'dojox/layout/ResizeHandle',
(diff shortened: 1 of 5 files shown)


caseycesari:
Thanks for the quick review!

PR #1164, opened 2018-12-17
Remove logger

caseycesari:
## Overview

The front-end logger no longer serves a purpose as the back-end logging endpoint has been removed (along with the rest of the back-end).

Connects #1156

### Notes

While the logger was made available for use by plugins, it doesn't appear it was [ever utilized outside of the framework](https://github.com/search?q=org%3ACoastalResilienceNetwork+logError&type=Code).  

In the future, we may want to consider something like [Rollbar](https://rollbar.com/).

## Testing Instructions

- Run the site, and no additional errors are thrown.
- Build the hawaii-region site on this branch: `python build.py hawaii-region --framework-branch cpc/remove-logger`
- Launch the site using the temporary files in the `build` folder or by extracting the zip output.
- Ensure the logger doesn't try to log an error message when the site fails to load, which is does do on the `feature/task-1` branch.


caseycesari:
Thanks!

Issue #1163, opened 2018-12-13
Look into reported DR translation issue


caseycesari:
From TNC:

> Error: Failed to load resource: /dr-secure/languages/es:1  the server responded with a status of 404 ()

caseycesari:
Is the site listed above, `dr-secure`, built off of the `dr-region` repo? I built the `dr-region` site, but didn't encounter this issue.

laflessn:
Yes, dr-secure is built from a “transition” branch we created from the master branch in dr-region. DR only has regional planning, and the region.json is set for language = es. You can see the full error vis the console here https://maps2.coastalresilience.org/dr-secure/ 
Please advise.

PR #1165, opened 2018-12-18
Use json file extension for translation files

caseycesari:
## Overview

The previous implementation used a dynamic endpoint to serve the translation file, which didn't have a file extension. When we initially migrated the translation backend to the static build, we left the file without an extension. This worked for most installations, but proved to be problematic when serving the app from IIS, which uses a whitelist for serving file extensions. Adding the json file extension also makes more practical sense than using no extension since the file is indeed JSON.

Connects #1163 

## Testing Instructions

- Build the `dr-region` site on the `feature/task-1` branch. 
- Deploy the region as virtual directory to IIS. I had success doing this by creating a website that had a root at `C:\inetpub\wwwroot\tnc\` and then adding a virtual directory under that site at `C:\inetpub\wwwroot\tnc\dr`.
- Access the site, and verify the request to return the `es` file fails.
- Rebuild the site on this branch.
- Deploy it to IIS in the same way as above.
- Access the site, and verify the request returns the `es.json` file succesfully.


mmcfarland:
Both branches fail for me in the same way, which is that json isn't a MIME type that IIS will serve

screenshot from 2018-12-19 15 01 17

```
HTTP Error 404.3 - Not Found
The page you are requesting cannot be served because of the extension configuration. If the page is a script, add a handler. If the file should be downloaded, add a MIME map.
```

Typically we fix this by adding a web.config that removes/adds the *.json extension as a valid MIME type.

However, if their server has already manually been configured to serve *.json, the code looks like it'd work - however, building with the syntax

```
python build.py dr-region --region-branch transition --framework-branch cpc/fix-translation
```

still produces language files without the file extension.

screenshot from 2018-12-19 15 05 32

PR #1177, opened 2019-03-13
Use correct region.json attribute for GA key

caseycesari:
## Overview

We were mistakenly using the Google URL shortener key instead of the Google Analytics key.

Connects #1176

### Demo

image

## Testing Instructions

- Build and serve the site with `./scripts/server.py`, and verify the correct key is used to populate the GA key.


caseycesari:
Thanks!

